PR #11, opened 2021-05-03
docs: improve README to fix #5

### LINKED ISSUE #5: Explain in the README what makes git-branchless better
The README says that `git-branchless` makes these things easier:

>   * Switching between work tasks.
>   * Separating minor cleanups/refactorings into their own commits, for ease of reviewability.
>   * Extensively rewriting local history before submitting code for review.
>   * Performing speculative work which may not be ultimately committed.
>   * Working on top of work that you or a collaborator produced, which is not yet checked in.
>   * Losing track of `git stash`es you made previously.

But it doesn't say _how_ it makes those tasks easier.

For example, I switch between work tasks today with [`git switch`](https://git-scm.com/docs/git-switch). How is `git-branchless` better than `git switch`? In the demo video, you demonstrate switching to a a commit as a detached HEAD instead of a branch. But I could have typed `git switch -c experiment1`. Was the whole advantage just "you don't have to type a name?" But then I have to use commit IDs as names. Seems about the same. What's the difference?

arxanas:
@dfabulich if you have time, would you mind reviewing this update to the README and tell me what you think? You can preview it directly at https://github.com/arxanas/git-branchless/tree/readme


dfabulich:
This is a huge improvement!

arxanas:
Thanks for looking it over :)

PR #13, opened 2021-06-21
Allow git-branchless to build on Windows

waych:
- Disable default-features for cursive, as ncurses-backend is the default,
which is not available on Windows and isn't required with
crossterm-backend selected.
- Use a bundled build of sqlite3.


arxanas:
@waych Thanks for fixing this! I don't have a Windows device to test on. Did you try it out, or did you just fix the build failures?

waych:
Yeah I kicked the tires a bit this morning. Haven't done much other than initialize a few repos and play with git-sl and git-undo a bit so far. I was a big fan of hg changeset evolution on piper, so I'm excited to try restack.

Thanks for the extensions!

arxanas:
@waych You might also be interested to try `git move` by installing from the latest master. It's similar to `hg rebase` in that it can rebase entire subtree. It tries an in-memory rebase first.

I intend to reimplement `git restack` in terms of it, which should improve the experience. The in-memory rebasing would also be useful for e.g. rewording commit messages.

The performance `git move` for large repos also needs to be fixed.

waych:
FYI I tried more stuff and found that `git restack` and `git move` commands are still broken on Windows, along with various tests.  Most of the issues seem to deal with the hooks not being directly executable on Windows (have to load via bash.exe) and uses of "true and /bin/echo". Nothing major.

I'm working through a stack of fixes that should fix the above.  Got everything to pass on Windows, and am now reworking it so my changes don't also break Linux use. Will send PR soon.

PR #20, opened 2021-06-22
Fix tests and execution on Windows

waych:
This PR fixes various issues affecting execution of git-branchless commands and tests on Windows.

- Fixes handling of PATH environment variable (separator)
- Calls "echo" from PATH on Windows, as git-bash actually keeps it at /usr/bin/echo
- Strips EL VT sequence from test output (injected by git.exe or ConPTY?)
- Use ":" for GIT_EDITOR to have git skip commit editing (during test invocations)
- Provide a default git.exe path (for when use_system_git == true)
- Call hooks via execution of sh, using git-bash on Windows when found.

With these changes all tests pass and commands appear to work as intended.

### DIFF
--- a/src/testing.rs
+++ b/src/testing.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
 
-use crate::util::{wrap_git_error, GitExecutable, GitVersion};
+use crate::util::{get_sh, wrap_git_error, GitExecutable, GitVersion};
 use anyhow::Context;
 use fn_error_context::context;
 
@@ -99,11 +99,16 @@ impl Git {
             .git_executable
             .parent()
             .expect("Unable to find git path parent");
-        std::env::join_paths(vec![
+        let bash = get_sh().expect("bash missing?");
+        let bash_path = bash.parent().unwrap();
+        std::env::join_paths(
+            vec![
                 // For Git to be able to launch `git-branchless`.
                 branchless_path,
                 // For our hooks to be able to call back into `git`.
                 git_path,
+                // For branchless to manually invoke bash when needed.
+                bash_path,
             ]
             .iter()
             .map(|path| path.to_str().expect("Unable to decode path component")),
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -274,6 +275,42 @@ impl FromStr for GitVersion {
...
+        })
+    })
+}
+
+/// Returns the path to bash.
+pub fn get_bash() -> Option<PathBuf> {
+    let exe_name = if cfg!(target_os = "windows") {
+        "bash.exe"
+    } else {
+        "bash"
```
arxanas, line 297: Not all Unix systems will have `bash` installed (some will have `dash` instead). But they should have `sh`, in order to be POSIX-compatible. Let's change this function to return `sh` in this case.

arxanas (commented):
@waych Thanks for working on this! One question, would you be able to get Github Actions to run CI for Windows as well? I'm almost certain to break the Windows build later if I don't have something to test on...

src/util.rs
waych: Switch to sh for the !windows case.

waych:
PTAL

Rebased and addressed feedback.

arxanas:
LGTM, thanks again!

arxanas:
@waych Do you think there's anything else which should be done to say that it supports Windows? If not, I can tag a new release and announce support.

waych:
@arxanas Can't think of anything else that needs to be done at this point other than updates to the changelog. 😃

src/testing.rs
```diff
@@ -150,15 +161,9 @@ impl Git {
             // messages. Usually, we can set this with `git commit -m`, but we have
             // no such option for things such as `git rebase`, which may call `git
             // commit` later as a part of their execution.
-            ("GIT_EDITOR", {
-                // Some systems have `/bin/true` and others have
-                // `/usr/bin/true`. Pick whichever one we can find.
-                if Path::new("/bin/true").exists() {
-                    "/bin/true"
-                } else {
-                    "/usr/bin/true"
```
nh2, line 159: Thanks, this also fixes tests on Linux distros where `/usr/bin` and `/bin` do not exist.

PR #22, opened 2021-06-23
readme: add a mention of my experimental VCS "Jujutsu"

martinvonz:
I'm glad you found my project and mentioned it in the wiki and on Hacker News :) Thanks! Also, I really like the interactive undo feature. I've wanted to add that kind of "time travel" feature to jj too.

This PR adds a mention of it to the READMe's list of alternative tools. Let me know what you think. I'd understand if you think it should not be listed either because it's still experimental, or because it's too different by not being able to run in an existing git working copy.


arxanas:
That jj wasn't in the list of alternatives was an oversight! Thanks for
adding. As I work on some of the git-branchless features, like undoing
staged changes, I've come to appreciate the effectiveness of a more
principled solution to source control...

śr., 23 cze 2021, 09:24 użytkownik Martin von Zweigbergk <
***@***.***> napisał:

PR #24, opened 2021-06-23
Split CI workflows up for the OS badges

waych:
Allows the reader to see that the project's CI supports their platform
at a glance.

Badge links format looks to have changed. Tested by pointing temporarily
at my own repo to see the badges in the rendered README.md.

### DIFF
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -1,0 +1,26 @@
+name: Static Analysis
+
+on: ["push", "pull_request"]
+
+jobs:
+  static-analysis:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt, clippy
+      - name: Run `cargo fmt`
+        uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
+      - name: Run `cargo clippy`
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: -- -D warnings
+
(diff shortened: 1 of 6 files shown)


.github/workflows/fmt.yml
```diff
@@ -0,0 +1,26 @@
+name: Static Analysis
```
arxanas, line 1: Let's name this file something like `lint.yml` or `linting.yml` or `static-analysis.yml`, since `clippy` is not a formatting linter.

.github/workflows/macos.yml
```diff
@@ -0,0 +1,18 @@
+name: macOS
```
arxanas, line 1: In my experience, it takes a while to provision a macOS machine, which gets in the way of actually merging PRs. They're also relatively expensive (10x the cost of Linux minutes). Can we instead run this job 1x/week or similar?

arxanas (changes requested):
Neat, thanks for doing this.

.github/workflows/macos.yml
waych: Switched to run twice a week.

I was under the impression that workflow minutes were free for opensource? In either case I understand desire to avoid macOS in the PR workflow as it can be frustratingly slow / unreliable.

.github/workflows/fmt.yml
waych: Switched to lint.yml / `Lint`

arxanas (approved):
We get 2000 free minutes per month. There's a 1x multiplier for Linux, 2x for Windows, and 10x for macOS. So the typical 3-minute build on Linux would count for 30 minutes on macOS!

arxanas:
Hm, maybe I was mistaken, and the repository gets unlimited minutes since it's not private!

waych:
Sweet.  In that case I'll send a PR to schedule the macOS everyday so its still out of the way but gets a better signal in case something regresses.

PR #38, opened 2021-06-26
Attempt to fix flaky undo tests (closes #23)

### LINKED ISSUE #23: flake: test_undo_hide
Just noticed this.  Happened on a run of run-tests against git `2.29.2`.

https://github.com/waych/git-branchless/runs/2898734801 

May be related to the `commit -a` observation?

```
---- command::test_undo::test_undo_hide stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Differences ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: undo_hide-5
Source: tests/command/test_undo.rs:282
-old snapshot
+new results
────────────────────────────────────────────────────────────────────────────────
stdout
────────────┬───────────────────────────────────────────────────────────────────
    0     0 │ O f777ecc9 (master) create initial.txt
    1     1 │ |\
    2       │-| o 62fc20d2 (test1) create test1.txt
          2 │+| @ 62fc20d2 (test1) create test1.txt
    3     3 │ |
    4       │-@ fe65c1fe create test2.txt
          4 │+o fe65c1fe create test2.txt
────────────┴───────────────────────────────────────────────────────────────────
thread 'command::test_undo::test_undo_hide' panicked at 'snapshot assertion for 'undo_hide-5' failed in line 282', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/insta-1.7.1/src/runtime.rs:1036:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    command::test_undo::test_undo_hide
```

arxanas:
- refactor(util): rename `GitExecutable` to `GitRunInfo`
- refactor(util): store more environment information in `GitRunInfo`

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -150,33 +150,28 @@ pub fn get_db_conn(repo: &git2::Repository) -> anyhow::Result<rusqlite::Connecti
 
 /// Path to the `git` executable on disk to be executed.
 #[derive(Clone, Debug)]
-pub struct GitExecutable(pub PathBuf);
+pub struct GitRunInfo(pub PathBuf);
 
 /// Run Git in a subprocess, and inform the user.
 ///
 /// This is suitable for commands which affect the working copy or should run
 /// hooks. We don't want our process to be responsible for that.
 ///
-/// Args:
-/// * `out`: The output stream to write to.
-/// * `err`: The error stream to write to.
-/// * `git_executable`: The path to the `git` executable on disk.
-/// * `event_tx_id`: The ID of the current event-log transaction, if any.
-/// * `args`: The list of arguments to pass to Git. Should not include the Git
+/// `args` contains the list of arguments to pass to Git, not including the Git
 /// executable itself.
 ///
-/// Returns: The exit code of Git (non-zero signifies error).
-#[context("Running Git ({:?}) with args: {:?}", git_executable, args)]
+/// Returns the exit code of Git (non-zero signifies error).
+#[context("Running Git ({:?}) with args: {:?}", git_run_info, args)]
 #[must_use = "The return code for `run_git` must be checked"]
 pub fn run_git<S: AsRef<str> + std::fmt::Debug>(
-    git_executable: &GitExecutable,
+    git_run_info: &GitRunInfo,
     event_tx_id: Option<EventTransactionId>,
     args: &[S],
 ) -> anyhow::Result<isize> {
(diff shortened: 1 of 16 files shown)


src/util.rs
```diff
@@ -275,17 +287,34 @@ pub fn run_hook(
...
+            working_directory: _,
+            env,
+        } = git_run_info;
         let mut child = Command::new(get_sh().context("shell needed to run hook")?)
             .arg("-c")
-            // This is always passed as a current dir unix path, as `sh` is processing it as an
-            // argument, even on Windows.
-            .arg(format!("./{} \"$@\"", hook_name))
+            .arg(format!(
+                "{} \"$@\"",
```
waych, line 303: FYI this argument needed the `./` prefix on Windows it wouldn't look in current directory otherwise.  I think this is due to gitbash's weird argument path translation, but am not sure.

arxanas: Technically, there's no stipulation that the hook directory has to be somewhere under the current repository, such as if `git.core.hooksPath` is set to something else, so I think it's wrong to use `./` in this case.

waych: Hmm actually I don't understand why the `./` isn't needed in Linux also: it isn't like we are expecting `.` to be in $PATH(?)

waych: Ah I see you removed the `.current_dir(hook_path)`  which made the `./` correct 😄

waych: I see what is happening.  See this line in the `bash` manpage:

```
If  the name is neither a shell function nor a builtin, and contains no slashes, bash searches each element of the PATH for a directory containing an executable file by that name.
```

and later

```If the search is successful, or if the command name contains one or more slashes, the shell executes the named program in a separate execution  environment.
```

Because `std::fs::canonicalize(&hook_path)` most certainly does contain `/`, the shell is interpreting it as a path to binary, whereas on Windows, lacking a '/' it is falling back to searching via PATH.  This explains why it worked for me when I added the `./` prefix unconditionally.


`std::fs::canonicalize(&hook_path)`

waych: Thinking about ways this could be fixed without resorting to path mangling, it feels like the simplest way to resolve is to just add the canonicalized hook_path to the start of PATH and invoke the hook by name directly.

arxanas:
Thanks for the help!

PR #40, opened 2021-06-26
Fix 'cannot spawn .git/hooks/post-rewrite' on Windows

waych:
Following is happening:

'''
error: cannot spawn .git/hooks/post-rewrite: No such file or directory: No such file or directory
'''

because the first line of the stubbed post-rewrite is a new-line, which
is interpretted by mingw as "not a shell script".

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -485,12 +485,11 @@ mod tests {
                     .join(".git")
                     .join("hooks")
                     .join("post-rewrite"),
-                r#"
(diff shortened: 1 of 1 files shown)


waych:
Updated commit to say this isn't bash that is interpreting that line, but rather mingw. `mingw_spawnvpe` calls `parse_interpreter`.

arxanas:
My hero ❤️

PR #131, opened 2021-10-19
feat(tui): add flag to allow overriding glyphs detection

### LINKED ISSUE #130: Smartlog displays uncolored output under `watch --color git sl`
I'd expect `watch --color git sl` to output a colored output, but it seems to display the uncolored output.

Perhaps smartlog needs something like `--color=always` to override whatever detection mechanism it's using for whether to display colors

bcongdon:
Adds a --glyphs flag to override the default detected glyphs setting.

This is useful when a user wants to force "pretty" glyphs to be printed, even if the terminal doesn't present itself as a TTY interface. (For example, when running with 'watch').

This fixes #130.

Example: `git-branchless --glyphs=pretty smartlog`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `git branchless init` takes a `--main-branch` option to specify the name of the main branch without interactive prompting.
+- The `--color=[auto,always,never]` flag can be used to override the automatically detected value for terminal colors.
+- The `CLICOLOR` and `NOCOLOR` environment variables are now respected.
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


arxanas:
I added you as a collaborate for the repository, so feel free to commit once the above comment is addressed.

epage:
Maybe give these two crates a try
- https://docs.rs/concolor-control/0.0.7/concolor_control/index.html
- https://docs.rs/concolor-clap/0.0.6/concolor_clap/

They provide a standardized, reusable, clap arguments for color control and then give you the ability to check whether the terminal / env say you can use control.

Here is an example of me adopting it: https://github.com/crate-ci/committed/pull/206/files

The `concolor_control` API for setting the color state from the CLI is behind a feature flag to communicate that its cross-crate control is unstable until we go 1.0.  In the mean time, its fine to use it for single-crate control.

PR #159, opened 2021-10-26
ci: add Nix build on Linux and macOS

hmenke:
As promised in #120.

This introduces a new file `default.nix` at the top-level of the repository. If it is in the way, I can also hide it somewhere else and just move it to the top-level in the CI.

The Nix build is rather slow because it transitively fetches all of the dependencies and nothing is cached paired with general build slowness of Rust. It's possible to obtain some caching but it involves a third-party provider. GitHub itself does not offer a Nix binary cache (yet?).

### DIFF
--- a/.github/workflows/nix-linux.yml
+++ b/.github/workflows/nix-linux.yml
@@ -1,0 +1,22 @@
+name: Nix on Linux
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  nix:
+    runs-on: ubuntu-latest
+    name: nix-build
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: cachix/install-nix-action@v14.1
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - run: nix build --print-build-logs --show-trace --override-input nixpkgs github:NixOS/nixpkgs
--- a/.github/workflows/nix-macos.yml
+++ b/.github/workflows/nix-macos.yml
@@ -1,0 +1,29 @@
+name: Nix on macOS
+
+on:
+  schedule:
+    # Run once every day at 6:40AM UTC.
+    - cron: "40 6 * * *"
+
+  push:
+    branches:
+      - master
+
+  pull_request:
+    branches:
+      - "ci-*"
+
+jobs:
+  nix:
+    runs-on: macos-latest
+    name: nix-build
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: cachix/install-nix-action@v14.1
+        with:
+          extra_nix_config: |
+            experimental-features = nix-command flakes
+      - run: nix build --print-build-logs --show-trace --override-input nixpkgs github:NixOS/nixpkgs
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,53 @@
+{
+  description = "git-branchless";
+
+  outputs = { self, nixpkgs, ... }:
+    let
+      lib = nixpkgs.lib;
+      systems = [
+        "aarch64-linux"
+        "aarch64-darwin"
+        "i686-linux"
+        "x86_64-darwin"
+        "x86_64-linux"
+      ];
+      foreachSystem = f: lib.foldl' (attrs: system: lib.recursiveUpdate attrs (f system)) { } systems;
+    in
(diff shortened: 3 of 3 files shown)


hmenke:
When the jobs succeeded they took around 12 mins so I set a timeout of 20 mins now.

SuperSandro2000:
The nix-build is run in a none interactive shell and I think that might be the issue.

```
git-branchless> test core::test_hooks::test_abandoned_commit_message ... ok
git-branchless> test core::test_hooks::test_interactive_rebase_noop ... ok
git-branchless> test core::test_hooks::test_merge_commit_recorded ... ok
git-branchless> test core::test_hooks::test_rebase_individual_commit ... ok
git-branchless> test core::test_hooks::test_fixup_no_abandoned_commit_message ... ok
git-branchless> test command::test_navigation::test_checkout_pty has been running for over 60 seconds
git-branchless> test command::test_navigation::test_next_ambiguous_interactive has been running for over 60 seconds
error: building of '/nix/store/71glgh79h0piw4kh9mlasw1w1z9f7p7m-git-branchless.drv!out' from .drv file timed out after 600 seconds of silence
```

arxanas:
@SuperSandro2000 The test spawns a PTY to run the test, so I don't think it affects the test whether or not it's been spawned in an interactive shell. In any case, I don't think that would explain the behavior of hanging forever.

@hmenke I wasn't able to resolve the issue on my end. If you can manually verify that the interactive behavior works (run `git branchless checkout` and confirm that it lets you select and check out a commit interactively), then we could set an environment variable or configuration flag and exclude PTY tests from the Nix build, and ship it that way.

hmenke:
The tests `test_checkout_pty` and `test_next_ambiguous_interactive` work perfectly when run interactively and I've verified that in a number of different settings when trying to make them fail outside the sandbox. For now I have disabled those tests and added a FIXME note.

arxanas:
@hmenke Thanks for the pull request! For now I've merged this. In the future, we're likely to add more PTY tests, and I think it would be easier if we didn't have to specify all of them in the Nix build file. Do you know if there's a way to pass the `--cfg` flag to `rustc`, or set an environment variable during compilation, so that we can indicate inline in the code which tests depend on a PTY?

arxanas:
@hmenke Hm, it looks like the Nix build failed immediately on `master` with this error:

```
  nix build --print-build-logs --show-trace --override-input nixpkgs github:NixOS/nixpkgs
  shell: /bin/bash -e {0}
  env:
    NIX_SSL_CERT_FILE: /nix/var/nix/profiles/default/etc/ssl/certs/ca-bundle.crt
error: unable to download 'https://api.github.com/repos/NixOS/nixpkgs/commits/HEAD': HTTP error 403 ('')

       response body:

       {"message":"API rate limit exceeded for 199.7.166.17. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}

       … while fetching the input 'github:NixOS/nixpkgs'

       … while updating the flake input 'nixpkgs'

       … while updating the lock file of flake 'git+file:///Users/runner/work/git-branchless/git-branchless?ref=master&rev=eebd15d2901a9fe7384b05f67f0eb42874b28cc3'
```

Workflow run: https://github.com/arxanas/git-branchless/runs/4087952810?check_suite_focus=true

We should be able to perform an authenticated request using the workflow token we have during the run: https://dev.to/github/the-githubtoken-in-github-actions-how-it-works-change-permissions-customizations-3cgp

Do you know if it's possible to get `nix build` to perform an authenticated request using this token?

arxanas:
For some reason, the build is only failing on nix on macOS.

hmenke:
Very sorry about this. This is a generic problem with the macOS runners which all run on the same physical machine. I hope this fixes it: https://github.com/hmenke/git-branchless/commit/f4337c0f7f10e7beb26d2fe02d23fac937ff3a64

arxanas:
It's no trouble, I'm happy that we can support Nix :)

jirutka:
FYI: I’m packaging git-branchless for Alpine Linux and I also have problem with tests hanging forever on our CI.

arxanas:
@jirutka we ended up just skipping these tests since I wasn't able to reproduce them:

https://github.com/arxanas/git-branchless/blob/f680141a0d258799e644e228ca2111a9a43724ea/flake.nix#L43-L46

If you can get repro steps (maybe with a Docker image?), then file an issue and I can take another look. (Adding Nix to the setup is somewhat harder for me to debug since I'm not familiar with it.)

PR #170, opened 2021-10-28
feat(amend): Initial amend command

bcongdon:
Create a new amend command that automatically amends changes to tracked files onto the HEAD commit, and can optionally perform an immediate restack.

(See e.g. #48)

### DIFF
--- a/src/commands/amend.rs
+++ b/src/commands/amend.rs
@@ -1,0 +1,165 @@
+//! Amend the current commit.
+//!
+//! This command amends the HEAD commit with changes to files
+//! that are already tracked in the repo. Following the amend,
+//! the command performs a restack.
+
+use std::convert::TryInto;
+use std::fmt::Write;
+use std::time::SystemTime;
+
+use eyre::Context;
+use itertools::Itertools;
+
+use crate::commands::gc::mark_commit_reachable;
+use crate::commands::restack;
+use crate::core::config::get_restack_preserve_timestamps;
+use crate::core::effects::Effects;
+use crate::core::eventlog::{Event, EventLogDb};
+use crate::core::formatting::Pluralize;
+use crate::git::{AmendFastOptions, FileStatus, GitRunInfo, Repo};
+use crate::opts::MoveOptions;
+
+/// Amends the existing HEAD commit.
+pub fn amend(
+    effects: &Effects,
+    git_run_info: &GitRunInfo,
+    move_options: &MoveOptions,
+) -> eyre::Result<isize> {
+    let now = SystemTime::now();
+    let repo = Repo::from_current_dir()?;
+    let conn = repo.get_db_conn()?;
+    let mut event_log_db = EventLogDb::new(&conn)?;
+
+    let head_oid = match repo.get_head_info()?.oid {
+        Some(oid) => oid,
+        None => {
+            writeln!(
+                effects.get_output_stream(),
+                "No commit is currently checked out. Check out a commit to amend and then try again.",
+            )?;
+            return Ok(1);
+        }
+    };
+    let head_commit = repo.find_commit_or_fail(head_oid)?;
+
+    let index = repo.get_index()?;
+    if index.has_conflicts() {
+        writeln!(
+            effects.get_output_stream(),
+            "Cannot amend, because there are unresolved merge conflicts. Resolve the merge conflicts and try again."
+        )?;
+        return Ok(1);
+    }
+
+    let event_tx_id = event_log_db.make_transaction_id(now, "amend")?;
+    let staged_index_paths = repo.get_staged_paths()?;
+    let (opts, dirty_working_tree) = if !staged_index_paths.is_empty() {
+        let dirty_working_tree = repo.has_changed_files(effects, git_run_info)?;
+        let opts = AmendFastOptions::FromIndex {
+            paths: staged_index_paths.into_iter().collect_vec(),
+        };
+        (opts, dirty_working_tree)
+    } else {
+        let status = repo.get_status(git_run_info, Some(event_tx_id))?;
+        let entries_to_amend = status
+            .into_iter()
+            .filter(|entry| match entry.working_copy_status {
+                FileStatus::Added
+                | FileStatus::Copied
+                | FileStatus::Deleted
+                | FileStatus::Modified
+                | FileStatus::Renamed => true,
+                FileStatus::Ignored
+                | FileStatus::Unmerged
+                | FileStatus::Unmodified
+                | FileStatus::Untracked => false,
+            })
+            .collect_vec();
+        let opts = AmendFastOptions::FromWorkingCopy {
+            status_entries: entries_to_amend,
+        };
+        (opts, true)
+    };
+    if opts.is_empty() {
+        writeln!(
+            effects.get_output_stream(),
+            "There are no uncommitted or staged changes. Nothing to amend."
+        )?;
+        return Ok(0);
+    }
+
+    let amended_tree = repo.amend_fast(&head_commit, &opts)?;
+    let timestamp = now.duration_since(SystemTime::UNIX_EPOCH)?.as_secs_f64();
+
+    let (author, committer) = (head_commit.get_author(), head_commit.get_committer());
+    let (author, committer) = if get_restack_preserve_timestamps(&repo)? {
+        (author, committer)
+    } else {
+        (
+            author.update_timestamp(now)?,
+            committer.update_timestamp(now)?,
+        )
+    };
+
+    let amended_commit_oid = head_commit.amend_commit(
+        Some("HEAD"),
+        Some(&author),
+        Some(&committer),
+        None,
+        Some(&amended_tree),
+    )?;
+    mark_commit_reachable(&repo, amended_commit_oid)
+        .wrap_err("Marking commit as reachable for GC purposes.")?;
+
+    event_log_db.add_events(vec![Event::RewriteEvent {
+        timestamp,
+        event_tx_id,
+        old_commit_oid: head_oid.into(),
+        new_commit_oid: amended_commit_oid.into(),
+    }])?;
+
+    if let AmendFastOptions::FromWorkingCopy { .. } = opts {
+        // TODO(#201): Figure out a way to perform "fast amend" on the working copy without needing a reset.
+        git_run_info.run(effects, Some(event_tx_id), &["reset"])?;
+    }
+
+    let restack_exit_code = restack::restack(
+        effects,
+        git_run_info,
+        vec![head_oid.to_string()],
+        move_options,
+    )?;
+    if restack_exit_code != 0 {
+        return Ok(restack_exit_code);
+    }
+
+    match opts {
+        AmendFastOptions::FromIndex { paths } => {
+            let staged_changes = Pluralize {
+                amount: paths.len().try_into()?,
+                plural: "staged changes",
+                singular: "staged change",
+            };
+            let mut message = format!("Amended with {}.", staged_changes.to_string());
+            // TODO: Include the number of uncommitted changes.
+            if dirty_working_tree {
+                message += " (Some uncommitted changes were not amended.)";
+            }
+            writeln!(effects.get_output_stream(), "{}", message)?;
+        }
+        AmendFastOptions::FromWorkingCopy { status_entries } => {
+            let uncommitted_changes = Pluralize {
+                amount: status_entries.len().try_into()?,
+                plural: "uncommitted changes",
+                singular: "uncommitted change",
+            };
+            writeln!(
+                effects.get_output_stream(),
+                "Amended with {}.",
+                uncommitted_changes.to_string(),
+            )?;
+        }
+    }
+    Ok(0)
+}
--- a/src/git/repo.rs
+++ b/src/git/repo.rs
@@ -11,7 +11,7 @@
 
 use std::borrow::{Borrow, Cow};
 use std::collections::{HashMap, HashSet};
-use std::convert::TryInto;
+use std::convert::{TryFrom, TryInto};
 use std::ffi::{OsStr, OsString};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -23,7 +23,9 @@ use cursive::theme::BaseColor;
 use cursive::utils::markup::StyledString;
 use eyre::{eyre, Context};
 use itertools::Itertools;
+use lazy_static::lazy_static;
 use os_str_bytes::{OsStrBytes, OsStringBytes};
+use regex::bytes::Regex;
 use tracing::{instrument, warn};
 
 use crate::core::commit_descriptors::{
@@ -31,6 +33,7 @@ use crate::core::commit_descriptors::{
 };
 use crate::core::config::get_main_branch_name;
 use crate::core::effects::{Effects, OperationType};
+use crate::core::eventlog::EventTransactionId;
 use crate::core::formatting::StyledStringBuilder;
 use crate::git::config::{Config, ConfigRead};
 use crate::git::oid::{make_non_zero_oid, MaybeZeroOid, NonZeroOid};
@@ -137,6 +140,30 @@ pub enum CherryPickFastError {
     },
 }
 
+/// Options for `Repo::amend_fast`
+pub enum AmendFastOptions {
+    /// Amend a set of paths from the current state of the working copy.
+    FromWorkingCopy {
+        /// The status entries for the files to amend.
+        status_entries: Vec<StatusEntry>,
+    },
+    /// Amend a set of paths from the current state of the index.
+    FromIndex {
+        /// The paths to amend.
+        paths: Vec<PathBuf>,
+    },
+}
+
+impl AmendFastOptions {
+    /// Returns whether there are any paths to be amended.
+    pub fn is_empty(&self) -> bool {
+        match &self {
+            AmendFastOptions::FromIndex { paths } => paths.is_empty(),
+            AmendFastOptions::FromWorkingCopy { status_entries } => status_entries.is_empty(),
+        }
+    }
+}
+
 /// A snapshot of all the positions of references we care about in the repository.
 #[derive(Debug)]
 pub struct RepoReferencesSnapshot {
@@ -198,6 +225,13 @@ impl Repo {
     /// is bare (has no working copy), returns `None`.
     pub fn get_working_copy_path(&self) -> Option<&Path> {
         self.inner.workdir()
+    }
+
+    /// Get the index file for this repository.
+    pub fn get_index(&self) -> eyre::Result<Index> {
+        Ok(Index {
+            inner: self.inner.index()?,
+        })
     }
 
     /// Get the configuration object for the repository.
@@ -518,6 +552,29 @@ Either create it, or update the main branch setting by running:
         Ok(Some(Diff { inner: diff }))
     }
 
+    /// Returns the set of paths currently staged to the repository's index.
+    pub fn get_staged_paths(&self) -> eyre::Result<HashSet<PathBuf>> {
+        let head_commit_oid = match self.get_head_info()?.oid {
+            Some(oid) => oid,
+            None => eyre::bail!("No HEAD to check for staged paths"),
+        };
+        let head_commit = self.find_commit_or_fail(head_commit_oid)?;
+        let head_tree = self.find_tree_or_fail(head_commit.get_tree()?.get_oid())?;
+
+        let diff = self.inner.diff_tree_to_index(
+            Some(&head_tree.inner),
+            Some(&self.get_index()?.inner),
+            None,
+        )?;
+        let paths = diff
+            .deltas()
+            .into_iter()
+            .flat_map(|delta| vec![delta.old_file().path(), delta.new_file().path()])
+            .flat_map(|p| p.map(PathBuf::from))
+            .collect();
+        Ok(paths)
+    }
+
     /// Get the file paths which were added, removed, or changed by the given
     /// commit.
     ///
@@ -608,6 +665,55 @@ Either create it, or update the main branch setting by running:
         } else {
             Ok(true)
         }
+    }
+
+    /// Returns the current status of the repo index and working copy.
+    pub fn get_status(
+        &self,
+        git_run_info: &GitRunInfo,
+        event_tx_id: Option<EventTransactionId>,
+    ) -> eyre::Result<Vec<StatusEntry>> {
+        let output = git_run_info.run_silent(
+            self,
+            event_tx_id,
+            &["status", "--porcelain=v2", "--untracked-files=no", "-z"],
+        )?;
+
+        let not_null_terminator = |c: &u8| *c != 0_u8;
+        let mut statuses = Vec::new();
+        let mut status_bytes = output.into_iter().peekable();
+
+        // Iterate over the status entries in the output.
+        // This takes some care, because NUL bytes are both used to delimit
+        // between entries, and as a separator between paths in the case
+        // of renames.
+        // See https://git-scm.com/docs/git-status#_porcelain_format_version_2
+        while let Some(line_prefix) = status_bytes.peek() {
+            let line = match line_prefix {
+                // Ordinary change entry.
+                b'1' => {
+                    let line = status_bytes
+                        .by_ref()
+                        .take_while(not_null_terminator)
+                        .collect_vec();
+                    line
+                }
+                // Rename or copy change entry.
+                b'2' => {
+                    let mut line = status_bytes
+                        .by_ref()
+                        .take_while(not_null_terminator)
+                        .collect_vec();
+                    line.push(0_u8); // Persist first null terminator in the line.
+                    line.extend(status_bytes.by_ref().take_while(not_null_terminator));
+                    line
+                }
+                _ => eyre::bail!("unknown status line prefix: {}", line_prefix),
+            };
+            let entry = line.as_slice().try_into()?;
+            statuses.push(entry);
+        }
+        Ok(statuses)
     }
 
     /// Create a new reference or update an existing one.
@@ -957,6 +1063,17 @@ Either create it, or update the main branch setting by running:
         }
     }
 
+    /// Like `find_tree`, but raises a generic error if the commit could not
+    /// be found.
+    #[instrument]
+    pub fn find_tree_or_fail(&self, oid: NonZeroOid) -> eyre::Result<Tree> {
+        match self.find_tree(oid) {
+            Ok(Some(tree)) => Ok(tree),
+            Ok(None) => eyre::bail!("Could not find tree with OID: {:?}", oid),
+            Err(err) => Err(err),
+        }
+    }
+
     /// Write the provided in-memory index as a tree into Git`s object database.
     /// There must be no merge conflicts in the index.
     #[instrument]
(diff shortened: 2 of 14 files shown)


src/git/repo.rs
```diff
@@ -1049,6 +1086,17 @@ impl Index {
             file_mode: entry.mode,
         })
     }
+
+    pub fn update_all(&mut self, pathspecs: &[&str]) -> eyre::Result<()> {
+        self.inner.update_all(pathspecs, None)?;
+        Ok(())
+    }
+
+    pub fn write_tree(&mut self) -> eyre::Result<NonZeroOid> {
```
arxanas, line 1095: We can't afford to write the entire index as a tree for performance reasons. To test performance, try running on https://github.com/mozilla/gecko-dev. You can also add a benchmark to `benches.rs` to confirm performance differences between implementations.

See the commentary and implementation for `Repo::cherry_pick_fast`. You have to take an existing tree and then augment it with only the changed paths, to ensure that we don't spend a lot of time doing unnecessary disk I/O:

https://github.com/arxanas/git-branchless/blob/331b7cf2a37d00a3d54ba5df00f3d702aa7b3948/src/git/repo.rs#L774-L789

src/git/repo.rs
```diff
@@ -965,6 +984,24 @@ Either create it, or update the main branch setting by running:
             .map_err(wrap_git_error)?;
         Ok(make_non_zero_oid(oid))
     }
+
+    /// Returns whether the repo currently has changes staged to the index.
+    pub fn has_staged_changes(&self) -> eyre::Result<bool> {
+        let mut opts = StatusOptions::new();
+        opts.include_untracked(true);
```
arxanas, line 991: I don't think we should include untracked files for the amend operation in general, as they may be very large or contain sensitive information. And in any event, untracked files can't contribute to the staged changes, can they?

src/commands/amend.rs
```diff
@@ -0,0 +1,63 @@
...
+        Some(oid) => oid,
+        None => eyre::bail!("No HEAD to amend to"),
+    };
+    let head_commit = repo.find_commit_or_fail(head_oid)?;
+
+    let mut index = repo.get_index()?;
+    // If the repo has staged changes, only commit those. If there
+    // are no staged changes, update all files from the working tree.
+    if !repo.has_staged_changes()? {
+        index.update_all(&["*"])?;
```
arxanas, line 39: We can't use `index.update_all` because `libgit2` won't use the `fsmonitor` system in use for large repositories. You really do have to shell out to a Git command in order to ensure that you use it. If you call `git status`, it will summarize both staged and unstaged changes in the same operation, and it will also update the index on disk, so you won't need to update it yourself.

tests/command/test_amend.rs
```diff
@@ -0,0 +1,193 @@
...
+#[test]
+fn test_amend_with_children() -> eyre::Result<()> {
+    let git = make_git()?;
+
+    if !git.supports_committer_date_is_author_date()? {
+        return Ok(());
+    }
+
+    git.init_repo()?;
+    git.run(&["config", "branchless.restack.preserveTimestamps", "true"])?;
```
arxanas, line 14: `branchless.restack.preserveTimestamps` should be the default as of #167, so no need to set it explicitly here.

tests/command/test_amend.rs
```diff
@@ -0,0 +1,193 @@
...
+        No abandoned commits to restack.
+        No abandoned branches to restack.
+        branchless: running command: <git-executable> checkout 3b98a960e6ebde39a933c25413b43bce8c0fd128
+        O f777ecc9 (master) create initial.txt
+        |
+        @ 3b98a960 create test1.txt
+        "###);
+    }
+
+    // Amend should only update existing files.
```
arxanas, line 165: I think by "existing" you mean "tracked"?

arxanas (changes requested):

arxanas:
@epage Targeting a commit other than `HEAD` makes sense (git-revise lets you do that), although it might not commute with the commits sense then. I don't know what git-revise does in that case. Personally, I think fixup commits are a low-level artifact used because of how the Git rebase implementation works. We shouldn't generate fixup commits now that we can target commits for editing in a more direct way.

epage:
> @epage Targeting a commit other than `HEAD` makes sense (git-revise lets you do that), although it might not commute with the commits sense then. I don't know what git-revise does in that case. Personally, I think fixup commits are a low-level artifact used because of how the Git rebase implementation works. We shouldn't generate fixup commits now that we can target commits for editing in a more direct way.

Granted, I've not given `git-branchless` a spin but I have a hard time imagining it smoothing out the rebase experience to make it so I never have to use fixup commits.  Part of it is a mental model / workflow.  For example, today I was fixing bugs in `git-stack` and I created multiple commits from it after I got it working, using `git add -p`.  I stepped through each set of changes I staged and commited it, deciding whether it should be a fixup or regular commit as I went.

The other reason I think fixups are useful is they can provide a nice unit of work for showing reviewers how you responded to feedback.  Some tools handle showing differences between force pushes well, most don't.  I just wish more merge systems auto-squashed them.

arxanas:
@epage In my workflow, I sometimes still make fixup commits, but I don't mark them as `fixup!` or use autosquashing. I assume a commit with `fixup!` in the commit message is what you meant by "fixup commit"? I just use interactive rebase to squash them into the appropriate commits. In most cases, I just run `git checkout`/`git commit --amend`/`git restack` to update a specific commit in a stack.

Since `git-branchless` tracks the rewrite history of commits over time, it would be possible to synthesize fixup commits as part of pushing to the code review tool, without keeping them around locally. (I've also been thinking about this for Github pull requests: instead of including all of the commits from the base pull request in a dependent pull request, you can synthesize a single aggregate commit containing all of the base commits, which might help the reviewer.) But in any event, integration with code review tools is a long while off.

Personally, I don't write meaningful messages for the fixup commits themselves — do you?

epage:
Yes, I am referring to `fixup!`.  I don't add meaningful messages to them but reviewers can look at these smaller chunks and general understand what review comment it is addressing

If this isn't of interest for the `git-branchless` workflow, thats, fine, just had an idea for generalizing the command that I thought I'd share.

src/git/repo.rs
bcongdon: That makes sense. The one thing I'm still unclear about is how to create `Oid`s for files that are modified in the working tree, but not staged. This is why I thought I needed to call `update_all` -- so that the index would stage those changes and make `Oid`s available.

My assumption is that if a change is staged, then it should have an index entry with an Oid. But, I'm not sure how to get from unstaged -> staged without interacting with the index.

(I'm not super familiar with Git internals, so apologies if this question is misguided.)

bcongdon: Ah, on the specific question of file -> OID, it looks like `Repo::blob_path` would work

src/commands/amend.rs
```diff
@@ -0,0 +1,70 @@
...
+    let mut event_log_db = EventLogDb::new(&conn)?;
+
+    let head_oid = match repo.get_head_info()?.oid {
+        Some(oid) => oid,
+        None => eyre::bail!("No HEAD to amend to"),
+    };
+    let head = repo.find_commit_or_fail(head_oid)?;
+
+    let mut index = repo.get_index()?;
+    let tree_oid = index.update_all(&["*"])?;
```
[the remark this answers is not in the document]
arxanas: > you mean executing git as a subprocess, and then parsing the `status` output

Yes, exactly.

> If we do that, and get a list of files that need to be updated in the index, can we call `update_all` on that list, or would that still have performance issues?

We could and it would work in a performant way, but we won't need to, because `git status` updates the index file on disk as a side-effect.

src/git/repo.rs
arxanas: We don't want to stage the changes as part of the commit process, or use an `Index` structure to ultimately make the commit. To get the OIDs for the changed files, you can use `Repo::blob_path` like you mentioned. To make the commit object, you'll want to take the current commit's tree and modify it by inserting the changed entries. You can use this crate's utility function `tree::hydrate_tree` to do that. Then pass the tree to this crate's `Repo::create_commit`.

bcongdon:
I took another pass at this, incorporating the review comments.

- This now uses `git status` to check the status of the index and working tree.
- This now uses a `dehydrate_tree` -> make modifications -> `hydrate_tree` approach to creating the tree for the amend commit
- This now distinguishes between operations that affect the working tree vs. the index
- Handles deletions and renames

src/git/run.rs
```diff
@@ -173,6 +173,15 @@ impl GitRunInfo {
...
         let args = {
             let mut result = vec!["-C", repo_path];
+            if let Some(working_copy_path) = repo.get_working_copy_path() {
+                let working_copy_path = working_copy_path.to_str().ok_or_else(|| {
+                    eyre::eyre!(
+                        "Path to Git repo could not be converted to UTF-8 string: {:?}",
+                        working_copy_path
+                    )
+                })?;
+                result.extend(vec!["--work-tree", working_copy_path]);
```
bcongdon, line 183: I needed to add this, else `git status` fails with "fatal: This operation must be run in a work tree"

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+            .get_paths_touched_by_commit(parent_commit)?
+            .ok_or_else(|| {
+                eyre::eyre!(
+                    "Could not get paths touched by commit: {:?}",
+                    &parent_commit
+                )
+            })?
+            .into_iter()
+            .collect_vec();
+        let changed_paths = {
```
arxanas, line 1048: Since you're not using `parent_commit_pathbufs` elsewhere in this function, you can collect it directly into a `HashSet<PathBuf>` and then add the elements and convert it into a `Vec`. (Actually, the interface to `dehydrate_commit` could be refactored, because I think all of the call-sites end up taking `HashSet`s anyways.

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+            AmendFastOptions::FromIndex { paths } => {
+                let index = self.get_index()?;
+                paths
+                    .iter()
+                    .filter_map(|path| match index.get_entry(path) {
+                        Some(IndexEntry {
+                            oid: MaybeZeroOid::Zero,
+                            ..
+                        }) => {
+                            warn!("index entry for {:?} was zero", path);
```
arxanas, line 1091: Use `warn!(?path, "index entry was zero")` to get the structured logging equivalent, which can be analyzed programmatically.

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+
+        let repo_path = self
+            .get_working_copy_path()
+            .expect("repo working copy path");
+        let amended_oids: HashMap<PathBuf, Option<NonZeroOid>> = match opts {
+            AmendFastOptions::FromWorkingTree { paths } => paths
+                .iter()
+                .map(|path| {
+                    let file_path = &repo_path.join(path);
+                    if file_path.exists() {
```
arxanas, line 1071: To ameliorate race conditions slightly, instead of checking if the file path `exist` and then attempting to read it, instead try to do the `blob_path` unconditionally and check for an error with code `NotFound` (or whatever error occurs when the file is not found on disk), and return `None` in that case. There are some other examples of this pattern such as in `Repo::find_commit`.

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+                        None => Some((path.clone(), None)),
+                    })
+                    .collect::<HashMap<_, _>>()
+            }
+        };
+
+        let amended_index_entries: HashMap<PathBuf, Option<(NonZeroOid, i32)>> = changed_paths
+            .into_iter()
+            .map(|changed_path| {
+                let value = if let Some(amended_oid) = amended_oids.get(changed_path) {
```
arxanas, line 1107: Instead of using `if let Some(x) = expr { ... } else { ... }`, use `match expr { Some(x) => { ... }, None => { ... } }`. If the proliferation of `Option`s makes this function confusing, then create an `enum` scoped to this function with elements like `Present(NonZeroOid)` and `Absent` (but pick good names).

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+                AmendFastOptions::FromIndex { paths } => paths,
+                AmendFastOptions::FromWorkingTree { paths } => paths,
+            };
+            result.extend(paths.iter().map(|path| path.as_path()));
+            result.iter().cloned().collect_vec()
+        };
+
+        let dehydrated_parent =
+            self.dehydrate_commit(parent_commit, changed_paths.as_slice(), true)?;
+        let dehydrated_parent_index = Index::from_tree(&dehydrated_parent.get_tree()?)?;
```
arxanas, line 1061: Why do we need to convert the parent commit's into a tree here? I suspect it will be slow, and we should be able to index into the tree's entries directly with `git2::Tree::get_path`.

src/git/repo.rs
```diff
@@ -1049,6 +1244,12 @@ impl Index {
             file_mode: entry.mode,
         })
     }
+
+    pub fn from_tree(tree: &Tree) -> eyre::Result<Self> {
```
arxanas, line 1248: As per above comment, I don't think we need this, and it would be good to omit from our interfaces since it's likely to be slow on large repositories.

src/git/repo.rs
```diff
@@ -1231,6 +1432,31 @@ impl<'repo> Commit<'repo> {
             (Some(parent_oid), Some(current_oid)) => Ok(Some(parent_oid != current_oid)),
         }
     }
+
+    /// Amend this existing commit.
+    /// Returns the OID of the resulting new commit.
+    #[instrument]
+    pub fn amend(
```
arxanas, line 1439: I would call this method `amend_commit` to make it clear what it's operating on.

src/git/repo.rs
```diff
@@ -1446,6 +1672,98 @@ impl<'repo> Branch<'repo> {
...
+        ' ' => FileStatus::Unmodified,
+        'M' => FileStatus::Modified,
+        'A' => FileStatus::Added,
+        'D' => FileStatus::Deleted,
+        'R' => FileStatus::Renamed,
+        'C' => FileStatus::Copied,
+        'U' => FileStatus::Unmerged,
+        '?' => FileStatus::Untracked,
+        '!' => FileStatus::Ignored,
+        _ => eyre::bail!("invalid status indicator: {}", status),
```
arxanas, line 1726: Let's just `warn!` in this case and use a sensible default (maybe `FileStatus::Untracked` or `FileStatus::Ignored`?).

src/git/repo.rs
```diff
@@ -1446,6 +1672,98 @@ impl<'repo> Branch<'repo> {
...
+    Ignored,
+}
+
+/// The status of a file in the repo.
+#[derive(Debug, PartialEq, Eq)]
+pub struct StatusEntry {
+    /// The status of the file in the index.
+    pub index_status: FileStatus,
+    /// The status of the file in the working tree.
+    pub work_tree_status: FileStatus,
```
arxanas, line 1697: Let's call this `working_copy_status`, as I think that's the terminology we use elsewhere in the project.

src/git/repo.rs
```diff
@@ -1446,6 +1672,98 @@ impl<'repo> Branch<'repo> {
...
+    /// The file path.
+    pub path: String,
+    /// The original path of the file (for renamed files).
+    pub orig_path: Option<String>,
```
arxanas, line 1701: Use `PathBuf`s here instead of `String`s.

src/git/repo.rs
```diff
@@ -1446,6 +1672,98 @@ impl<'repo> Branch<'repo> {
...
+        'C' => FileStatus::Copied,
+        'U' => FileStatus::Unmerged,
+        '?' => FileStatus::Untracked,
+        '!' => FileStatus::Ignored,
+        _ => eyre::bail!("invalid status indicator: {}", status),
+    };
+    Ok(status)
+}
+
+impl FromStr for StatusEntry {
```
arxanas, line 1731: File paths may be any sequence of non-null bytes, not just UTF-8-encoded strings. Let's parse those properly. (There are a few cases in the rest of the program where we operate on strings instead of `PathBuf`s, but those are due to libgit2 limitations, and they could be fixed in the future.)

To operate on bytes, you'll want to use the `&[u8]` type. I think a lot of iterator methods also end up working on byte sequences, such as `split`. You can also use `regex::bytes` if necessary.

I saw this comment in the documentation, which should make parsing the output from `git status` easier:

> There is also an alternate -z format recommended for machine parsing. In that format, the status field is the same, but some other things change. First, the -> is omitted from rename entries and the field order is reversed (e.g from -> to becomes to from). Second, a NUL (ASCII 0) follows each filename, replacing space as a field separator and the terminating newline (but a space still separates the status field from the first filename). Third, filenames containing special characters are not specially formatted; no quoting or backslash-escaping is performed.

src/git/run.rs
arxanas: Is this still true? I commented this out and the amend tests still passed.

src/git/repo.rs
```diff
@@ -965,6 +1022,144 @@ Either create it, or update the main branch setting by running:
...
+        };
+
+        let amended_index_entries: HashMap<PathBuf, Option<(NonZeroOid, i32)>> = changed_paths
+            .into_iter()
+            .map(|changed_path| {
+                let value = if let Some(amended_oid) = amended_oids.get(changed_path) {
+                    match amended_oid {
+                        Some(added_blob_oid) => {
+                            let file_path = repo_path.join(changed_path);
+                            let file_mode = if file_path.is_executable() {
```
arxanas, line 1111: Rather than read from disk again in a potentially racy way to determine if the file mode is executable, it would probably be better to use the `--porcelain=v2` format to `git status` (assuming that it's supported by Git v2.24, our lowest-supported version), which includes this information.

src/commands/amend.rs
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+        let paths = status
+            .iter()
+            .filter_map(|entry| match entry.work_tree_status {
+                FileStatus::Added
+                | FileStatus::Copied
+                | FileStatus::Deleted
+                | FileStatus::Modified
+                | FileStatus::Renamed => Some(entry.paths()),
+                _ => None,
```
arxanas, line 79: Use an exhaustive pattern match here, just in case something changes in the future.

src/commands/amend.rs
```diff
@@ -0,0 +1,134 @@
...
+    let conn = repo.get_db_conn()?;
+    let mut event_log_db = EventLogDb::new(&conn)?;
+
+    let head_oid = match repo.get_head_info()?.oid {
+        Some(oid) => oid,
+        None => eyre::bail!("No HEAD to amend to"),
+    };
+    let head_commit = repo.find_commit_or_fail(head_oid)?;
+
+    let status = repo.get_status(git_run_info)?;
```
arxanas, line 38: Calling `git status` may be an expensive operation because it has to walk the working copy, but accessing the index should be quick. Check for staged index entries and, if present, use those. Only if they're not present should you make a call to check `git status`.

arxanas (changes requested):
I think we're getting pretty close!

bcongdon (commented):
> I think we're getting pretty close!

I think so too -- also, I continue to appreciate the review comments. 👍

src/commands/amend.rs
```diff
@@ -0,0 +1,128 @@
...
+        new_commit_oid: amended_commit_oid.into(),
+    }])?;
+
+    // Reset the index
+    git_run_info.run(effects, Some(event_tx_id), &["reset"])?;
+    git_run_info.run(
+        effects,
+        Some(event_tx_id),
+        &["checkout", &amended_commit_oid.to_string()],
+    )?;
```
arxanas, line 121: What does calling `git reset` do here? And what does checking out the commit do? I thought the point of passing `HEAD` above was that it updated the reference for us, so we wouldn't have to check anything out ourselves.

Can we avoid calling `git reset` and `git checkout` here? Both of those require checking the working copy again, which slows down the operation. We can leave it to a future commit if you don't want to do it in this one.

src/git/repo.rs
```diff
@@ -516,6 +548,30 @@ Either create it, or update the main branch setting by running:
...
+        let diff = self.inner.diff_tree_to_index(
+            Some(&head_tree.inner),
+            Some(&self.get_index()?.inner),
+            None,
+        )?;
+        let paths = diff
+            .deltas()
+            .into_iter()
+            .map(|delta| vec![delta.old_file().path(), delta.new_file().path()])
+            .flatten()
```
arxanas, line 569: Can you also use `flat_map` here?

src/git/repo.rs
```diff
@@ -608,6 +664,51 @@ Either create it, or update the main branch setting by running:
         }
     }
 
+    /// Returns the current status of the repo index and working copy.
+    pub fn get_status(&self, git_run_info: &GitRunInfo) -> eyre::Result<Vec<StatusEntry>> {
+        let output = git_run_info.run_silent(
+            self,
+            None,
```
arxanas, line 671: Pass the event transaction ID from the caller through this parameter as well. (It shouldn't really be relevant for now, since this operation shouldn't affect the logical state of the repository, but maybe one day we'll use this information for profiling or something?)

src/git/repo.rs
```diff
@@ -608,6 +664,51 @@ Either create it, or update the main branch setting by running:
...
+            &["status", "--porcelain=v2", "--untracked-files=no", "-z"],
+        )?;
+
+        let not_null_terminator = |c: &&u8| **c != 0_u8;
+        let mut statuses = Vec::new();
+        let mut status_bytes = output.iter().peekable();
```
arxanas, line 677: Use `output.into_iter()` here so that we can get rid of a layer of references and deal with `&u8` instead of `&&u8`. Then you should also be able to just write `StatusEntry::from_bytes(&line)`.

src/git/repo.rs
```diff
@@ -965,6 +1077,134 @@ Either create it, or update the main branch setting by running:
...
+                        // Try to create a new blob OID based on the current on-disk
+                        // contents of the file in the working copy.
+                        match self.inner.blob_path(file_path) {
+                            Ok(oid) => Ok((
+                                path,
+                                Some((make_non_zero_oid(oid), entry.working_copy_file_mode.into())),
+                            )),
+                            // If the file doesn't exist, it needs to be explicitly marked as
+                            // such, as a tombstone to override if the file exists in the parent tree.
+                            Err(ref err) if err.code() == git2::ErrorCode::NotFound => {
```
arxanas, line 1142: Why do we need to capture `err` by `ref` here?

src/git/repo.rs
```diff
@@ -965,6 +1077,134 @@ Either create it, or update the main branch setting by running:
...
+            .map(|path| path.as_path())
+            .collect_vec();
+
+        let dehydrated_parent =
+            self.dehydrate_commit(parent_commit, changed_paths.as_slice(), true)?;
+        let dehydrated_parent_tree = dehydrated_parent.get_tree()?;
+
+        let repo_path = self
+            .get_working_copy_path()
+            .expect("repo working copy path");
```
arxanas, line 1126: Use `ok_or_else` here instead, so we can get a better stack trace.

src/git/repo.rs
```diff
@@ -1446,6 +1711,152 @@ impl<'repo> Branch<'repo> {
...
+    pub fn paths(&self) -> Vec<PathBuf> {
+        let mut result = vec![self.path.clone()];
+        if let Some(orig_path) = &self.orig_path {
+            result.push(orig_path.clone());
+        }
+        result
+    }
+}
+
+fn parse_file_status(status: u8) -> eyre::Result<FileStatus> {
```
arxanas, line 1756: It would be best to implement `From`/`TryFrom` here.

src/git/repo.rs
```diff
@@ -1446,6 +1711,152 @@ impl<'repo> Branch<'repo> {
...
+            FileStatus::Untracked
+        }
+    };
+    Ok(status)
+}
+
+// Parses the string representation of a filemode for a status entry.
+// Git only supports a small subset of Unix octal file mode permissions.
+// See http://git-scm.com/book/en/v2/Git-Internals-Git-Objects
+fn parse_file_mode(file_mode: &str) -> eyre::Result<FileMode> {
```
arxanas, line 1778: Likewise. There's also a `FromStr` trait you could increment for specifically the case of converting from `&str`.

src/git/repo.rs
```diff
@@ -1446,6 +1711,152 @@ impl<'repo> Branch<'repo> {
...
+        _ => eyre::bail!("unknown file mode: {}", file_mode),
+    };
+    Ok(file_mode)
+}
+
+impl StatusEntry {
+    /// Parses an entry of the git porcelain v2 status format.
+    /// See https://git-scm.com/docs/git-status#_porcelain_format_version_2
+    #[instrument]
+    fn from_bytes(line: &[u8]) -> eyre::Result<StatusEntry> {
```
arxanas, line 1795: Maybe all this complexity indicates that we should use a bytestring regex instead?

src/git/run.rs
arxanas: I see. `repo_path` here is passed a directory like `foo/bar/.git`, which is not legal to run `git status` inside of. I think the right thing to do is to just set `-C` to `repo.get_working_copy_path()` if it's present, or `repo_path` if not.

arxanas (approved):
I think this is good overall, feel free to merge after addressing feedback!

src/git/run.rs
arxanas: I suppose it's also a mistake that the process invocation can return error without it being detected. The caller should have to opt into ignoring the exit code (which is the right thing to do in certain cases, such as when running user-defined hooks, but not always).

src/commands/amend.rs
```diff
@@ -0,0 +1,128 @@
...
+    move_options: &MoveOptions,
+) -> eyre::Result<isize> {
+    let now = SystemTime::now();
+    let repo = Repo::from_current_dir()?;
+    let conn = repo.get_db_conn()?;
+    let mut event_log_db = EventLogDb::new(&conn)?;
+
+    let head_oid = match repo.get_head_info()?.oid {
+        Some(oid) => oid,
+        None => eyre::bail!("No HEAD to amend to"),
```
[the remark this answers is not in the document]
bcongdon: Done

src/commands/amend.rs
```diff
@@ -0,0 +1,128 @@
...
+    };
+    let head_commit = repo.find_commit_or_fail(head_oid)?;
+
+    let index = repo.get_index()?;
+    if index.has_conflicts() {
+        writeln!(
+            effects.get_output_stream(),
+            "Cannot amend, because there are unresolved merge conflicts."
```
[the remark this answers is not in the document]
bcongdon: Done

src/commands/amend.rs
```diff
@@ -0,0 +1,128 @@
...
+        return Ok(1);
+    }
+
+    let staged_index_paths = repo.get_staged_paths()?;
+    let opts = if !staged_index_paths.is_empty() {
+        let dirty_working_tree = repo.has_changed_files(effects, git_run_info)?;
+        if dirty_working_tree {
+            writeln!(
+                effects.get_output_stream(),
+                "Found staged files in the index. Unstaged changes will not be amended."
```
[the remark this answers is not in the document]
bcongdon: > let's remove this message and add a summary message to immediately after the amend operation took place. It should also contain some brief summary information. I think it would be good to say "Amended with X uncommitted changes"/"Amended with X staged changes (Y uncommitted changes were not amended)."

Great idea; done. Though, I didn't implement the `(Y uncommitted changes were not amended)` portion -- and left it ambiguous as `(Some uncommitted changes were not amended)`, since I think getting the number of modified files in the working copy would require a `git_status` run. I left a TODO to consider adding that in the future.

src/commands/amend.rs
```diff
@@ -0,0 +1,128 @@
...
+                | FileStatus::Unmodified
+                | FileStatus::Untracked => false,
+            })
+            .collect_vec();
+        AmendFastOptions::FromWorkingTree {
+            status_entries: entries_to_amend,
+        }
+    };
+    if opts.is_empty() {
+        writeln!(effects.get_output_stream(), "Nothing to amend.")?;
```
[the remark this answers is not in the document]
bcongdon: Done

src/git/run.rs
bcongdon: > I think the right thing to do is to just set -C to repo.get_working_copy_path() if it's present, or repo_path if not

Done. Also filed #200

src/git/repo.rs
bcongdon: Done

src/git/repo.rs
bcongdon: Done

src/git/repo.rs
bcongdon: 👍

src/git/repo.rs
bcongdon: Done

src/git/repo.rs
bcongdon: Whoops, I think this was a copy/paste fail.

src/git/repo.rs
bcongdon: Done

src/git/repo.rs
bcongdon: Done, though I had to create a wrapper type for `FileMode` (which, is probably for the best anyways, to prevent the proliferation of `git2` types)

src/git/repo.rs
bcongdon: Switched to `regex::bytes`. The regex is fairly complicated, so I broke it out over multiple lines.

src/commands/amend.rs
bcongdon: It looks like the `git checkout` is no longer necessary -- I think I added that when I was experimenting with not updating `HEAD` as part of the amend.

`git reset` is used to reset the index following the amend. Without this, when you amend something from the working copy, the index retains a "shadow" diff following the amend. The index and working copy contain inverse diffs of each other that get cancelled out by a `reset`. It's rather strange (which likely indicates I'm missing something here...)

To repro this, remove the checkout line and run:

```
$ echo "test" > test.txt
$ git commit -am "test"
$ echo "updated contents" > foo.txt
$ git amend
$ git status
HEAD detached from afdaf9a
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   foo.txt

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   foo.txt

$ git diff
diff --git a/foo.txt b/foo.txt
index 014fd71..1385f26 100644
--- a/foo.txt
+++ b/foo.txt
@@ -1 +1 @@
-test
+updated contents

$ git diff --cached
diff --git a/foo.txt b/foo.txt
index 1385f26..014fd71 100644
--- a/foo.txt
+++ b/foo.txt
@@ -1 +1 @@
-updated contents
+test
```

---


Note that this _doesn't_ happen if you amend a change that is staged to the index -- only when you amend something unstaged from the working copy.

bcongdon: Moved this conversation to #201

PR #187, opened 2021-10-31
feat(navigation): Improve warnings for navigation commands when checkout fails

### LINKED ISSUE #182: Improve "checkout" error when checking out the selected commit fails
### Description of the bug

Similar to #181, but with the interactive checkout command.

If you have a dirty working copy that prevents you from checking out the selected commit, the error message isn't clear enough.

### Expected behavior

A "loud" error message should be displayed, and we should skip printing the smartlog if checkout fails.

### Actual behavior

The `git checkout` output is in grey, and doesn't indicate failure

master

### Version of `rustc`

_No response_

bcongdon:
Fixes #182.

- Adds warnings when `next`/`prev`/`checkout` fails to check out the requested commit.
- Adds a warning to `next` when the user requests advancing by more than the number of available children.

### DIFF
--- a/src/commands/navigation.rs
+++ b/src/commands/navigation.rs
@@ -5,11 +5,11 @@ use std::fmt::Write;
 use eden_dag::DagAlgorithm;
 use tracing::instrument;
 
-use crate::commands::smartlog::{make_smartlog_graph, smartlog};
+use crate::commands::smartlog::make_smartlog_graph;
 use crate::core::config::get_next_interactive;
 use crate::core::eventlog::{EventLogDb, EventReplayer};
-use crate::core::formatting::printable_styled_string;
-use crate::git::{sort_commit_set, CommitSet, Dag, GitRunInfo, NonZeroOid, Repo};
+use crate::core::formatting::{printable_styled_string, Pluralize};
+use crate::git::{check_out_commit, sort_commit_set, CommitSet, Dag, GitRunInfo, NonZeroOid, Repo};
 use crate::tui::prompt_select_commit;
 use crate::tui::Effects;
 
@@ -20,19 +20,11 @@ pub fn prev(
     git_run_info: &GitRunInfo,
     num_commits: Option<isize>,
 ) -> eyre::Result<isize> {
-    let exit_code = match num_commits {
-        None => git_run_info.run(effects, None, &["checkout", "HEAD^"])?,
-        Some(num_commits) => git_run_info.run(
-            effects,
-            None,
-            &["checkout", &format!("HEAD~{}", num_commits)],
-        )?,
+    let target = match num_commits {
+        None => "HEAD^".into(),
+        Some(num_commits) => format!("HEAD~{}", num_commits),
     };
-    if exit_code != 0 {
-        return Ok(exit_code);
-    }
-    smartlog(effects, &Default::default())?;
-    Ok(0)
+    check_out_commit(effects, git_run_info, None, &target)
 }
 
 /// Some commits have multiple children, which makes `next` ambiguous. These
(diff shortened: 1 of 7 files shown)


src/commands/navigation.rs
```diff
@@ -79,15 +73,33 @@ fn advance(
...
-                // It would also make sense to issue an error here, rather than
-                // silently stop going forward commits.
+                writeln!(
+                    effects.get_output_stream(),
+                    "{}",
+                    printable_styled_string(
+                        glyphs,
+                        StyledString::styled(
+                            format!(
+                                "No more child commits to go to after traversing {}.",
```
arxanas, line 96: This is just a warning, not an error, right? Does it still make sense to use red for this message? I would expect yellow.

src/commands/navigation.rs
```diff
@@ -192,11 +198,33 @@ pub fn checkout(effects: &Effects, git_run_info: &GitRunInfo) -> eyre::Result<is
...
+    let result = git_run_info.run(effects, None, &["checkout", &target.to_string()])?;
+    smartlog(effects, &Default::default())?;
+    if result != 0 {
+        writeln!(
+            effects.get_output_stream(),
+            "{}",
+            printable_styled_string(
+                effects.get_glyphs(),
+                StyledString::styled(
+                    format!("Failed to checkout commit {}.", target.to_string(),),
```
arxanas, line 223: This should read `Failed to check out commit: {}`:

- The verb form is "check out", and the noun form is "checkout". It's unfortunate that the Git command is called `checkout`; even in its own documentation, it uses the phrase "check out" for the verb.
- Likewise, rename this function to `check_out_commit`.
- In the event that the user wants to copy-and-paste the commit identifier, it's best to not include the `.` as part of the word, so that their selection doesn't accidentally copy the extra `.`.

Can you also update `execute::check_out_updated_head` to use this function?

src/commands/navigation.rs
```diff
@@ -192,11 +198,33 @@ pub fn checkout(effects: &Effects, git_run_info: &GitRunInfo) -> eyre::Result<is
...
     }
 }
+
+fn checkout_commit<T>(
+    effects: &Effects,
+    git_run_info: &GitRunInfo,
+    target: T,
+) -> eyre::Result<isize>
+where
+    T: ToString,
```
arxanas, line 212: You call `to_string` more than once in this function, so you're allocating twice unnecessarily. It would be better to accept a `&str` and have the caller allocate the `String` only once (or zero times if it's a static string!).

As a side note, you can write this trait bound in any of the following ways:

- `foo(s: impl ToString)`
- `foo<T: ToString>(s: T)`
- `foo<T>(s: T) where T: ToString`

For this project, you should prefer them in that order, and use the more complex versions only when necessary to express more complicated constraints.

src/commands/navigation.rs
```diff
@@ -79,15 +73,33 @@ fn advance(
             .children(CommitSet::from(current_oid))?
             .difference(&dag.obsolete_commits);
         let children = sort_commit_set(repo, dag, &children)?;
+
+        let children_pluralize = Pluralize {
```
arxanas, line 77: 👍

arxanas (changes requested):

src/commands/navigation.rs
arxanas: This function should also take an explicit `Option<EventTransactionId>` and forward it to `run`. If the checkout is part of a larger overall operation (such as a rebase/move/undo),  then it should be associated with the overall operation, so that the operation can be rendered/undone as a single unit.

bcongdon: 👍 
(Good thought about removing the ".")

src/commands/navigation.rs
bcongdon: Done. (And I'll keep the trait style in mind for the future)

arxanas (approved):

src/core/rewrite/execute.rs
```diff
@@ -258,10 +260,11 @@ pub fn check_out_updated_head(
         }
     };
 
-    let result = git_run_info.run(
+    let result = check_out_commit(
         effects,
+        git_run_info,
         Some(event_tx_id),
-        &[OsStr::new("checkout"), &checkout_target],
+        &checkout_target.to_string_lossy(),
```
arxanas, line 267: Ah, I missed this. Branch names are backed by filenames, so we want to use `OsStr`/`OsString` whenever possible to reference them. You can use `impl AsRef<OsStr>` to accept either a `str` or an `OsStr`, rather than potentially mis-encoding the branch name. I fixed this in #192.

PR #189, opened 2021-11-03
feat(navigation): show smartlog only if checkout succeeded

arxanas:


arxanas:
cc @bcongdon, I was caught by surprise and missed the fact that the checkout failed after a `git move`, because the smartlog was displayed, so I assumed that the operation succeeded.

bcongdon:
Nice. Eliding the smartlog if the checkout fails seems like a good change

PR #198, opened 2021-11-05
feat(checkout): render additional commit descriptors in interactive checkout

arxanas:


tests/command/test_navigation.rs
```diff
@@ -636,3 +636,40 @@ fn test_navigation_failed_to_check_out_commit() -> eyre::Result<()> {
...
+    git.run(&["checkout", "--detach", "master"])?;
+    git.commit_file("test3", 3)?;
+
+    run_in_pty(
+        &git,
+        &["branchless", "checkout"],
+        &[
+            PtyAction::WaitUntilContains("> "),
+            PtyAction::Write("master"),
+            PtyAction::WaitUntilContains("> master"),
```
bcongdon, line 658: (If the tests are still failing) You may need to change this to `> 62fc20d2`. This is intended to match what `skim` thinks is the selected item -- e.g. what it will return when the carriage return is entered.

Screen Shot 2021-11-05 at 6 37 38 AM

In this case, `> master` just verifies that the query was entered successfully, but there is a race condition in skim actually updating the displayed results. (You'd think key presses and the displayed results would be in ~lockstep, but in my experience there's a slight delay)

arxanas: Thanks, that fixed it! That probably saved me a lot of time :)

PR #199, opened 2021-11-05
feat(checkout): allow seeding interactive checkout with an initial query

arxanas:


arxanas:
cc @bcongdon re this stack

bcongdon:
Nice, I was thinking the other day that this would be a helpful addition 👍

arxanas:
One other idea is to implement an "I'm feeling lucky" style search, where if there's only one result for the query, then it's automatically selected and checked out to.

On one hand, it expedites the workflow of using branches or a different commit identifier (in my case, Phabricator revision numbers of the form `D123`). But on the other hand, it means that the checkout operation is only sometimes interactive, which may be confusing in practice.

PR #203, opened 2021-11-06
fix(run): Default to treating git failures as errors

### LINKED ISSUE #200: `GitRunInfo::run_slient` should return an error for non-zero exit codes (by default)
In `run.rs`, `GitRunInfo::run` and and `GitRunInfo::run_silent` do not return an error if git exits with a nonzero code. This is ~fine for existing usage, but in some circumstances it's useful to know if the git command failed or not.

As an example, when writing the harness around `git status`, we weren't able to (easily) detect cases where `status` failed.

We should consider adding an option to the `run` methods so that the caller has to opt-in to _not_ getting an error if the exit code is nonzero, similar to `testing::Git`

---
Original context:

> I suppose it's also a mistake that the process invocation can return error without it being detected. The caller should have to opt into ignoring the exit code (which is the right thing to do in certain cases, such as when running user-defined hooks, but not always).

_Originally posted by @arxanas in https://github.com/arxanas/git-branchless/pull/170#r743447758_

bcongdon:
Fixes #200.

- `run_silent` now returns stderr, stdout, and exit_code
- `run_silent` now defaults to returning an error if git exits with a nonzero code (though, this behavior can be disabled with an option)

### DIFF
--- a/src/git/run.rs
+++ b/src/git/run.rs
@@ -48,6 +48,37 @@ impl std::fmt::Debug for GitRunInfo {
     }
 }
 
+pub struct GitRunOpts {
+    /// If set, a non-zero exit code will be treated as an error.
+    treat_git_failure_as_error: bool,
+}
+
+impl Default for GitRunOpts {
+    fn default() -> Self {
+        Self {
+            treat_git_failure_as_error: true,
+        }
+    }
+}
+
+pub struct GitRunResult {
+    pub exit_code: i32,
+    pub stdout: Vec<u8>,
+    pub stderr: Vec<u8>,
+}
+
+impl std::fmt::Debug for GitRunResult {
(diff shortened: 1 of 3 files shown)


src/git/run.rs
```diff
@@ -210,11 +256,13 @@ impl GitRunInfo {
         repo: &Repo,
         event_tx_id: Option<EventTransactionId>,
         args: &[S],
-    ) -> eyre::Result<Vec<u8>> {
+        opts: Option<GitRunOpts>,
```
arxanas, line 259: I think it's better to force `opts: GitRunOpts` here and have the user pass `Default::default` where they intend to use the default options.

arxanas (approved):

arxanas:
Merge at your leisure!

src/git/run.rs
bcongdon: Fixed 👍

PR #205, opened 2021-11-07
cleanup: remove `man-pages` feature

arxanas:


martinvonz:
> Why did you remove it?

The commit message says this:
```
This feature added a lot of extra complexity and didn't work too well. We might want to revisit it in the future, once Rust has better support for installing man-pages.
```

I don't know if you had missed that or if you're looking for additional justification.

arxanas:
Yeah, it was a lot of complexity. I particularly didn't like that we build a bunch of duplicate executables just to get man-pages for them (although I think it can be avoided). Also, `clap_mangen` exists now to accomplish the man-page generation portion of it.

Issue #219, opened 2021-11-18
Don't overwrite user aliases upon installation


arxanas:
See #218. Although we no longer overwrite aliases, apparently our config file takes precedence over the user one, so we still end up overwriting them in practice. (But at least the aliases can be safely removed later when uninstalling.)

One further complication is that we may want to update git-branchless aliases which already exist upon a fresh run of `init`, just not user aliases.

typesanitizer:
I was thinking the way to fix this would be to include other configuration files after including `branchless/config`.

```
[include]
        path = branchless/config
        path = "$PREFIX/etc/gitconfig" 
        path = "$XDG_CONFIG_HOME/git/config"
        path = "$HOME/.gitconfig"
```

(I'm not sure whether the above will work for Windows though; I was just following the `git-config` docs.)

Does that sound like a reasonable fix? Not entirely sure what complication you're referring to in:

> One further complication is that we may want to update git-branchless aliases which already exist upon a fresh run of init, just not user aliases.

IIUC this should work without problems if you add/delete aliases in `.git/branchless/config`, without having to care about which aliases are user defined, so long as you've added paths as above.

arxanas:
That would be ideal, but I don't know if

* libgit2 (or even the Git CLI) lets you control where to put the include
* if multiple non-consecutive include sections are allowed, which would potentially complicate the semantics

Actually, I suppose the include should go at the top of the file by default, so that any repo-specific aliases are also respected.

typesanitizer:
Yeah, I looked at it a bit more closely, but I don't think doing this using the API or CLI is possible.

We could just modify the file "by hand". Maybe we include comments as section markers.
```
# git-branchless section start
[include]
  # includes added by git-branchless
# git-branchless section end
# old configuration follows...
```
The section markers would make it easier to make the initial setup idempotent (which I understand is the case today).

It feels a little hacky, but I don't have any other bright ideas. Happy to submit a patch with an implementation if you're agree that's the right call.

arxanas:
@typesanitizer Will it still work if the alias is defined in the user's global `.gitconfig`, but the includes are defined in a repo-specific one? My understanding was that the git-branchless aliases would still override the global aliases, regardless of where they're placed in the file.

typesanitizer:
Yep, it should still work; I can include a test case or two 😄. In my testing, included configuration files get the priority of the file they were included in. So those aliases would gain higher priority.

typesanitizer:
[...]

PR #232, opened 2021-11-28
fix(init): avoid accidentally overriding user-defined aliases

typesanitizer:
Draft fix for #219.

What works: if you put an alias in `~/.gitconfig`, then it gets picked over aliases defined by `git-branchless`. So it does solve the immediate problem. This is verified by a new test case.

What doesn't work:
1. If `HOME` is unset, this will give an error. Is it safe to assume that `HOME` will always be set on Unixes?
2. Any other config files, such as `$XDG_CONFIG_HOME/git/config` get lower priority even if they are included later, you were right about that. It's not clear to me why `~` gets special treatment, but it does. 

    This is also why I initially got excited by the solution in the issue discussion, because I accidentally ended up trying the thing that worked manually first. 😬

Because of 1., a bunch of test cases start failing (HOME is unset), and I will probably need to add a special case for Windows. That said, given the drawbacks of this approach, I'm not sure if you want to go with this now. If you're fine with the approach, I'm happy to fix the tests + add the Windows special case. The other alternative is I can change the implementation strategy if you have a better idea on how all cases could be handled at once.

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -7,6 +7,7 @@ use std::path::{Path, PathBuf};
 use console::style;
 use eyre::Context;
 use path_slash::PathExt;
+use regex::Regex;
 use tracing::{instrument, warn};
 
 use crate::core::config::{get_core_hooks_path, get_default_branch_name};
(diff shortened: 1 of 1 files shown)


arxanas:
If you add an include for `~/.gitconfig`, what happens to multivar configuration (like `core.gitproxy`)? Do they appear twice when listed with `git config --get-all`? If so, then I think we shouldn't re-include an already-applied configuration file, or it may unexpectedly change the behavior of something.

My suggested workaround is to figure out whether an alias of the given name already exists and was not defined by `git-branchless`, and if so, to skip installing that alias and delete the existing `git-branchless` alias, if any. It's not ideal, but it's something 🤷

arxanas (changes requested):
Putting back in your queue.

typesanitizer:
> If you add an include for ~/.gitconfig, what happens to multivar configuration (like core.gitproxy)? Do they appear twice when listed with git config --get-all? If so, then I think we shouldn't re-include an already-applied configuration file, or it may unexpectedly change the behavior of something.

Yep, it does show up twice.

> My suggested workaround is to figure out whether an alias of the given name already exists and was not defined by git-branchless, and if so, to skip installing that alias and delete the existing git-branchless alias, if any. It's not ideal, but it's something 🤷

Sounds like a reasonable compromise. I will get to this in a day or two.

arxanas:
Closing for now, re-open if you want a re-review.

PR #233, opened 2021-11-28
docs: add Contributing section to README

typesanitizer:
Initially, I thought there wasn't any developer documentation, because the Developer Docs section in the Wiki TOC is well-hidden under the long User Docs section. 😅

Highlighting it in the README, so that it is less likely that someone else misses it in the future.


arxanas:
Thanks 😊

PR #238, opened 2021-11-30
feat(init): warn if `core.hooksPath` is set

### LINKED ISSUE #235: Hooks installed to global `core.hooksPath` and enabled in all repos
### Description of the bug

When `core.hooksPath` is set to a path outside the local Git repository (e.g. `~/.config/git/hooks`), `git branchless init` places hooks in the global hooks path which are then applied to all Git repositories.

I don't think `core.hooksPath` is commonly changed by users, so this is probably a low priority issue. And this may be more of a problem with how `core.hooksPath` works with non-relative paths, and not `git-branchless`. Regardless, I thought I'd mention it.

### Expected behavior

`git-branchless` hooks were installed in the repository-local hooks directory (`.git/hooks`) and applied to the current Git repository only.

### Actual behavior

`git-branchless` hooks were installed in my global `core.hooksPath` directory (`~/.config/git/hooks`) and applied to Git repositories where I hadn't run `git branchless init`.

9861d7093093ae2a92ff087a44ffd273ae051d37

### Version of `rustc`

_No response_

arxanas:
Closes #235.


evanrelf (approved):
Thanks! ❤️

PR #242, opened 2021-12-03
feat(checkout): add various flags

arxanas:


arxanas:
cc @bcongdon: the idea is to make `git co` mostly interchangeable with `git checkout` (for common uses), so interactive mode is only triggered if no checkout target is specified, or if `-i`/`--interactive` is passed. @KevinWuWon mentioned in https://github.com/arxanas/git-branchless/discussions/218#discussion-3692017 that they mostly ues `git co` with a target and with `-b`.

PR #267, opened 2022-02-10
fix(init): skip adding aliases to a repo when a user already has that alias defined

rslabbert:
As per https://github.com/arxanas/git-branchless/issues/219.

This is a simple change to skip adding an alias if it's already defined in the user's default configs ("global, XDG and system configuration files" as per https://docs.rs/git2/latest/git2/struct.Config.html#method.open_default).

### DIFF
--- a/src/commands/init.rs
+++ b/src/commands/init.rs
@@ -260,13 +260,37 @@ fn should_use_wrapped_command_alias() -> bool {
 }
 
 #[instrument]
-fn install_alias(repo: &Repo, config: &mut Config, from: &str, to: &str) -> eyre::Result<()> {
+fn install_alias(
+    effects: &Effects,
+    repo: &Repo,
+    config: &mut Config,
+    default_config: &Config,
+    from: &str,
+    to: &str,
+) -> eyre::Result<()> {
+    let alias_key = format!("alias.{}", from);
+
+    let existing_alias: Option<String> = config.get(&alias_key)?;
+    if existing_alias.is_some() {
+        config.remove(&alias_key)?;
+    }
+
+    let default_alias: Option<String> = default_config.get(&alias_key)?;
+    if default_alias.is_some() {
+        writeln!(
+            effects.get_output_stream(),
+            "Alias {} already installed, skipping",
+            from
+        )?;
+        return Ok(());
+    }
+
     let alias = if should_use_wrapped_command_alias() {
         format!("branchless-{}", to)
     } else {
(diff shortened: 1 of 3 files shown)


src/commands/init.rs
```diff
@@ -260,13 +260,30 @@ fn should_use_wrapped_command_alias() -> bool {
...
+fn install_alias(
+    effects: &Effects,
+    repo: &Repo,
+    config: &mut Config,
+    default_config: &Config,
+    from: &str,
+    to: &str,
+) -> eyre::Result<()> {
+    let alias_key = format!("alias.{}", from);
+    if default_config.get::<String, _>(&alias_key)?.is_some() {
```
arxanas, line 272: This would probably be more readable:

```
let default_alias: String = default_config.get(&alias_key)?;
if default_alias.is_some() {
```

tests/command/test_init.rs
```diff
@@ -82,6 +84,44 @@ fn test_alias_installed() -> eyre::Result<()> {
...
+    let env = HashMap::from([(
+        "HOME".to_string(),
+        fake_home_dir.to_string_lossy().to_string(),
+    )]);
+    let git_run_options = GitRunOptions {
+        env,
+        ..GitRunOptions::default()
+    };
+
+    let _ = git.run_with_options(&["branchless", "init"], &git_run_options)?;
```
arxanas, line 114: You can just write

```
git.run_with_options(&["branchless", "init"], &git_run_options)?;
```

, we do that in other places already.

arxanas (approved):
Does this commit make it so that we can still overwrite aliases created by `git-branchless`?

rslabbert:
I've updated the PR to ensure that rerunning init will reconfigure the aliases. I've also updated the test to confirm this happens.

src/commands/init.rs
```diff
@@ -260,13 +260,37 @@ fn should_use_wrapped_command_alias() -> bool {
...
+    config: &mut Config,
+    default_config: &Config,
+    from: &str,
+    to: &str,
+) -> eyre::Result<()> {
+    let alias_key = format!("alias.{}", from);
+
+    let existing_alias: Option<String> = config.get(&alias_key)?;
+    if existing_alias.is_some() {
+        config.remove(&alias_key)?;
```
rslabbert, line 275: We might benefit from an idempotent remove

arxanas (approved):
Thanks!

PR #276, opened 2022-02-16
ci: make Nix derivation more overridable

hmenke:
This makes it easier for Nix users to use a different version of Rust to compile git-branchless, e.g.

```nix
# ... using git-branchless overlay ...
pkgs.git-branchless.override {
  rustPlatform = pkgs.rustPackages_1_57.rustPlatform;
}
```

The motivation behind this PR is that the stable NixOS 21.11 is still on Rust 1.56 by default for `rustPlatform` and https://github.com/arxanas/git-branchless/commit/a7ea5ccff54e3a01db8393413c6f5cd49c9a22c6#diff-af29e66a62cac9974dc3f8f7132badf0438c5c4d3426a03492425cf56c691b15R256 introduced `Command::get_envs()` which needs the library feature `command_access` which is only available in Rust 1.57.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,37 +15,56 @@
     in
     {
       overlay = (final: prev: {
-        git-branchless = with final; rustPlatform.buildRustPackage {
-          name = "git-branchless";
+        git-branchless = final.callPackage
+          (
+            { lib
+            , git
+            , libiconv
+            , ncurses
+            , openssl
+            , pkg-config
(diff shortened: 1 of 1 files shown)


arxanas:
Thanks. I think I should try to target as early of a version of Rust as possible anyways. For now, that's Rust v1.56.0, which is the first version supporting edition 2021.

PR #287, opened 2022-03-13
feat(sync): add test for `git sync --pull`

arxanas:


tests/command/test_sync.rs
```diff
@@ -97,3 +99,71 @@ fn test_sync_up_to_date() -> eyre::Result<()> {
...
+    {
+        let (stdout, _stderr) = cloned_repo.run(&["sync", "-p"])?;
+        insta::assert_snapshot!(stdout, @r###"
+        branchless: running command: <git-executable> fetch --all
+        Fetching origin
+        Attempting rebase in-memory...
+        [1/1] Committed as: 8e521a10 create test3.txt
+        branchless: processing 1 update: branch foo
+        branchless: processing 1 rewritten commit
+        branchless: running command: <git-executable> checkout foo
```
martinvonz, line 147: Sorry about an off-topic comment for this PR, but these message have always seemed to me like they should be at debug level. Or do you think users care about them (I've only used git-branchless as few times, so I don't count myself as a user)?

arxanas: In this particular case, since it's delegating to a sub-operation, a lot of the messages are grayed out:

Zrzut ekranu 2022-03-13 o 01 22 36

For the `branchless:` messages, the reasoning was originally:

1) to help debug in early stages of development.
2) to inform users that git-branchless reacts to/tracks user events.

My intention was to add an option to disable these messages for users who understood how git-branchless works, but in practice, I never got around to it because it didn't bother me. (I've been careful to output only a constant number of lines per operation, unlike Git, which might output 100 lines after a commit with 100 files.)

martinvonz: Makes sense. Thanks for explaining.

PR #293, opened 2022-03-15
cleanup(bug_report): remove unnecessary detail from a match expression

martinvonz:


arxanas:
This is actually intentional! The idea is that if I add any new event types or fields, then I'll be forced to audit this use-site to evaluate whether it should be redacted. It could certainly use a comment, though :)

martinvonz:
> This is actually intentional! The idea is that if I add any new event types or fields, then I'll be forced to audit this use-site to evaluate whether it should be redacted. It could certainly use a comment, though :)

Oh! I looked at the code in the PR that introduced it and couldn't figure out what the point was, but it makes sense that you want to be extra careful about it. I can repurpose this PR to be about adding a comment.

martinvonz:
> I can repurpose this PR to be about adding a comment.

Done.

arxanas:
Thanks for the quick adjustment!

PR #294, opened 2022-03-15
tests: ignore system config in tests

martinvonz:
My system has a hook installed (for checking for accidentally
committed secrets), which interfered with tests. Let' set
`GIT_CONFIG_NOSYSTEM` to fix that.


arxanas:
Thanks!

PR #296, opened 2022-03-16
feat(smartlog): add `--only-branches` flag

claytonrcarter:
I believe that this implements an `--only-branches` flag for `smartlog`. This is my first time using Rust (hence the "I believe" part), so any comments or suggestions to make this more idiomatic for rust or the project are warmly welcomed. Also, I was 100% like `\o/` "it compiles!" :smile:

This includes a test, and I've wired up all of the connections that I was able to discern by looking at examples from other commands, etc. Please let me know if I missed something, or should remove any of it.

The logic change was actually pretty easy: instead of `observed_commits.difference(obsolete_commits)`, I only needed to do `observed_commits.intersection(branch_commits)`, so no changes were needed for `query_active_heads` as discussed in #283. *That said*, though this is passing the tests and the handful of examples I have thrown at it, it's entirely possible that I've missed something and this implementation is too naive.

You'll see that I also configured `smartlog` to not allow `--hidden` and `--only-branches` to be used together. This was actually the trickiest part for me to think through, but I ended up realizing hidden branches are shown by default anyway, so `hidden` is just adding hidden *non-branch* commits to the smartlog, which is at odds w/ the goal of `only-branches`. Happy to reconsider this if I've missed something.

### DIFF
--- a/src/commands/smartlog.rs
+++ b/src/commands/smartlog.rs
@@ -224,6 +224,7 @@ mod graph {
         event_replayer: &EventReplayer,
         event_cursor: EventCursor,
         remove_commits: bool,
+        only_branches: bool,
     ) -> eyre::Result<SmartlogGraph<'repo>> {
         let (effects, _progress) = effects.start_operation(OperationType::MakeGraph);
 
@@ -231,14 +232,16 @@ mod graph {
             let (effects, _progress) = effects.start_operation(OperationType::WalkCommits);
 
             let public_commits = dag.query_public_commits()?;
-            let active_heads = if remove_commits {
-                dag.query_active_heads(
-                    &public_commits,
-                    &dag.observed_commits.difference(&dag.obsolete_commits),
-                )?
+
+            let observed_commits = if only_branches {
+                dag.branch_commits.clone()
+            } else if remove_commits {
+                dag.observed_commits.difference(&dag.obsolete_commits)
             } else {
-                dag.query_active_heads(&public_commits, &dag.observed_commits.clone())?
+                dag.observed_commits.clone()
             };
+
+            let active_heads = dag.query_active_heads(&public_commits, &observed_commits)?;
 
             walk_from_active_heads(
                 &effects,
@@ -520,6 +523,9 @@ mod render {
         /// Whether to also show commits in the smartlog which would normally not be
         /// visible.
         pub show_hidden_commits: bool,
+
+        /// Whether to only show commits on branches.
+        pub only_show_branches: bool,
     }
 }
 
@@ -532,6 +538,7 @@ pub fn smartlog(
 ) -> eyre::Result<()> {
     let SmartlogOptions {
         show_hidden_commits,
+        only_show_branches,
     } = options;
 
     let repo = Repo::from_dir(&git_run_info.working_directory)?;
@@ -556,6 +563,7 @@ pub fn smartlog(
         &event_replayer,
         event_cursor,
         !show_hidden_commits,
+        *only_show_branches,
     )?;
 
     let lines = render_graph(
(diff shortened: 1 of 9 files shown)


src/commands/smartlog.rs
```diff
@@ -556,6 +571,7 @@ pub fn smartlog(
         &event_replayer,
         event_cursor,
         !show_hidden_commits,
+        *only_show_branches,
```
claytonrcarter, line 574: It's been a *long* time since I had to deal w/ dereferencing, so I'm not totally clear on this, but the compiler told me I needed to and it seems to be working. I also tried `!!only_show_branches` which seemed to work, but this seemed closer to my (vague) understanding of Rust.

src/commands/smartlog.rs
```diff
@@ -224,22 +224,25 @@ mod graph {
...
 
             let public_commits = dag.query_public_commits()?;
-            let active_heads = if remove_commits {
-                dag.query_active_heads(
-                    &public_commits,
-                    &dag.observed_commits.difference(&dag.obsolete_commits),
-                )?
+
+            let observed_commits = if only_branches {
+                dag.observed_commits.intersection(&dag.branch_commits)
```
arxanas, line 237: That was easy, nice find 😄.

After looking through the logic of `Dag::query_active_heads`, I think we can simplify this even further!

```suggestion
                CommitSet::empty()
```

The reason is that `query_active_heads` will already union the provided set with the list of branches (which we would comment there). We could also just write this for clarity:

```suggestion
                dag.branch_commits.clone()
```

Either one of those seems reasonable to me, so up to you. (Regarding `clone`: the underlying set should be backed by a reference-counted pointer, so it's essentially free.)

arxanas (changes requested):
This looks pretty good overall, just a few minor changes.

src/commands/smartlog.rs
claytonrcarter: Oh, that makes sense! I had a hard time wrapping my head around what `query_active_heads` was doing w/ the 2 params. FWIW I went w/ `dag.branch_commits.clone()` just to be more explicit. Thank you!

tests/command/test_smartlog.rs
```diff
@@ -429,6 +429,70 @@ fn test_show_hidden_commits() -> eyre::Result<()> {
...
+    git.run(&["checkout", "master"])?;
+    git.commit_file("test5", 5)?;
+    git.detach_head()?;
+    git.commit_file("test6", 6)?;
+    git.run(&["checkout", "master"])?;
+    git.commit_file("test7", 7)?;
+
+    git.run(&["branch", "branch-2", &test2_oid.to_string()])?;
+    git.run(&["branch", "branch-4", &test4_oid.to_string()])?;
+    git.run(&["hide", &test4_oid.to_string()])?;
```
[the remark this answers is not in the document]
claytonrcarter: Sounds good. Done.

src/commands/smartlog.rs
```diff
@@ -532,8 +538,17 @@ pub fn smartlog(
...
     } = options;
 
+    if show_hidden_commits & only_show_branches {
+        writeln!(
+            effects.get_output_stream(),
+            "The --hidden and --only-branches options cannot both be provided."
+        )?;
+        return Ok(());
+    }
+
```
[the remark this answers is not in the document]
claytonrcarter: OK, that makes sense. I've removed this. And thanks for the info on `&&`, I missed that when [not very thoroughly, apparently] reading the docs.

src/commands/smartlog.rs
claytonrcarter: That makes total sense now that you've explained it. Thank you!

claytonrcarter:
I think I've taken cafe of all of the changes you've requested and rebased this onto `master`. Thanks again! I've already a local alias set up for this `git slb` 😄

arxanas:
Looks good, thanks for doing this!

Issue #179, opened 2021-10-29
Implement a `git reword` command


bcongdon:
As discussed in #48.

epage:
I'm curious, do you expect `git reword` to be tied to `git branchless init` or would it be like https://github.com/arxanas/git-branchless/issues/183?  I'd like
- `git revise -ie` like behavior
- Branches to be preserved
- Able to be used independent of `git branchless`.

Trying to decide if its worth making my own or if I can somehow install yours without the other parts (or just not run the other parts)

arxanas:
I think it would be reasonable to make `git reword` work without `git branchless init`/hooks.

bcongdon:
> If you reword multiple commits at once, then it opens an editor with all of their commit messages separated by some marker.

That sounds like a useful feature. 👍 I was thinking that even in the single-commit case, there should be support for dropping into the `$EDITOR`. e.g. `git reword -m "new message"` is legal, but the bare `git reword` would drop you into the commit message editor similar to `git commit --amend`. Maybe the single-commit case is just a special case of multi-commit reword.

arxanas:
I wonder what the behavior should be for `git reword -m "new message" 'revset_expr_returning_multiple_commits()'`?

For the future implementor, `dialoguer` has a function to spawn an editor, which might be useful:https://docs.rs/dialoguer/0.9.0/dialoguer/struct.Editor.html

claytonrcarter:
Perhaps something like:

- `git reword` - open `$EDITOR`, amend `HEAD`
- `git reword <commit>` - open `$EDITOR`, amend `<commit>`
- `git reword <multiple commits>` - open `$EDITOR` for bulk editing of `<multiple commits>`
- `git reword -m <message>` - use `<message>` to amend `HEAD`
- `git reword <commit> -m <message>` - use `<message>` to amend `<commit>`
- `git reword <multiple commits> -m <message>` - use same `<message>` to amend all `<multiple commits>`, possibly printing a warning to stderr or something like that

Personally, I could see myself using all of these. The last is the oddest, of course, but could be useful for something like `git reword <buncha commits on branch> -m 'fixup!' && git rebase -i --autosquash main`

claytonrcarter:
For those playing along at home, `git reword` was recently merged into `master` via #304. To the best of my knowledge, it handles everything discussed here *except* bulk edit mode a la `git revise -ie`. I would love hear your thoughts and experiences from using it in it's present form.

[...]

PR #304, opened 2022-03-21
feat: implement `git reword` command (#179)

claytonrcarter:
**Updated PR description:**

This supports all of the use-cases that [I mentioned on #179](https://github.com/arxanas/git-branchless/issues/179#issuecomment-1071087014), except for the fancy bulk editing case. That is to say:

 - Reword single or multiple commits within or across subtrees.
 - Currently, the same message is applied to all commits.
 - If multiple message flags are given on the CLI, they are combined into separate paragraphs.
 - If no message flags are passed on the CLI, `$EDITOR` is opened w/ either the message of the commit being reworded (if just rewording a single commit) or a stub message when rewording multiple commits.
 - Bails out if the edited message is identical to the input message, or if it's blank.

A few things to make special note of:

 -  I'm still *very much* learning my way through Rust, so please let me know where I can improve my usage of the language or be more in line w/ the community or project. 
 -  I think I've taken a pretty good stab at error handling, but please keep a special eye out for possible places that I missed something or was over eager w/ my usage of `?` or `.unwrap()`, etc

Finally, there a few things left to do:

 -  [x] I haven't done anything to look at @epage request that it [operate independent of `branchless init`](https://github.com/arxanas/git-branchless/issues/179#issuecomment-955109646). At this time, I don't even know where to start w/ that, so any pointers would be most welcome.
 -  [x] The bulk editing feature is not implemented; I am voting :hand: for us to review this as is and add that later, though I'm also fine to push on to work on bulk editing now if you'd prefer.
 -  [x] During my latest read through, I decided that we should probably a "footer" message to the `$EDITOR` akin to the default `commit` message from `git`. I'll work on this soon.
 -  [x] I'd like to take another look through my error handling and borrow/clone/dereferencing/option/basically/everything usage again.

I *think* that covers most of it, but I'm sure I'm missing something! :smile:

**Original PR description below:**

Still *very* draft, but it's working for all of the use-cases that [I mentioned on #179](https://github.com/arxanas/git-branchless/issues/179#issuecomment-1071087014), except for the fancy bulk editing case. That is to say:

 - Reword single or multiple commits within or across subtrees.
 - Currently, the same message is applied to all commits.
 - If multiple message flags are given on the CLI, they are combined into separate paragraphs.
 - If no message flags are passed on the CLI, `$EDITOR` is opened w/ either the message of the commit being reworded (if just rewording a single commit) or a stub message when rewording multiple commits.
 - Prompt for confirmation when appling the same message to multiple commits. (`--force` flag skips confirmation.)
 - Bails out if the edited message is identical to the input message, or if it's blank.

So, lots to love there already, but still lots to do:

 -  I'm still *very much* learning my way through Rust, so expect lots of :man_facepalming: moments if you're brave enough to read through this, especially when it comes to borrowing, scope and such.
 - This is very "happy pathy" right now. I need to go back and flesh out the error handling.
 - ~~For expediency, I implemented this by altering `r#move::r#move` (rewording is kind of just a special case of moving, at least, it seems so to me) but I just did that so that I could focus on the fun parts and not get bogged down trying to figure out the plumbing. (It's just a 4 line change in `r#move`.) I intend to break this entanglement and move the actual rebase planning into `reword`, but it's still in place at this time. Because of this, though, it's doing a restack/smartlog for every subtree that's being reworded, instead of just doing a single restack and smartlog at the end.~~ **Edit:** I've removed the coupling w/ `r#move` and rewritten the rebase planning and execution based on `sync`.
 - Also note that I did not implement this via a new `RebaseCommand`,. eg `RebaseCommand::Reword`. Instead I just updated`RebaseCommand::Pick` to support an optional message since a reword is just a pick but with a different message and the change was trivial. I'm inclined to leave this in place unless you feel otherwise.
 - B/c reword is treated as just a special case of `move`, I wanted to reuse all of the normal `MoveOptions`, *but* I wanted to reserve the shorthand `-m` flag for `--messages`, not for `merge` as in `MoveOptions` ... so I just duplicated `MoveOptions` into `RebaseOption` except w/o the `-m` flag. This is still TBD, but I'd like a way to just reuse `MoveOptions` while overriding the binding of `-m`, but I haven't delved into how/if this can be done in clap.
 - I haven't done anything to look at @epage request that it [operate independent of `branchless init`](https://github.com/arxanas/git-branchless/issues/179#issuecomment-955109646). At this time, I don't even know where to start w/ that, so any pointers would be most welcome.
 - Bulk editting? I wonder if bulk editing might be best considered a feature unto itself. If so, this existing command could be finished off and proper bulk editing support (ie different messages for different commits) could be handled via a separate PR.

I'm going to park this here for a bit while I go away and read more about borrowing, Boxes, memory mgmt, error handling etc in Rust. :smile: I hesitate to even ask for input yet, but I think it would be good to know if I'm heading in the right direction or not. (And of course even if I'm way off base and this is just totally not viable, it was still a fun learning exercise!)

Oh, and I *totally* used this to squash all of my ugly working commits. This is a slightly edited version:

```
❯ ./target/debug/git-branchless reword --message 'hacking on `reword`' <base commit on branch>
❯ ./target/debug/git-branchless reword --message 'fixup! hacking on `reword`' <lots of descendent commits>
❯ git rebase --autosquash -i master 
❯ ./target/debug/git-branchless reword --message 'feat: implement `reword` command'
```

Thanks for your consideration!

### DIFF
--- a/src/commands/reword.rs
+++ b/src/commands/reword.rs
@@ -15,7 +15,7 @@ use crate::core::config::{get_comment_char, get_restack_preserve_timestamps};
 use crate::core::dag::{resolve_commits, CommitSet, Dag, ResolveCommitsResult};
 use crate::core::effects::Effects;
 use crate::core::eventlog::{EventLogDb, EventReplayer};
-use crate::core::formatting::{printable_styled_string, Glyphs};
+use crate::core::formatting::{printable_styled_string, Glyphs, Pluralize};
 use crate::core::node_descriptors::{render_node_descriptors, CommitOidDescriptor, NodeObject};
 use crate::core::rewrite::{
     execute_rebase_plan, BuildRebasePlanOptions, ExecuteRebasePlanOptions, ExecuteRebasePlanResult,
@@ -231,13 +231,34 @@ fn build_messages(
             },
             _ => {
                 // TODO(bulk edit) build a bulk edit message for multiple commits
-                format!(
-                    "{} Enter a commit message to apply to {} commits",
-                    comment_char.unwrap(),
-                    commits.len()
-                )
+                String::from("")
             }
         };
+
+        // If the original commit message didn't end w/ a newline, add one. This makes it easy to
+        // ensure that our "help" text will always be separated from the commit message by a blank
+        // line, and messages passing through Editor will end up with a newline appended anyway.
+        if !message_for_editor.ends_with('\n') {
+            message_for_editor.push('\n');
+        };
+
+        message_for_editor.push_str(
+            format!(
+                "\n\
+                {} Rewording: Please enter the commit message to apply to {}. Lines\n\
+                {} starting with '{}' will be ignored, and an empty message aborts rewording.",
+                comment_char,
+                Pluralize {
+                    determiner: Some(("this", "these")),
+                    amount: commits.len() as isize,
+                    unit: ("commit", "commits"),
+                },
+                comment_char,
+                comment_char,
+            )
+            .as_str(),
+        );
+
         (message_for_editor, true)
     } else {
         (message, false)
--- a/src/commands/reword.rs
+++ b/src/commands/reword.rs
@@ -252,17 +252,21 @@ fn build_messages(
             .require_save(false)
             .edit(message.as_str())?
             .unwrap();
+
         if edited_message == message {
             return Ok(BuildRewordMessageResult::IdenticalMessage);
         }
-        edited_message
+
+        // clean up the commit message: remove comment lines, finish w/ a newline, etc
+        // TODO move this to git module
+        git2::message_prettify(edited_message, comment_char)?
     } else {
+        // ensure that CLI messages also end w/ a newline
+        if !message.ends_with('\n') {
+            message.push('\n');
+        }
         message
     };
-
-    // clean up the commit message: remove comment lines, finish w/ a newline, etc
-    // TODO ask if this should be here or if it should be moved to the git module per project wiki
-    let message = git2::message_prettify(message, comment_char)?;
 
     if message.trim().is_empty() {
         return Ok(BuildRewordMessageResult::EmptyMessage);
--- a/src/commands/reword.rs
+++ b/src/commands/reword.rs
@@ -1,35 +1,35 @@
 //! Update commit messages
 
 use rayon::ThreadPoolBuilder;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt::Write;
 use std::time::SystemTime;
 
-use dialoguer::{Confirm, Editor};
+use dialoguer::Editor;
 use eden_dag::DagAlgorithm;
-use tracing::instrument;
+use eyre::Context;
+use tracing::{instrument, warn};
 
-use crate::core::config::get_restack_preserve_timestamps;
+use crate::core::config::{get_comment_char, get_restack_preserve_timestamps};
 use crate::core::dag::{resolve_commits, CommitSet, Dag, ResolveCommitsResult};
-use crate::core::effects::{Effects, OperationType};
+use crate::core::effects::Effects;
 use crate::core::eventlog::{EventLogDb, EventReplayer};
+use crate::core::formatting::{printable_styled_string, Glyphs};
+use crate::core::node_descriptors::{render_node_descriptors, CommitOidDescriptor, NodeObject};
 use crate::core::rewrite::{
-    execute_rebase_plan, BuildRebasePlanError, BuildRebasePlanOptions, ExecuteRebasePlanOptions,
-    ExecuteRebasePlanResult, RebasePlan, RebasePlanBuilder, RepoResource,
+    execute_rebase_plan, BuildRebasePlanOptions, ExecuteRebasePlanOptions, ExecuteRebasePlanResult,
+    RebasePlanBuilder, RepoResource,
 };
-use crate::git::{Commit, GitRunInfo, NonZeroOid, Repo};
-use crate::opts::RebaseOptions;
+use crate::git::{CheckOutCommitOptions, Commit, GitRunInfo, MaybeZeroOid, NonZeroOid, Repo};
 
 /// Reword a commit and restack it's descendants.
 #[instrument]
 pub fn reword(
     effects: &Effects,
-    force: bool,
     hashes: Vec<String>,
     messages: Vec<String>,
     git_run_info: &GitRunInfo,
-    rebase_options: &RebaseOptions,
 ) -> eyre::Result<isize> {
     let repo = Repo::from_current_dir()?;
     let references_snapshot = repo.get_references_snapshot()?;
@@ -50,99 +50,76 @@ pub fn reword(
         None => return Ok(1),
     };
 
-    let messages = match build_messages(&messages, &commits)? {
+    let messages = match build_messages(&messages, &commits, get_comment_char(&repo)?)? {
         BuildRewordMessageResult::Succeeded { messages } => messages,
         BuildRewordMessageResult::IdenticalMessage => {
-            println!("The message wasn't edited; nothing to do. Stopping.");
+            writeln!(
+                effects.get_output_stream(),
+                "The message wasn't edited; nothing to do. Stopping."
+            )?;
             return Ok(1);
         }
         BuildRewordMessageResult::EmptyMessage => {
-            eprintln!("Error: the reworded message was empty. Stopping.");
-            return Ok(1);
-        }
-        BuildRewordMessageResult::Failed { message } => {
-            if message != None {
-                eprintln!("{}", message.unwrap());
-            }
-            eprintln!("Error: problem processing commit message. Nothing reworded.");
+            writeln!(
+                effects.get_error_stream(),
+                "Error: the reworded message was empty. Stopping."
+            )?;
             return Ok(1);
         }
     };
 
-    if commits.len() != 1
-        && !force
-        && !Confirm::new()
-            .with_prompt(format!(
-                "Warning: attempting to apply the same message to {} commits. Continue?",
-                commits.len()
-            ))
-            .default(false)
-            .interact()?
-    {
-        println!("Ok. Nothing reworded.");
-        return Ok(1);
-    }
-
     let subtree_roots = find_subtree_roots(&repo, &dag, &commits)?;
 
-    // TODO instead of calling into `r#move` shoud we just be building the rebase plan ourselves?
-    // TODO look at `sync` to see how it handles return values for multiple subtree moves
-    // TODO what if there is no parent?
-    // TODO what if there are multiple parents?
-    let RebaseOptions {
-        force_in_memory,
-        force_on_disk,
-        detect_duplicate_commits_via_patch_id,
-        resolve_merge_conflicts,
-        dump_rebase_constraints,
-        dump_rebase_plan,
-    } = *rebase_options;
-
-    let subtree_roots_and_plans: Vec<(Commit, Option<RebasePlan>)> = {
+    let rebase_plan = {
         let pool = ThreadPoolBuilder::new().build()?;
         let repo_pool = RepoResource::new_pool(&repo)?;
-        let builder = RebasePlanBuilder::new(&dag);
+        let mut builder = RebasePlanBuilder::new(&dag);
 
-        let subtree_roots_and_plans =
-            subtree_roots
-                .into_iter()
-                .map(
-                    |root_commit| -> eyre::Result<
-                        Result<(Commit, Option<RebasePlan>), BuildRebasePlanError>,
-                    > {
-                        // Keep access to the same underlying caches by cloning the same instance of the
-                        // builder.
-                        let mut builder = builder.clone();
+        for root_commit in subtree_roots {
+            let only_parent_id = root_commit.get_only_parent().map(|parent| parent.get_oid());
+            if only_parent_id == None {
+                eyre::bail!(
+                    "###
+                    Refusing to reword commit {}, which has {} parents.
+                    Rewording only supported for commits with 1 parent.
+                    Stopping; nothing reworded.
+                    ###",
+                    root_commit.get_oid(),
+                    root_commit.get_parents().len(),
+                );
+            }
+            builder.move_subtree(root_commit.get_oid(), only_parent_id.unwrap())?;
+        }
 
-                        let rebase_plan = {
-                            let only_parent_id =
-                                root_commit.get_only_parent().map(|parent| parent.get_oid());
-                            if only_parent_id == None {
-                                // TODO multiple parents!
-                            }
-                            builder.move_subtree(root_commit.get_oid(), only_parent_id.unwrap())?;
-                            builder.reword_commits(&messages)?;
-                            builder.build(
-                                effects,
-                                &pool,
-                                &repo_pool,
-                                &BuildRebasePlanOptions {
-                                    dump_rebase_constraints,
-                                    dump_rebase_plan,
-                                    detect_duplicate_commits_via_patch_id,
-                                },
-                            )?
-                        };
+        for commit in commits.iter() {
+            let message = messages.get(&commit.get_oid()).unwrap();
+            // This looks funny, but just means "leave everything but the message as is"
+            let replacement_oid =
+                commit.amend_commit(None, None, None, Some(message.as_str()), None)?;
+            builder.replace_commit(commit.get_oid(), replacement_oid)?;
+        }
 
-                        Ok(rebase_plan.map(|rebase_plan| (root_commit, rebase_plan)))
-                    },
-                )
-                .collect::<eyre::Result<Vec<_>>>()?
-                .into_iter()
-                .collect::<Result<Vec<_>, BuildRebasePlanError>>();
-
-        match subtree_roots_and_plans {
-            Ok(subtree_roots_and_plans) => subtree_roots_and_plans,
+        match builder.build(
+            effects,
+            &pool,
+            &repo_pool,
+            &BuildRebasePlanOptions {
+                dump_rebase_constraints: false,
+                dump_rebase_plan: false,
+                detect_duplicate_commits_via_patch_id: false,
+            },
+        )? {
+            Ok(Some(rebase_plan)) => rebase_plan,
+            Ok(None) => {
+                writeln!(
+                    effects.get_output_stream(),
+                    "###
+                    BUG: rebase plan indicates nothing to do, but rewording should always do something.
+                    Stopping; nothing reworded.
+                    ###"
+                )?;
+                return Ok(1);
+            }
             Err(err) => {
                 err.describe(effects, &repo)?;
                 return Ok(1);
@@ -156,97 +133,34 @@ pub fn reword(
         now,
         event_tx_id,
         preserve_timestamps: get_restack_preserve_timestamps(&repo)?,
-        force_in_memory,
-        force_on_disk,
-        resolve_merge_conflicts,
-        check_out_commit_options: Default::default(),
+        force_in_memory: true,
+        force_on_disk: false,
+        resolve_merge_conflicts: false,
+        check_out_commit_options: CheckOutCommitOptions {
+            additional_args: Default::default(),
+            render_smartlog: false,
+        },
     };
+    let result = execute_rebase_plan(effects, git_run_info, &repo, &rebase_plan, &execute_options)?;
 
-    let (_success_commits, _merge_conflict_commits, _skipped_commits) = {
-        let mut success_commits: Vec<Commit> = Vec::new();
-        let mut merge_conflict_commits: Vec<Commit> = Vec::new();
-        let mut skipped_commits: Vec<Commit> = Vec::new();
+    let exit_code = match result {
+        ExecuteRebasePlanResult::Succeeded { rewritten_oids } => {
+            render_status_report(&repo, effects, &commits, &rewritten_oids)?;
 
-        let (effects, progress) = effects.start_operation(OperationType::RebaseCommits);
-        progress.notify_progress(0, subtree_roots_and_plans.len());
-
-        for (root_commit, rebase_plan) in subtree_roots_and_plans {
-            let rebase_plan = match rebase_plan {
-                Some(rebase_plan) => rebase_plan,
-                None => {
-                    // Nothing to do ... but why? Bug?
-                    skipped_commits.push(root_commit);
-                    continue;
-                }
-            };
-
-            let result = execute_rebase_plan(
-                &effects,
-                git_run_info,
-                &repo,
-                &rebase_plan,
-                &execute_options,
-            )?;
-            progress.notify_progress_inc(1);
-
-            match result {
-                ExecuteRebasePlanResult::Succeeded => {
-                    success_commits.push(root_commit);
-                }
-                ExecuteRebasePlanResult::DeclinedToMerge { merge_conflict: _ } => {
-                    // FIXME not sure this is even possible here...
-                    merge_conflict_commits.push(root_commit);
-                }
-                ExecuteRebasePlanResult::Failed { exit_code } => {
-                    return Ok(exit_code);
-                }
-            }
+            0
         }
+        ExecuteRebasePlanResult::DeclinedToMerge { merge_conflict: _ } => {
+            writeln!(
+                effects.get_error_stream(),
+                "BUG: Merge conflict detected, but rewording shouldn't cause any conflicts."
+            )?;
 
-        (success_commits, merge_conflict_commits, skipped_commits)
+            1
+        }
+        ExecuteRebasePlanResult::Failed { exit_code } => exit_code,
     };
 
-    // FIXME how much of this is relevant? User asked to reword commits, but we've rebased subtrees
-    // If we really want to show a relevant message perhaps it should be something like
-    // "executed X rebases to reword Y commits"?
-    //
-    // for success_commit in success_commits {
-    //     writeln!(
-    //         effects.get_output_stream(),
-    //         "{}",
-    //         printable_styled_string(
-    //             &glyphs,
-    //             StyledStringBuilder::new()
-    //                 .append_plain("Synced ")
-    //                 .append(success_commit.friendly_describe(&glyphs)?)
-    //                 .build()
-    //         )?
-    //     )?;
-    // }
-    //
-    // for merge_conflict_commit in merge_conflict_commits {
-    //     writeln!(
-    //         effects.get_output_stream(),
-    //         "{}",
-    //         printable_styled_string(
-    //             &glyphs,
-    //             StyledStringBuilder::new()
-    //                 .append_plain("Merge conflict for ")
-    //                 .append(merge_conflict_commit.friendly_describe(&glyphs)?)
-    //                 .build()
-    //         )?
-    //     )?;
-    // }
-    //
-    // for skipped_commit in skipped_commits {
-    //     writeln!(
-    //         effects.get_output_stream(),
-    //         "Not moving up-to-date stack at {}",
-    //         printable_styled_string(&glyphs, skipped_commit.friendly_describe(&glyphs)?)?
-    //     )?;
-    // }
-
-    Ok(0)
+    Ok(exit_code)
 }
 
 /// Turn a list of ref-ish strings into a list of Commits.
@@ -287,45 +201,37 @@ pub enum BuildRewordMessageResult {
 
     /// The reworded message was empty.
     EmptyMessage,
-
-    /// Misc failure.
-    Failed {
-        /// The failue message, if any.
-        message: Option<String>,
-    },
 }
 
 /// Builds the message(s) that will be used for rewording. These are mapped from each commit's
 /// NonZeroOid to the relevant message.
+#[instrument]
 fn build_messages(
     messages: &[String],
     commits: &[Commit],
+    comment_char: Option<u8>,
 ) -> eyre::Result<BuildRewordMessageResult> {
     let message = messages.join("\n\n").trim().to_string();
 
-    // TODO(maybe?) stdin?
-
     let (message, load_editor) = if message.is_empty() {
-        let message_for_editor = match commits.len() {
-            1 => {
-                // FIXME lots of error checking to do
-                let commit = commits.first().unwrap();
-                let msg = commit.get_message_raw()?;
-                let msg = msg.into_string();
-                match msg {
-                    Ok(msg) => msg,
-                    _ => {
-                        return Ok(BuildRewordMessageResult::Failed {
-                            message: Some(String::from(
-                                "Reword: Error decoding original commit message!",
-                            )),
-                        })
-                    }
-                }
-            }
-            len => {
-                // TODO build a bulk edit message for multiple commits
-                format!("Enter a commit message to apply to {} commits", len)
+        let message_for_editor = match commits.to_vec().as_slice() {
+            [commit] => match commit.get_message_raw()?.into_string() {
+                Ok(msg) => msg,
+                _ => eyre::bail!(
+                    "###
+                    Error decoding original message for commit: {:?}.
+                    Stopping; nothing reworded.
+                    ###",
+                    commit.get_oid()
+                ),
+            },
+            _ => {
+                // TODO(bulk edit) build a bulk edit message for multiple commits
+                format!(
+                    "{} Enter a commit message to apply to {} commits",
+                    comment_char.unwrap(),
+                    commits.len()
+                )
             }
         };
         (message_for_editor, true)
@@ -334,8 +240,10 @@ fn build_messages(
     };
 
     let message = if load_editor {
-        // Editor::edit will only return None if saving is required; b/c we've disabled the save
-        // requirement, we don't need to worry about it.
+        // Editor::edit() normally requires that the file being editted is saved before the editor
+        // is closed. If it's not, it will return None; and in all other cases, it will return
+        // Some(message). We don't care if the file has been saved, only if it hasn't changed, so
+        // here we call `require_save(false)` and just ignore the None case.
         let edited_message = Editor::new()
             .require_save(false)
             .edit(message.as_str())?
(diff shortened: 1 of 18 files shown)


claytonrcarter:
Sorry for the churn. I *think* I've fixed all of the formatting and clippy issues.

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+        r#move(
+            effects,
+            git_run_info,
+            Some(base_commit.get_oid().to_string()),
+            Some(parent.get_oid().to_string()),
+            None, // base
+            Some(messages.clone()),
+            &move_options,
+        )?;
+    }
```
arxanas: Yep, we'll want to build and execute the rebase plan ourselves. We don't want to call `move` directly because it prints a bunch of auxiliary output that won't be relevant for `git reword`. I think it will also make the `find_subtree_roots` call unnecessary. (For future reference, I think what you wanted was `dag.query().roots(commit_set)`?)

 It should look something like this:

```rust
let rebase_plan = {
    let mut builder = RebasePlanBuilder::new(...);
    for (commit_oid, message) in messages {
        let replacement_oid = repo.amend_commit(None, ...)?;
        builder.replace_commit(commit_oid, replacement_oid)?;
        // TODO: handle the case of not-exactly-one parent.
        builder.move_subtree(commit_oid, commit_oid.get_only_parent());
    }
};
execute_rebase_plan(...)?;
```

It should actually be pretty straightforward to make `move_subtree` take a variable number of parents, but let's skip that for now and just `eyre::bail!` if we get that situation. I can fix it later.

arxanas (changes requested):
This is really exciting!

One note: before we merge this, squash together the non-functional changes like formatting and linting into the appropriate original commits. You can use `git rebase -i` to do this, or something like `git co <commit> && cargo fmt && cargo clippy --fix && git amend -m`.

src/commands/reword.rs
```diff
@@ -97,68 +97,154 @@ pub fn reword(
...
-                };
-                execute_rebase_plan(effects, git_run_info, &repo, &rebase_plan, &options)?
-            }
+
+    let subtree_roots_and_plans: Vec<(Commit, Option<RebasePlan>)> = {
+        let pool = ThreadPoolBuilder::new().build()?;
+        let repo_pool = RepoResource::new_pool(&repo)?;
+        let builder = RebasePlanBuilder::new(&dag);
+
+        let subtree_roots_and_plans =
```
arxanas, line 106: Ah, you already figured out what I commented in my last review. To simplify this, attach all of the rewords to the same `RebasePlanBuilder` and just execute one rebase plan. `git sync` is different because some of the stacks may fail, and we want to proceed to process other stacks. In this case, none of the rebases should fail.

src/opts.rs
```diff
@@ -48,6 +48,41 @@ pub struct MoveOptions {
     pub dump_rebase_plan: bool,
 }
 
+/// Options for rebasing commits. These are the same as MoveOptions, but without the shorthand -m.
+#[derive(Args, Debug)]
+pub struct RebaseOptions {
```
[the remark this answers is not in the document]
claytonrcarter: OK, that makes sense. To confirm, you mean that we don't need any of these as options via CLI flags on `git reword`, and that instead we can just use/set these default values directly, during the rebase build/execution?

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+        (message_for_editor, true)
+    } else {
+        (message, false)
+    };
+
+    let message = if load_editor {
+        // Editor::edit will only return None if saving is required; b/c we've disabled the save
+        // requirement, we don't need to worry about it.
+        let edited_message = Editor::new()
+            .require_save(false)
```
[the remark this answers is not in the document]
claytonrcarter: It's just a configurable behavior of the `dialoguer::Editor`. By default, if you close the editor w/o saving, this would return `None`. I don't think it's relevant to care about the file being saved, only if the resulting output message is the same as the input message, so I disabled this requirement. I'll update the comment to make that more clear.

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+            .unwrap();
+        if edited_message == message {
+            return Ok(BuildRewordMessageResult::IdenticalMessage);
+        }
+        edited_message
+    } else {
+        message
+    };
+
+    // TODO process the message: remove comment lines, trim, etc
```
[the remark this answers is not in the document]
claytonrcarter: Oh, that sounds *way* easier than what I would have come up with. Thanks!

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+    // TODO FIXME what if the bulk edit message doesn't include messages for != `commits.len`
+    // TODO FIXME what if the bulk edit message includes messages for commits not in `commits`
+
+    let messages: HashMap<NonZeroOid, String> = {
+        let mut m = HashMap::new();
+        for commit in commits.iter() {
+            m.insert(commit.get_oid(), message.clone());
+        }
+        m
+    };
```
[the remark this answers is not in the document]
claytonrcarter: Oh yes, much nicer! Thank you!

src/opts.rs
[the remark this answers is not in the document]
arxanas: @claytonrcarter yep, that's right, you would manually construct a `MoveOptions` value and pass it in where necessary. (I guess it's misleading that one thing is called "move options" and another thing is called a "rebase plan", instead of both of them being called either "move" or "rebase".)

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    if commits.len() != 1
+        && !force
+        && !Confirm::new()
+            .with_prompt(format!(
+                "Warning: attempting to apply the same message to {} commits. Continue?",
+                commits.len()
+            ))
+            .default(false)
+            .interact()?
```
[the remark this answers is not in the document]
claytonrcarter: Oh, yes, that makes complete sense. I added this mostly because I wasn't sure if I was wiring everything up to make `undo` work correctly.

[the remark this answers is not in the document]
arxanas: Conveniently, if we pass everything through `execute_rebase_plan`, then we should get `undo` for free. Note that you'll need to pass an `EventTransactionId` in via the `ExecuteRebasePlanOptions` (which is used to correlate related events for `git undo`). You can do it like this:

```rust
    let event_tx_id = event_log_db.make_transaction_id(now, "reword")?;
```

src/commands/reword.rs
claytonrcarter: The part about `git sync` being different b/c of possible failures/nothing-to-dos makes sense, but can you clarify what you mean about a single rebase plan? Do you mean that we should be rewriting multiple subtrees w/i a single plan, or that we should instead be tracing the dag back to a single common ancestor and just executing a rebase on that tree to catch all of the requested rewordings? If the former: cool! I hadn't tried that b/c I didn't realize that was possible!; if the latter: I would be concerned that that could potentially encompass a *lot* of commits pretty quickly. Regardless, I look forward to your clarification.

claytonrcarter: Oh, nevermind, I just read your other comment where you spelled it all out. Thanks!

arxanas: Yep, just call `.move_subtree` on the same `RebasePlanBuilder` multiple times. You can pass in unrelated trees or commits which are ancestors of one another and everything should just work.  (It's a pretty nifty API if I do say so myself 😎)

src/commands/reword.rs
claytonrcarter: Gah! So brilliantly simple! So much code I didn't need to write! (I'll just keep telling myself it was fun.) And I was so proud of my `find_subtree_roots` algorithm ... of course there's a ready made 1-liner to replace it! 😄  I'll work on cleaning this up tomorrow (or later this week.) Thanks again!

src/commands/reword.rs
arxanas: The example in restack.rs might be a better guide than the one in sync.rs:

https://github.com/arxanas/git-branchless/blob/45ace45579804697f8daeb551750424bf8e8538b/src/commands/restack.rs#L119-L172

src/commands/reword.rs
```diff
@@ -0,0 +1,314 @@
...
+                            )),
+                        })
+                    }
+                }
+            }
+            len => {
+                // TODO build a bulk edit message for multiple commits
+                format!("Enter a commit message to apply to {} commits", len)
+            }
+        };
```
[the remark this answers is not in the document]
claytonrcarter: I made this change, but I couldn't just do `match commits.as_slice()`, I had to do `match commits.to_vec().as_slice()` to keep clippy & the compiler happy. If you'd like me to push on it a little bit more to get `commits.as_slice()` to work, please let me know.

src/commands/reword.rs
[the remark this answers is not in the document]
claytonrcarter: Thank for this suggestion. I was able to clean up at least one other loop into a `map/filter_map` based on this.

src/commands/reword.rs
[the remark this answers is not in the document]
claytonrcarter: This has been removed. `undo` FTW!

src/commands/reword.rs
claytonrcarter: Ah, yes, this is *much* nicer. I'm going to leave `find_subtree_roots` in place for the time being, but in a drastically simplified format that uses `dag.query().roots()`. I might inline that later, but for now it still feels like nice encapsulation.

src/commands/reword.rs
claytonrcarter: Thank you, `git restack` was helpful. And yes, the API is slick. The hardest part for me has just been getting up to speed w/ the language and figuring out the idioms. After that, the feature basically wrote itself. 😄

claytonrcarter:
I just pushed a bunch of changes, but FYI they're not complete. I've taken care of all of the "easy" suggestions, now I'll start working on the juicier bits. I squashed all of the changes that you looked at yesterday into a single commit (noted as such) to maybe make it clearer what's been added when. Thanks again!

src/core/rewrite/plan.rs
```diff
@@ -152,6 +158,9 @@ pub struct RebasePlanBuilder<'repo> {
     /// `y`.
     initial_constraints: HashMap<NonZeroOid, HashSet<NonZeroOid>>,
 
+    /// FIXME
+    messages_to_apply: Option<HashMap<NonZeroOid, String>>,
```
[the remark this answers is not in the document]
claytonrcarter: I suspect that I'm misunderstanding something, but I tried implementing this and have gotten stuck. The rewrite and rebase appear to proceed w/o error, but the graph is left incorrect, w/ both the old commit and new/reworded commit still visible and branch/head not updated. For example, from the first test case (`test_reword_head`):
```
# initial state
    :
    O 62fc20d2 (test1) create test1.txt
    |
    @ 96d1c37a (> master) create test2.txt

# expected state
    :
    O 62fc20d2 (test1) create test1.txt
    |
    @ b38039bd (> master) foo

# actual state I'm getting
    :
    O 62fc20d2 (test1) create test1.txt
    |\
    | o b38039bd foo
    |
    @ 96d1c37a (> master) create test2.txt
```

So, it's creating the correctly and putting it at the right place, but it's not moving `master` and updating HEAD.

FWIW I pushed a `wip` commit w/ my current work. Basically:
1. in `reword`: create new amended commits, give these to `builder.replace_commit()`
2. in `builder.build`: add `Pick` to the rebase plan as normal but use the replacement commit oid instead of the original oid, if a replacement exists
3. in `execute_rebase_plan`, just deal w/ the `Pick` normally.  In fact, this file no longer has any changes from `master`, as far as I can tell. 

I'm assuming I'm missing a step that I can dig out of another command, but I need to stop for the night for now. If you get to this before I do, do I need to explicitly call restack after the rebase or something? Or am I missing a step somewhere in `execute_rebase_plan`? Sorry to be a bother and to need so much support on this!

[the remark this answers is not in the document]
arxanas: @claytonrcarter Oops, I didn't realize that there would be more steps to make the replacement commits work. We just need to update the `rebase_in_memory` function to let it know that replaced commits should be considered "rewritten" commits. I pushed the following commit to the branch `reword` on your repo (which apparently I'm allowed to do!):

```diff
commit 5ca1782d48eadc0fbd31bad7f25e712251e7f2c1
Author: Waleed Khan <me@waleedkhan.name>
Date:   Tue Mar 22 20:37:08 2022

    wip: correctly rewrite replaced commits

diff --git a/src/core/rewrite/execute.rs b/src/core/rewrite/execute.rs
index 12a0d4c..8b84024 100644
--- a/src/core/rewrite/execute.rs
+++ b/src/core/rewrite/execute.rs
@@ -273,6 +273,7 @@ pub fn check_out_updated_head(
 }
 
 /// Information about a merge conflict that occurred while moving commits.
+#[derive(Debug)]
 pub struct MergeConflictInfo {
     /// The OID of the commit that, when moved, caused a conflict.
     pub commit_oid: NonZeroOid,
@@ -443,12 +444,15 @@ mod in_memory {
                     current_oid = *commit_oid;
                 }
 
-                RebaseCommand::Pick { commit_oid } => {
+                RebaseCommand::Pick {
+                    original_commit_oid,
+                    commit_to_apply_oid,
+                } => {
                     let current_commit = repo
                         .find_commit_or_fail(current_oid)
                         .wrap_err("Finding current commit")?;
                     let commit_to_apply = repo
-                        .find_commit_or_fail(*commit_oid)
+                        .find_commit_or_fail(*commit_to_apply_oid)
                         .wrap_err("Finding commit to apply")?;
                     i += 1;
 
@@ -461,11 +465,11 @@ mod in_memory {
 
                     if commit_to_apply.get_parent_count() > 1 {
                         warn!(
-                            ?commit_oid,
+                            ?commit_to_apply_oid,
                             "BUG: Merge commit should have been detected during planning phase"
                         );
                         return Ok(RebaseInMemoryResult::CannotRebaseMergeCommit {
-                            commit_oid: *commit_oid,
+                            commit_oid: *commit_to_apply_oid,
                         });
                     };
 
@@ -483,7 +487,7 @@ mod in_memory {
                         Ok(rebased_commit) => rebased_commit,
                         Err(CherryPickFastError::MergeConflict { conflicting_paths }) => {
                             return Ok(RebaseInMemoryResult::MergeConflict(MergeConflictInfo {
-                                commit_oid: *commit_oid,
+                                commit_oid: *commit_to_apply_oid,
                                 conflicting_paths,
                             }))
                         }
@@ -493,7 +497,7 @@ mod in_memory {
                     let commit_message = commit_message.to_str().ok_or_else(|| {
                         eyre::eyre!(
                             "Could not decode commit message for commit: {:?}",
-                            commit_oid
+                            commit_to_apply_oid
                         )
                     })?;
 
@@ -526,8 +530,8 @@ mod in_memory {
                         )?,
                     )?;
                     if rebased_commit.is_empty() {
-                        rewritten_oids.push((*commit_oid, MaybeZeroOid::Zero));
-                        maybe_set_skipped_head_new_oid(*commit_oid, current_oid);
+                        rewritten_oids.push((*original_commit_oid, MaybeZeroOid::Zero));
+                        maybe_set_skipped_head_new_oid(*original_commit_oid, current_oid);
 
                         writeln!(
                             effects.get_output_stream(),
@@ -537,8 +541,10 @@ mod in_memory {
                             commit_description
                         )?;
                     } else {
-                        rewritten_oids
-                            .push((*commit_oid, MaybeZeroOid::NonZero(rebased_commit_oid)));
+                        rewritten_oids.push((
+                            *original_commit_oid,
+                            MaybeZeroOid::NonZero(rebased_commit_oid),
+                        ));
                         current_oid = rebased_commit_oid;
 
                         writeln!(
@@ -954,6 +960,7 @@ pub struct ExecuteRebasePlanOptions<'a> {
 
 /// The result of executing a rebase plan.
 #[must_use]
+#[derive(Debug)]
 pub enum ExecuteRebasePlanResult {
     /// The rebase operation succeeded.
     Succeeded,
diff --git a/src/core/rewrite/plan.rs b/src/core/rewrite/plan.rs
index a3534a6..8233d14 100644
--- a/src/core/rewrite/plan.rs
+++ b/src/core/rewrite/plan.rs
@@ -44,7 +44,10 @@ pub enum RebaseCommand {
 
     /// Apply the provided commit on top of the rebase head, and update the
     /// rebase head to point to the newly-applied commit.
-    Pick { commit_oid: NonZeroOid },
+    Pick {
+        original_commit_oid: NonZeroOid,
+        commit_to_apply_oid: NonZeroOid,
+    },
 
     Merge {
         /// The original merge commit to copy the commit message from.
@@ -83,7 +86,10 @@ impl ToString for RebaseCommand {
         match self {
             RebaseCommand::CreateLabel { label_name } => format!("label {}", label_name),
             RebaseCommand::Reset { target } => format!("reset {}", target.to_string()),
-            RebaseCommand::Pick { commit_oid } => format!("pick {}", commit_oid),
+            RebaseCommand::Pick {
+                original_commit_oid: _,
+                commit_to_apply_oid: commit_oid,
+            } => format!("pick {}", commit_oid),
             RebaseCommand::Merge {
                 commit_oid,
                 commits_to_merge,
@@ -346,12 +352,15 @@ impl<'repo> RebasePlanBuilder<'repo> {
             } else {
                 // Normal one-parent commit (or a zero-parent commit?), just
                 // rebase it and continue.
-                let commit_oid = current_commit.get_oid();
-                let commit_oid = match self.replacement_commits.get(&commit_oid) {
+                let original_commit_oid = current_commit.get_oid();
+                let commit_oid = match self.replacement_commits.get(&original_commit_oid) {
                     Some(replacement_oid) => *replacement_oid,
-                    None => commit_oid,
+                    None => original_commit_oid,
                 };
-                acc.push(RebaseCommand::Pick { commit_oid });
+                acc.push(RebaseCommand::Pick {
+                    original_commit_oid,
+                    commit_to_apply_oid: commit_oid,
+                });
                 acc.push(RebaseCommand::DetectEmptyCommit {
                     commit_oid: current_commit.get_oid(),
                 });
@@ -714,7 +723,10 @@ impl<'repo> RebasePlanBuilder<'repo> {
                 | RebaseCommand::Reset { target: _ }
                 | RebaseCommand::RegisterExtraPostRewriteHook
                 | RebaseCommand::DetectEmptyCommit { commit_oid: _ } => None,
-                RebaseCommand::Pick { commit_oid }
+                RebaseCommand::Pick {
+                    original_commit_oid: _,
+                    commit_to_apply_oid: commit_oid,
+                }
                 | RebaseCommand::Merge {
                     commit_oid,
                     commits_to_merge: _,
diff --git a/tests/command/test_move.rs b/tests/command/test_move.rs
index dc69c0e..10a734c 100644
--- a/tests/command/test_move.rs
+++ b/tests/command/test_move.rs
@@ -67,15 +67,15 @@ fn test_move_stick() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
-                        message: None,
+                        original_commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
+                        commit_to_apply_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
-                        message: None,
+                        original_commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
+                        commit_to_apply_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
@@ -255,15 +255,15 @@ fn test_move_with_source_not_in_smartlog() -> eyre::Result<()> {
                             ),
                         },
                         Pick {
-                            commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
-                            message: None,
+                            original_commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
+                            commit_to_apply_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
                         },
                         DetectEmptyCommit {
                             commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
                         },
                         Pick {
-                            commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
-                            message: None,
+                            original_commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
+                            commit_to_apply_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
                         },
                         DetectEmptyCommit {
                             commit_oid: NonZeroOid(355e173bf9c5d2efac2e451da0cdad3fb82b869a),
@@ -362,8 +362,8 @@ fn test_move_merge_conflict() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(e85d25c772a05b5c73ea8ec43881c12bbf588848),
-                        message: None,
+                        original_commit_oid: NonZeroOid(e85d25c772a05b5c73ea8ec43881c12bbf588848),
+                        commit_to_apply_oid: NonZeroOid(e85d25c772a05b5c73ea8ec43881c12bbf588848),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(e85d25c772a05b5c73ea8ec43881c12bbf588848),
@@ -438,15 +438,15 @@ fn test_move_base() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
-                        message: None,
+                        original_commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
+                        commit_to_apply_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
-                        message: None,
+                        original_commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
+                        commit_to_apply_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(70deb1e28791d8e7dd5a1f0c871a51b91282562f),
@@ -571,8 +571,8 @@ fn test_move_checkout_new_head() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
-                        message: None,
+                        original_commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
+                        commit_to_apply_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
@@ -642,8 +642,8 @@ fn test_move_branch() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(98b9119d16974f372e76cb64a3b77c528fc0b18b),
-                        message: None,
+                        original_commit_oid: NonZeroOid(98b9119d16974f372e76cb64a3b77c528fc0b18b),
+                        commit_to_apply_oid: NonZeroOid(98b9119d16974f372e76cb64a3b77c528fc0b18b),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(98b9119d16974f372e76cb64a3b77c528fc0b18b),
@@ -832,8 +832,8 @@ fn test_move_in_memory_gc() -> eyre::Result<()> {
                         ),
                     },
                     Pick {
-                        commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
-                        message: None,
+                        original_commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
+                        commit_to_apply_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
                     },
                     DetectEmptyCommit {
                         commit_oid: NonZeroOid(96d1c37a3d4363611c49f7e52186e189a04c531f),
@@ -1632,8 +1632,8 @@ fn test_move_merge_commit() -> eyre::Result<()> {
                             ),
                         },
                         Pick {
-                            commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
-                            message: None,
+                            original_commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
+                            commit_to_apply_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
                         },
                         DetectEmptyCommit {
                             commit_oid: NonZeroOid(fe65c1fe15584744e649b2c79d4cf9b0d878f92e),
```

After this change, `cargo test reword` passes.

The above patch isn't sufficient to get it to work for on-disk rebases, but don't worry about that. I'll fix it myself when (if?) we have a reason to use `reword` with on-disk rebases.

> Sorry to be a bother and to need so much support on this!

It's no trouble at all, I'm excited for this feature to land! 😃

claytonrcarter:
Do you think this should have a basic "status report" after the reword, similar to what `git hide` shows? I'm implemented this locally, but wanted to check before I push it. It looks something like this:
```
...
In-memory rebase succeeded.
Reworded commit 34538cdd more file-b again REWORDED as 9b6f8ced asdf
Reworded commit 9337b169 edit file-b as 79a75f06 asdf
Reworded commit 5605e4e6 more file-c as 3f05d582 asdf
Reworded 3 commits with same message. If this was unintentional, run: git undo
```
(This format is totally up for discussion, but currently it's `"Reworded commit <old commit> as <new commit>"` w/ `friendly_describe()`, and the "Reworded 3 commits..." line is only printed if the command run on >1 commit.)

I ask b/c the most obvious implementation is to propagate `rewritten_oids` from `RebaseInMemoryResult::Succeeded` (which is only used w/i `execute.rs`) into `ExecuteRebasePlanResult::Succeeded` so that the `old commit <-> new commit` mapping is available w/i the command. This is straightforward, but means updating 3 other commands that don't need this data.

On one hand, I think displaying a list of what's happened is desirable, but I want to confirm that w/ you before I solidify it and push it.

For reference, here's the output w/ this while doing a rewording of 3 commits that results in rewriting a tree of 5 commits:
```
❯ git sl
⋮
◇ 9337b169 55d edit file-b
┣━┓
┃ ◯ 5d86daa1 55d add file-c
┃ ┃
┃ ● 5605e4e6 1d (branch-c) more file-c
┃
◇ 079ca218 20d (main) more file-b
┃
◯ 34538cdd 1d more file-b again REWORDED

❯ git reword -m asdf 34538cdd 9337b169 5605e4e6
Attempting rebase in-memory...
[1/5] Committed as: 79a75f06 asdf
[2/5] Committed as: b4893191 more file-b
[3/5] Committed as: 9b6f8ced asdf
[4/5] Committed as: 4367170e add file-c
[5/5] Committed as: 3f05d582 asdf
branchless: processing 2 updates: branch branch-c, branch main
branchless: processing 5 rewritten commits
branchless: running command: git checkout 3f05d582ea23c7582e87a82f8c1f24f2a9940c86
Previous HEAD position was 5605e4e more file-c
branchless: processing 1 update: ref HEAD
HEAD is now at 3f05d58 asdf
branchless: processing checkout
⋮
◇ 79a75f06 0s asdf
┣━┓
┃ ◯ 4367170e 0s add file-c
┃ ┃
┃ ● 3f05d582 0s (branch-c) asdf
┃
◇ b4893191 0s (main) more file-b
┃
◯ 9b6f8ced 0s asdf
In-memory rebase succeeded.
Reworded commit 34538cdd more file-b again REWORDED as 9b6f8ced asdf
Reworded commit 9337b169 edit file-b as 79a75f06 asdf
Reworded commit 5605e4e6 more file-c as 3f05d582 asdf
Reworded 3 commits with same message. If this was unintentional, run: git undo
```
```

**Comment by arxanas** [comment]
> Do you think this should have a basic "status report" after the reword, similar to what git hide shows?

Whatever you think is best is fine here 🙂. I think it's not necessary to show the smartlog, though, since it shouldn't have changed as part of the reword.

> I ask b/c the most obvious implementation is to propagate rewritten_oids from RebaseInMemoryResult::Succeeded (which is only used w/i execute.rs) into ExecuteRebasePlanResult::Succeeded so that the old commit <-> new commit mapping is available w/i the command. This is straightforward, but means updating 3 other commands that don't need this data.

That's perfectly fine. We'll probably use this data for something else in the future too.

**Comment by arxanas** [comment]
Also, see this regarding the failing Windows CI runs: https://github.com/arxanas/git-branchless/pull/306#issuecomment-1077134449

**Comment by claytonrcarter** [comment]
> Whatever you think is best is fine here

OK, I'm leaving it in, and I've disabled the smartlog. The format that I settled on is `Rewording commit <old oid> as <new oid> <new title>`, for example:
```
❯ /Users/crcarter/src/Other/git-branchless/target/debug/git-branchless reword c91c645c 40983c8a -m qwerty
Attempting rebase in-memory...
[1/2] Committed as: 838fb6c0 qwerty
[2/2] Committed as: 28174d68 qwerty
branchless: processing 1 update: branch branch-c
branchless: processing 2 rewritten commits
branchless: running command: git checkout branch-c
Previous HEAD position was 40983c8 zxvc
Switched to branch 'branch-c'
branchless: processing checkout
In-memory rebase succeeded.
Reworded commit c91c645c as 28174d68 qwerty
Reworded commit 40983c8a as 838fb6c0 qwerty
Reworded 2 commits with same message. If this was unintentional, run: git undo
```

I think I've taken care of all of the outstanding review items with my latest push. I would still like to do another full read through before I call it ready, but that is mostly to look for odd or non-idiomatic constructs, things I could simplify, etc. I believe that all of the features are in place and working.

There are a few outstanding `TODO`s: 2 of them are related to the as-yet-unimplemented bulk edit feature, and 2 more I'll call out specifically in PR comments. 

To confirm, are you OK if we leave the bulk edit feature out of this and add it later, after this has been merged? Or would you prefer to go for broke and add bulk editing now? On one hand, I think it would be nice to get this merged into master so folks could try it out as is; on the other hand, I don't think that bulk editing will be *too* difficult.

File: src/commands/reword.rs [hunk 10]
```diff
@@ -0,0 +1,314 @@
...
+                match msg {
+                    Ok(msg) => msg,
+                    _ => {
+                        return Ok(BuildRewordMessageResult::Failed {
+                            message: Some(String::from(
+                                "Reword: Error decoding original commit message!",
+                            )),
+                        })
+                    }
+                }
```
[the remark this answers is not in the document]
claytonrcarter: I tried this and couldn't get it to work. I was getting an error about `wrap_err` not satisfying a trait constraint related to `OsString`, and I have to admit that I just didn't understand what it was getting at. I ended up this:
```rust
            [commit] => match commit.get_message_raw()?.into_string() {
                Ok(msg) => msg,
                _ => eyre::bail!(
                    "###
                    Error decoding original message for commit: {:?}.
                    Stopping; nothing reworded.
                    ###",
                    commit.get_oid()
                ),
            },
```

src/core/rewrite/execute.rs
```diff
@@ -954,9 +971,14 @@ pub struct ExecuteRebasePlanOptions<'a> {
...
 #[must_use]
+#[derive(Debug)]
 pub enum ExecuteRebasePlanResult {
     /// The rebase operation succeeded.
-    Succeeded,
+    Succeeded {
+        /// Mapping from old OID to new/rewritten OID. Will always be empty for on disk rebases.
+        // TODO should this be an Option?
+        rewritten_oids: HashMap<NonZeroOid, MaybeZeroOid>,
+    },
```
claytonrcarter, line 981: Regarding this `TODO`: only in mem rebases will be using this property, b/c on disk rebases don't have access to (as far as I can tell) the list of rewritten commits, and are currently only ever using `Default::default()` for this property, but that could leave us unable to tell if a `Succeeded` result actually rewrote 0 commits, or if it's just a misleading results from an on disk rebase. Using an `Option<...>` here could solve that b/c we can then say that all on-disk rebases use `None`.

Ok, so that said, I like the `Option<...>` feels more correct, but given the very limited usage of this (at this time), I suspect that this is OK as is. What do you think?

tests/command/test_reword.rs
```diff
@@ -0,0 +1,297 @@
...
+    // try adding several messages that start w/ '#'
+    git.run(&["reword", "-m", "foo", "-m", "# bar", "-m", "#", "-m", "buz"])?;
+
+    // confirm the '#' messages aren't present
+    let (stdout, _stderr) = git.run(&["log", "-n", "1", "--format=%h%n%B"])?;
+    insta::assert_snapshot!(stdout, @r###"
+    2666020
+    foo
+
+    buz
```
claytonrcarter, line 125: I just added this test today, to test that we're removing comment lines. See other comment about how comments are handled, though.

src/commands/reword.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        for root_commit in subtree_roots {
+            let only_parent_id = root_commit.get_only_parent().map(|parent| parent.get_oid());
+            if only_parent_id == None {
+                eyre::bail!(
+                    "###
+                    Refusing to reword commit {}, which has {} parents.
+                    Rewording only supported for commits with 1 parent.
+                    Stopping; nothing reworded.
+                    ###",
```
claytonrcarter, line 86: So, some of my error messages are kind of long. Am I being too verbose w/ these?

src/commands/reword.rs
```diff
@@ -0,0 +1,354 @@
...
+                    commit.get_oid()
+                ),
+            },
+            _ => {
+                // TODO(bulk edit) build a bulk edit message for multiple commits
+                format!(
+                    "Enter a commit message to apply to {} commits",
+                    commits.len()
+                )
```
claytonrcarter: It occurs to me that we might want to have a more descriptive default message. Perhaps we should be appending something like the default commit message when prepping the string for the Editor:

```
# Please enter the commit message for your changes. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
```

This would be easy to adapt from `commit` to `reword`.

src/commands/reword.rs
```diff
@@ -0,0 +1,354 @@
...
+            return Ok(BuildRewordMessageResult::IdenticalMessage);
+        }
+        edited_message
+    } else {
+        message
+    };
+
+    // clean up the commit message: remove comment lines, finish w/ a newline, etc
+    // TODO ask if this should be here or if it should be moved to the git module per project wiki
+    let message = git2::message_prettify(message, comment_char)?;
```
claytonrcarter, line 261: 2 things:

1. The [wiki](https://github.com/arxanas/git-branchless/wiki/Coding#the-git-module) states that git2 code should not be used directly. Would you prefer that I move this into our own git module, or is this trivial enough that it can probably stay here?
2. By doing this processing here, vs up a few lines after the Editor returns, it means that we'll be prettifing *all* messages, including those submitting on the CLI. For example, `git reword -m '#123 is fixed'` will be removed and the user will get an "empty message" notice that nothing was reworded. I *think* that this is OK for consistency and simplicity, but I do have a feeling that suggests that if someone puts a message like that in via CLI that they really intend for it work. (Even though such a message in the editor would not work.) It would be easy enough (I think) to move this, but it does mean that we won't be able to test this. (We currently aren't testing the Editor code path.)

src/commands/reword.rs
arxanas: No, I think these error messages are quite good. (They adhere to Toulmin-style argumentation as suggested in https://www.microsoft.com/en-us/research/publication/how-should-compilers-explain-problems-to-developers/)

Stylistically, we use `Aborting` everywhere else in the codebase so far, so let's be consistent about that. (I think you could argue that `Aborting.` would be enough here, and the `nothing reworded` would be implied. It's up to you if you want to remove that second bit or not.)

```suggestion
                    "###
                    Refusing to reword commit {}, which has {} parents.
                    Rewording is only supported for commits with 1 parent.
                    Aborting; nothing reworded.
                    ###",
```

arxanas: Also, let's write this message to `effects.get_error_stream()` and return `1` as the exit code. `eyre::bail!` should be reserved for programming errors.

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+                    "###
+                    Refusing to reword commit {}, which has {} parents.
+                    Rewording only supported for commits with 1 parent.
+                    Stopping; nothing reworded.
+                    ###",
+                    root_commit.get_oid(),
+                    root_commit.get_parents().len(),
+                );
+            }
+            builder.move_subtree(root_commit.get_oid(), only_parent_id.unwrap())?;
```
arxanas, line 91: Use `match` instead of `unwrap`, like this:

```suggestion
            let only_parent_id = match only_parent_id {
                Some(only_parent_id) => only_parent_id,
                None => {
                    eyre::bail!(...); // or `return`
                }
            }
            builder.move_subtree(root_commit.get_oid(), only_parent_id)?;
```

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+            &repo_pool,
+            &BuildRebasePlanOptions {
+                dump_rebase_constraints: false,
+                dump_rebase_plan: false,
+                detect_duplicate_commits_via_patch_id: false,
+            },
+        )? {
+            Ok(Some(rebase_plan)) => rebase_plan,
+            Ok(None) => {
+                writeln!(
```
arxanas, line 114: This is actually a case where we could use `eyre::bail!` 😂 (but this is certainly fine too).

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+            render_smartlog: false,
+        },
+    };
+    let result = execute_rebase_plan(effects, git_run_info, &repo, &rebase_plan, &execute_options)?;
+
+    let exit_code = match result {
+        ExecuteRebasePlanResult::Succeeded { rewritten_oids } => {
+            render_status_report(&repo, effects, &commits, &rewritten_oids)?;
+
+            0
```
arxanas, line 150:
```suggestion
            render_status_report(&repo, effects, &commits, &rewritten_oids)?;
            0
```

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+#[instrument]
+fn build_messages(
+    messages: &[String],
+    commits: &[Commit],
+    comment_char: Option<u8>,
+) -> eyre::Result<BuildRewordMessageResult> {
+    let message = messages.join("\n\n").trim().to_string();
+
+    let (message, load_editor) = if message.is_empty() {
+        let message_for_editor = match commits.to_vec().as_slice() {
```
arxanas, line 217: Oops, `commits` was already a slice, so we didn't need to call `as_slice()` on it at all.

```suggestion
        let message_for_editor = match commits {
```

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+            [commit] => match commit.get_message_raw()?.into_string() {
+                Ok(msg) => msg,
+                _ => eyre::bail!(
```
arxanas, line 220: It's better to use exhaustive pattern-matches when possible.

```suggestion
                Err(_) => eyre::bail!(
```

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+                    Stopping; nothing reworded.
+                    ###",
+                    commit.get_oid()
+                ),
+            },
+            _ => {
+                // TODO(bulk edit) build a bulk edit message for multiple commits
+                format!(
+                    "{} Enter a commit message to apply to {} commits",
+                    comment_char.unwrap(),
```
arxanas, line 232: Won't the leading comment here be a number like `20` instead of the actual character? Oh, I guess this is the bulk-editing case, so it's not fleshed out yet.

src/commands/reword.rs
arxanas: > Would you prefer that I move this into our own git module

Let's move this into our own `git` module. In particular, the wrapper should have `#[instrument]`, or else it won't appear in span-traces.

> It would be easy enough (I think) to move this, but it does mean that we won't be able to test this

I think that's fine.

src/commands/reword.rs
```diff
@@ -0,0 +1,358 @@
...
+                continue;
+            }
+        };
+        let replacement_commit = repo.find_commit(*replacement_oid)?.unwrap();
+        writeln!(
+            effects.get_output_stream(),
+            "Reworded commit {} as {}",
+            printable_styled_string(
+                &glyphs,
+                // Commit doesn't offer `friendly_describe_oid`, so we'll do it ourselves
```
arxanas, line 336: 👍

src/core/config.rs
```diff
@@ -30,6 +30,18 @@ pub fn get_main_branch_name(repo: &Repo) -> eyre::Result<String> {
...
+#[instrument]
+pub fn get_comment_char(repo: &Repo) -> eyre::Result<Option<u8>> {
+    let config = repo.get_readonly_config()?;
+    let comment_char: Option<String> = config.get("core.commentChar")?;
+    let comment_char = match comment_char {
+        Some(c) => Some(c.chars().next().unwrap() as u8),
+        None => git2::DEFAULT_COMMENT_CHAR,
+    };
+    Ok(comment_char)
+}
```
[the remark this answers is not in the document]
arxanas: Yeah, let's just leave the `auto` case unimplemented for now 😂

src/core/rewrite/execute.rs
```diff
@@ -833,6 +840,16 @@ mod on_disk {
...
 
+        if rebase_plan.commands.iter().any(|command| match command {
+            RebaseCommand::Pick {
+                original_commit_oid,
+                commit_to_apply_oid,
+            } => original_commit_oid != commit_to_apply_oid,
+            _ => false,
+        }) {
+            eyre::bail!("Not implemented: replacing commits in an on disk rebase");
+        }
```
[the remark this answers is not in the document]
arxanas: This seems fine.

src/core/rewrite/execute.rs
arxanas: > on disk rebases don't have access to (as far as I can tell) the list of rewritten commits

The on-disk rebases use the `post-rewrite` hook to get access to this information, but it's a complicated workflow, and the information isn't returned synchronously. I think using an `Option` here is warranted to make that clear, and we can `unwrap` it on the other end.

arxanas (approved):
> I haven't done anything to look at @epage request that it https://github.com/arxanas/git-branchless/issues/179#issuecomment-955109646. At this time, I don't even know where to start w/ that, so any pointers would be most welcome.

I think `git reword` should actually work as-is without `git branchless init` (although it will lazily construct the repository commit graph so that it knows how to rebase descendant commits). However, I think there's a bug and it'll only work if the main branch happens to be named `master`. If that's the case, it should be fixed by changing `get_main_branch_name` to delegate to `get_default_branch_name`, rather than the hardcoded `master` value.

To test this, try just running `git branchless reword` in a repository where you haven't run `git branchless init` and see what happens.

>  The bulk editing feature is not implemented; I am voting ✋ for us to review this as is and add that later, though I'm also fine to push on to work on bulk editing now if you'd prefer.

Certainly, let's defer that.

>  During my latest read through, I decided that we should probably a "footer" message to the $EDITOR akin to the default commit message from git. I'll work on this soon.

We can also ship what you have now and you can follow up in future PRs if you like. Besides a few minor points above, I think this is good to ship when you're ready.

src/commands/reword.rs
claytonrcarter: Done on all points, thank you. I've changed "Stopping; nothing reworded" to "Aborting"; it's more succinct and matches git's own language.

src/commands/reword.rs
claytonrcarter: Ah, of course!

src/commands/reword.rs
claytonrcarter: 🤦 Yes. The dangers of pushing code late at night. I've fixed this, but also rewritten this whole message in my latest round. Thanks for catching this.

src/commands/reword.rs
claytonrcarter: OK. I moved that into into the `repo` mod, which was the closest/most relevant that I could find. Happy to move elsewhere if that's not right. This also gave me the chance to update the `get_comment_char` to return a `char`, which feels more natural than the funny feeling (to me) `Option<u8>` that git2 was using.

tests/command/test_reword.rs
claytonrcarter: I updated this to confirm that we *are* allowing "comments" from the CLI.

src/core/rewrite/execute.rs
claytonrcarter: This is an `Option` now.

src/core/config.rs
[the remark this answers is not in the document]
claytonrcarter: Great. FWIW I added a comment/note about this.

src/commands/reword.rs
claytonrcarter: OK: "stopping" -> "aborting" and I changed this to use the error stream instead of `bail!`.

src/commands/reword.rs
claytonrcarter: I added this "help" message to Editor mode, and I'm glad I did: I had been *completely* misunderstanding how Rust handles newlines and whitespace in strings. I was using some sort of `"### ... ###"` syntax based on the snapshots in the tests b/c I thought that was some special Rust syntax to "dedent" those strings (ie remove leading whitespace from each line), but hadn't realized that it's just `insta` that's doing that. 

That was all sorts of wrong and I've rewritten my multi-line strings using `\n\` syntax, which actually achieves what I was going for initially.

claytonrcarter:
> To test this, try just running git branchless reword in a repository where you haven't run git branchless init and see what happens.

Yup, good call. I confirmed this and fixed it. I also confirmed that `reword` works as expected w/o `init`. 🚀

claytonrcarter:
OK, I think I taken care of the latest round of review comments. Note that I have squashed all of the last round of reviewed work into 25346b9d, and all new work is built on that. I gave everything a read through and think that this is good to go. I don't know what else to add or (more importantly) what else to remove. 😄

I'll leave a few more comments to draw your attention to a couple of details, but I think we can call this done.

src/core/config.rs
```diff
@@ -14,20 +14,41 @@ pub fn get_core_hooks_path(repo: &Repo) -> eyre::Result<PathBuf> {
...
+
+    if let Some(branch_name) = config.get("branchless.mainBranch")? {
+        return Ok(branch_name);
+    }
+
+    if let Some(branch_name) = get_default_branch_name(repo)? {
+        return Ok(branch_name);
+    }
+
+    Ok("master".to_string())
```
claytonrcarter, line 39: I had originally written this w/ nested matches, but I think they were 3 deep (b/c we had to unwrap 3 different config Options. I changed it to this which -- to me -- feels easier to grok at a glance. I might not be very rust-y, though. I can revert to something like that if you'd prefer

arxanas:
🚀 HYPE 🚀

Issue #280, opened 2022-02-24
Use git2 hash-shortening function


arxanas:
### Discussed in https://github.com/arxanas/git-branchless/discussions/279

<div type='discussions-op-text'>

<sup>Originally posted by **KevinWuWon** February 21, 2022</sup>
A few times, the 8-character hash has been too short and it has given me a collision. It's quite annoying when this happens because it just gives me an error message like:

```
Git error Ambiguous: ambiguous SHA1 prefix - multiple matches for prefix: 1e0397ec4cc2148c6587e30bc85baae8b3f26002 1e0397ecb633e9115514b7c256560ddc27594289
```

and I still don't know which one I want.

My two feature requests are:

1. in this error message, print also the commit title so I can tell them apart
2. give an option to increase the default hash prefix length to display in `git sl`</div>

yujong-lee:
@arxanas  Maybe we can close it by #306?

arxanas:
@yujong-lee yes, thanks for fixing this!

PR #306, opened 2022-03-23
Use git2 hash-shortening function

yujong-lee:
I am interested in this project and I want to contribute to code although I'm not good at rust. I think I can be better by contributing this project.

Can you help me with this?

I was trying to pick up #280.

I thought `oid` in this function should be calculated using `git2`.
https://github.com/arxanas/git-branchless/blob/45ace45579804697f8daeb551750424bf8e8538b/src/core/node_descriptors.rs#L183-L199

Then I might have to change constructor.

https://github.com/arxanas/git-branchless/blob/45ace45579804697f8daeb551750424bf8e8538b/src/core/node_descriptors.rs#L178

So my current status in ccc277c37c6f138bb17e2d0ac19898bcc82fc3f3 is I brought `&Repo` to constructor.
```rust
    pub fn new(repo: &Repo, use_color: bool) -> eyre::Result<Self> {
        Ok(CommitOidDescriptor { use_color })
    }
```

And I got 3 problems.

1. I was thinking [this](https://github.com/arxanas/git-branchless/discussions/279#discussioncomment-2221208), but  `repo: &Repo` don't have `raw()`. Which is strange. Because it is wrapper around `git2::Repository` and I thought `git2::Repository` has `raw()`.
2.  To enable `describe_node` to access to `repo` as `self.repo`, I need to do something like this.
```rust
pub struct CommitOidDescriptor<'a> {
    repo: &'a Repo,
    use_color: bool,
}
```
But It makes error message all around. If this is right way to do it, then I will study about `lifetime` and go forward. But I am not sure about it because of `1`.

3. `bugreport!` error.  I don't know what should I do.
스크린샷 2022-03-23 오후 2 05 49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - (#267) Aliases like `git amend` are now installed only if the user does not already have aliases with the same name. Thanks to @rslabbert for implementing this.
 - Improved performance up to 15x for `git restack` on large commit histories.
+- (#280) Use git2's get_short_oid function to avoide commit hash collision.
 
 ## [0.3.9] - 2022-02-08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - (#267) Aliases like `git amend` are now installed only if the user does not already have aliases with the same name. Thanks to @rslabbert for implementing this.
 - Improved performance up to 15x for `git restack` on large commit histories.
-- (#280) Use git2's get_short_oid function to avoide commit hash collision.
+- (#280) Ambiguous commit hashes are no longer printed in output. (Additional characters of the hash will be appended as necessary.) Thanks to @yujonglee for fixing this.
 
 ## [0.3.9] - 2022-02-08
 
--- a/src/core/node_descriptors.rs
+++ b/src/core/node_descriptors.rs
@@ -55,6 +55,16 @@ impl<'repo> NodeObject<'repo> {
             NodeObject::GarbageCollected { oid } => *oid,
         }
     }
+
+    fn get_short_oid(&self) -> eyre::Result<String> {
+        match self {
+            NodeObject::Commit { commit } => Ok(commit.get_short_oid()?),
+            NodeObject::GarbageCollected { oid } => {
+                // `7` is the default value for config setting `core.abbrev`.
+                Ok(oid.to_string()[..7].to_string())
+            }
+        }
+    }
 }
 
 /// Object responsible for redacting sensitive information, so that it can be
@@ -187,8 +197,7 @@ impl NodeDescriptor for CommitOidDescriptor {
         _glyphs: &Glyphs,
         object: &NodeObject,
     ) -> eyre::Result<Option<StyledString>> {
-        let oid = object.get_oid();
-        let oid = &oid.to_string()[..8];
+        let oid = object.get_short_oid()?;
         let oid = if self.use_color {
             StyledString::styled(oid, BaseColor::Yellow.dark())
         } else {
--- a/tests/command/test_undo.rs
+++ b/tests/command/test_undo.rs
@@ -99,7 +99,7 @@ fn test_undo_help() -> eyre::Result<()> {
         )?;
         insta::assert_snapshot!(screen_to_string(&screenshot1), @r###"
         ┌───────────────────────────────────────────────────┤ Commit graph ├───────────────────────────────────────────────────┐
-        │O f777ecc9 (master) create initial.txt                                                                                │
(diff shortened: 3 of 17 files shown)


arxanas:
Discussed further at https://discord.com/channels/915309546984050709/940852244880965673/956060354595676170

src/core/node_descriptors.rs
```diff
@@ -187,8 +194,7 @@ impl NodeDescriptor for CommitOidDescriptor {
         _glyphs: &Glyphs,
         object: &NodeObject,
     ) -> eyre::Result<Option<StyledString>> {
-        let oid = object.get_oid();
-        let oid = &oid.to_string()[..8];
```
yujong-lee, line 191: Is removing `[..8]` ok?

arxanas: Let's move the `[..8]` case to `NodeObject::get_short_id` in the `GarbageCollected` case. In that case, it doesn't really matter what OID we show, since the object is no longer available anyways, so we might as well keep it short to make it easy to read the output.

arxanas (approved):
Looks good to me other than that minor change about `[..8]`. Besides that:

- I'll make sure CI is passing before we merge. You can run `cargo test` and `cargo clippy` locally to confirm.
- Combine your commits into a single commit. The commit message should start with `feat(core): ` in the style of some of the other commit messages in the repository history.
- Update `CHANGELOG.md` with your change.

src/core/node_descriptors.rs
```diff
@@ -55,6 +55,13 @@ impl<'repo> NodeObject<'repo> {
             NodeObject::GarbageCollected { oid } => *oid,
         }
     }
+
+    fn get_short_oid(&self) -> eyre::Result<String> {
+        match self {
+            NodeObject::Commit { commit } => Ok(commit.get_short_oid()?),
+            NodeObject::GarbageCollected { oid } => Ok(oid.to_string()[..8].to_string()),
```
yujong-lee, line 62: After this, every commit id in snapshots is shorted. (length `8 -> 7`). And I think it is strange

claytonrcarter: I believe that's the point: it means that it's working correctly! 

The docs for [`short_id`](https://docs.rs/git2/0.8.0/git2/struct.Object.html#method.short_id) say:
> Get a short abbreviated OID string for the object
> 
> This starts at the "core.abbrev" length (default 7 characters) and iteratively extends to a longer string if that length is > ambiguous. The result will be unambiguous (at least until new objects are added to the repository).

yujong-lee: Yes you are totally right. I thought it will generate something short but still 8+ length. I should have check docs. Thanks for sharing.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - (#267) Aliases like `git amend` are now installed only if the user does not already have aliases with the same name. Thanks to @rslabbert for implementing this.
 - Improved performance up to 15x for `git restack` on large commit histories.
+- (#280) Use git2's get_short_oid function to avoide commit hash collision.
```
arxanas, line 26:
```suggestion
- (#280) Use `git2`'s `short_id` function to disambiguate commit hashes when printed.
```

arxanas (approved):
Looks pretty good! I think there are still a few failing tests, like here: https://github.com/arxanas/git-branchless/runs/5670307173?check_suite_focus=true

After they're resolved, we can merge this.

arxanas:
By the way, the Windows CI tests are failing with this error:

```
Message:  attempt to add with overflow
Location: C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\indicatif-0.17.0-rc.9\src\state.rs:420
```

This is an unrelated issue and you can ignore it. Only the Linux runs are important for this change.

tests/command/test_undo.rs
```diff
@@ -455,7 +455,7 @@ fn test_historical_smartlog_visibility() -> eyre::Result<()> {
         insta::assert_snapshot!(screen_to_string(&screenshot1), @r###"
```
martinvonz, line 455: If you look at the line above this one (it seems GitHub doesn't let me comment on it because it's not modified), you'll see the answer to why this passes locally and fails on CI: CI's git version apparently does not support reference transactions. So it's probably a good idea if you can find a git version that doesn't and run the tests locally with that version as well. The CI logs indicate that it's using git-v2.24.3, so try installing that and pointing `TEST_GIT` and `TEST_GIT_EXEC_PATH` to it.

CHANGELOG.md
martinvonz: What is the changelog's target audience? If it's end users, then it probably doesn't make sense to mention `git2` here. Say something about always making them unambiguous and to respect `core.abbrev` (if I understood correctly what this PR means to the end user).

arxanas: @martinvonz good point; I've updated the changelog with a more user-facing description. (Oops, I forgot to mention `core.abbrev`, but it's probably not too important.)

PR #310, opened 2022-03-25
tests: don't propagate environment into `GitRunInfo`

martinvonz:
When I ran the tests with `LANG=sv_SE.UTF-8` exported in my shell,
`test_run_silent_failures` failed. This patch fixes it. yujong-lee@
reported many more language-related failures but I wasn't able to
reproduce those.


arxanas:
Nice, thanks for fixing this and cleaning this up. I think there was a time when we were relying on inheriting the parent environment when spawning subprocesses, but thankfully we now manually forward only the environment variables we care about.

PR #312, opened 2022-03-26
build: warn about all `as` conversions

arxanas:


claytonrcarter:
\#TIL and sorry/thanks!

arxanas:
@claytonrcarter It's not your fault, I was meaning to do this anyways 🙂. It's not even the case that `as`-conversions are generally accepted to be bad practice; see https://github.com/console-rs/indicatif/pull/406#issuecomment-1077067342.

PR #322, opened 2022-03-29
Support using `commit.template` in `git reword`

arxanas:


arxanas:
@claytonrcarter In this PR, I added a `-c`/`--clear` flag, but I wonder what you think would be the best API? I also considered `-t`/`--template`. I'm not sure if it's more intuitive to have one or two mechanisms for starting from an empty commit message vs starting from the commit template.

src/commands/reword.rs
```diff
@@ -52,7 +62,19 @@ pub fn reword(
...
+        // Editor::edit() normally requires that the file being edited is saved before the editor
+        // is closed. If it's not, it will return None; and in all other cases, it will return
+        // Some(message). We don't care if the file has been saved, only if it hasn't changed, so
+        // here we call `require_save(false)` and just ignore the None case.
+        let result = Editor::new()
+            .require_save(false)
+            .edit(message)?
+            .expect("`Editor::edit` should not return `None` when `require_save` is `false`");
+        Ok(result)
+    };
```
claytonrcarter, line 75: Sorry if I'm being dense, but what does this closure do for us vs just leaving it inline? I don't see it being used anywhere else, or being swapped out for another closure or something like that. Does it have something to do w/ the `#[instrument(skip(edit_message_fn))]` line?

claytonrcarter:
This makes a ton of sense. I think the args you've suggested are fine, but I wonder if `--discard` might be a contender, as in "discard the existing message and start over".

src/commands/reword.rs
arxanas: I just thought it would be hard to read the next line, since it's the subject of the `match` expression.

arxanas:
I updated the flag to be `-d`/`--discard`.

PR #331, opened 2022-04-01
feat(reword): use `$GIT_EDITOR` and `core.editor`, if configured (close #328)

### LINKED ISSUE #328: `git reword` doesn't respect git config's `core.editor`
### Description of the bug

When I `git reword <hash>` vim is used. `EDITOR` is not exported on my machine so this aligns with the documentation.

Note: my simple workaround was: `export EDITOR=<editor> && git reword <hash>`

### Expected behavior

I expect the same editor to be used for `git commit` and `git branchless reword`.

In other words: `$EDITOR` is used as a fallback, but `core.editor` takes precedence.

### Actual behavior

vim opens regardless

restack PR (#326)

### Version of `git`

2.32.0

### Version of `rustc`

1.59.0

### Automated bug report

_No response_

claytonrcarter:
This is my attempt to address #328.

There are no tests for the code path that involves the editor, but I was able to test this myself manually via:
```
$ cargo run -- reword                   # used `vi`
$ GIT_EDITOR=nano cargo run -- reword   # used `nano`
$ git config core.editor nano
$ cargo run -- reword                   # used `nano`
$ git config --unset core.editor
$ cargo run -- reword                   # used `vi`
```

### DIFF
--- a/src/commands/reword.rs
+++ b/src/commands/reword.rs
@@ -11,7 +11,9 @@ use eden_dag::DagAlgorithm;
 use eyre::Context;
 use tracing::{instrument, warn};
 
-use crate::core::config::{get_comment_char, get_commit_template, get_restack_preserve_timestamps};
+use crate::core::config::{
+    get_comment_char, get_commit_template, get_editor, get_restack_preserve_timestamps,
+};
 use crate::core::dag::{resolve_commits, CommitSet, Dag, ResolveCommitsResult};
 use crate::core::effects::Effects;
 use crate::core::eventlog::{EventLogDb, EventReplayer};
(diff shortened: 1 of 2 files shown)


src/commands/reword.rs
```diff
@@ -63,11 +65,12 @@ pub fn reword(
...
-        // is closed. If it's not, it will return None; and in all other cases, it will return
-        // Some(message). We don't care if the file has been saved, only if it hasn't changed, so
-        // here we call `require_save(false)` and just ignore the None case.
-        let result = Editor::new()
+        let mut editor = Editor::new();
+        let editor = match get_editor(&repo)? {
+            Some(prog) => editor.executable(prog),
+            None => &mut editor,
+        };
+        let result = editor
```
claytonrcarter, line 73: This works just fine, but I'm 100% open to suggestions if there's a more idiomatic way to handle optional chaining calls in Rust. The only other thing that came to my mind was to have `get_editor` do all of the resolution, and then just always call `.executable(get_editor(&repo)?.unwrap())` or similar here. Doing all of the resolution would probably be fine, but it felt like overkill since dialoguer already does some of it, and my implementation would just be copying from them, I think. 😄

arxanas: This seems good to me. Let's not duplicate the logic inside `Editor` itself. I also notice that builder patterns tend to be a little clunky in Rust.

src/core/config.rs
```diff
@@ -90,6 +91,29 @@ pub fn get_default_branch_name(repo: &Repo) -> eyre::Result<Option<String>> {
...
+    if editor != None {
+        return Ok(editor);
+    }
+
+    let config = repo.get_readonly_config()?;
+    let editor: Option<String> = config.get("core.editor")?;
+    match editor {
+        Some(editor) => Ok(Some(editor.into())),
+        None => Ok(None),
+    }
```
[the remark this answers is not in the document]
arxanas: That seems sensible. It's also possible to use `.map` for the last `match` expression, but it's not obviously better.

arxanas (approved):
Thanks for fixing!

PR #343, opened 2022-04-10
Preliminary git-record library

arxanas:


arxanas:
Currently, it looks like this:

Screen_Shot_2022-04-07_at_11 40 34_PM

It needs a lot of work. The core types aren't completely ironed out yet, either. (In particular, to support three-way merges, I think I need to adjust the before/after collections to instead be upstream/parent/local.)

git-record/src/ui.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+        let RecordState { files } = &self.state;
+
+        let count_changed_hunks = |&FileHunks { ref hunks }| {
+            hunks
+                .iter()
+                .filter(|hunk| match hunk {
+                    Hunk::Unchanged { .. } => false,
+                    Hunk::Changed { .. } => true,
+                })
```
martinvonz, line 109: I think you can use `matches!(hunk, Changed { .. })` instead of the `match hunk { ... }` if you like

git-record/src/ui.rs
```diff
@@ -0,0 +1,1041 @@
...
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct FileKey {
+    file_num: usize,
+}
+
+impl Display for FileKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Self { file_num } = self;
+        write!(f, "FileKey({})", file_num)
```
martinvonz, line 30: The documentation for `Display` says:

> Display is similar to Debug, but Display is for user-facing output, and so cannot be derived.

How about making this a custom function instead?  (Maybe `view_name()` or `view_id()` or something?)

git-record/src/ui.rs
```diff
@@ -0,0 +1,1041 @@
...
+                                        // Do nothing.
+                                    }
+                                }
+                            })
+                            .with_name(file_key.to_string()),
+                    )
+                    .child(TextView::new({
+                        let mut s = StyledString::new();
+                        s.append_plain(" /");
+                        s.append_styled(path.to_string_lossy(), Effect::Bold);
```
martinvonz, line 196: I wonder if the path should be a `String` to start with. Does this crate need it to be a `PathBuf`?

martinvonz (commented):
Nice, this looks like a great start! I think we'll want to hide most context lines by default (as I saw that you also mentioned in one of the commit). I'm not sure how to best present that.

Cargo.toml
```diff
@@ -5,7 +5,7 @@ debug = 0
 
 [workspace]
 default-members = ["git-branchless"]
-members = ["git-branchless", "git-branchless-lib"]
+members = ["git-branchless", "git-branchless-lib", "git-record"]
```
epage, line 8: I find it easiest if I put all crates in a `crates/` and for this I do `"crates/*"`

arxanas: I'll keep it in mind; for now, I don't want to increase the directory nesting level too much if I can avoid it.

git-record/src/ui.rs
arxanas: In this case, I want to preserve exhaustiveness-checking.

git-record/src/ui.rs
arxanas: Makes sense to me.

PR #344, opened 2022-04-10
ci: build and test all packages in the workspace

arxanas:


.github/workflows/linux.yml
```diff
@@ -79,19 +79,29 @@ jobs:
...
 
       - name: Run Rust tests (all features)
         timeout-minutes: 10
         run: |
           export TEST_GIT="$PWD"/git
           export TEST_GIT_EXEC_PATH=$(dirname "$TEST_GIT")
-          cargo test --all-features
+          cargo test --all-features --examples --tests --workspace
+
+      - name: Run Rust doc-tests (all features)
```
martinvonz, line 92: Why not combine this with the test run above? You could just add `--doc` to line 90, right?

.github/workflows/linux-git-devel.yml
```diff
@@ -64,11 +64,11 @@ jobs:
         run: |
           export TEST_GIT="$PWD"/git-master/git
           export TEST_GIT_EXEC_PATH=$(dirname "$TEST_GIT")
-          (cd git-branchless && cargo test --all-features)
+          (cd git-branchless && cargo test --all-features --examples --tests --workspace)
```
martinvonz, line 67: Why not use `--all-targets` here so you don't need to update it if you add doc tests later?

.github/workflows/linux.yml
arxanas: Nope, you can't combine them for some reason 😢

.github/workflows/linux-git-devel.yml
arxanas: I tried that at first, but it also runs benchmarks, which in this case panic because the environment isn't set up.

.github/workflows/linux.yml
martinvonz: Hmm, that's surprising. I'm curious why (but I don't expect you to answer that :) )

.github/workflows/linux-git-devel.yml
epage: [`--all-targets`  doesn't run doctests](https://github.com/rust-lang/cargo/issues/6669)

.github/workflows/linux-git-devel.yml
```diff
@@ -64,11 +64,11 @@ jobs:
...
 
       - name: Run Rust tests on Git `next`
         timeout-minutes: 10
         run: |
           export TEST_GIT="$PWD"/git-next/git
           export TEST_GIT_EXEC_PATH=$(dirname "$TEST_GIT")
-          (cd git-branchless && cargo test --all-features)
+          (cd git-branchless && cargo test --all-features --examples --tests --workspace)
```
epage, line 74: There a reason you `cd` rather than doing `-p git-branchless`?

PR #375, opened 2022-04-26
build: mark `Cargo.lock` as `binary` for merging

arxanas:


.gitattributes
```diff
@@ -0,0 +1 @@
+Cargo.lock merge=binary
```
epage, line 1: > Otherwise, Git tries to resolve the conflict textually and introduces conflict markers, which make `cargo` unable to process the file (and therefore unable to resolve the conflicts automatically, such as with `cargo update`).

Started looking into this to consider whether I should change my workflow and/or suggest adding this to `cargo new`.  My workflow is to just delete the conflicting `Cargo.lock` and re-generate it.  The benefit to allowing merges is Github is more likely to merge the result when handling a lot of Dependabot updates (really wish there was an option to batch update lock file updates).

What do you like about your workflow?

As a side note, cargo does have the following feature request that would be great to have:  https://github.com/rust-lang/cargo/issues/1818

arxanas: @epage The workflow problem is that often I enter a state of merge conflict and want to immediately run `cargo check`/`cargo test`/etc., but those commands fail outright because of the presence of merge conflict marks in `Cargo.lock`. (`cargo` doesn't attempt to automatically regenerate it in that case.)

It's strictly more steps to resolve when textually merging than when binary merging. With the binary merge, you can just run `cargo update` without having to delete the file first.

In principle, it would be incorrect to accept the textual merge of two `Cargo.lock` files even if it happened to merge cleanly, although in practice, the checksums in the `.lock` file should prevent it from being accepted by `cargo`.

PR #376, opened 2022-04-27
refactor(dag): `CommitSet::from_iter()` -> `collect::<CommitSet>()`

claytonrcarter:
> Oops, I wrote CommitSet::from_iter(foo.collect()) but it should be just foo.collect(). Anyways, you can leave this as is, and I'll update this site and the other uses in the codebase to use .collect() later.

Is this what you had in mind over at https://github.com/arxanas/git-branchless/pull/366#discussion_r858156321?

Out of curiosity, do you generally prefer `let ... : CommitSet = ... .collect()` or `let ... = ... collect::<CommitSet>()`? Or does it depend on the context? My understanding ATM is that they're roughly equivalent ways of nudging the compiler towards the desired type, right?

### DIFF
--- a/git-branchless-lib/src/core/dag.rs
+++ b/git-branchless-lib/src/core/dag.rs
@@ -139,43 +139,29 @@ impl Dag {
             branch_oid_to_names,
         } = references_snapshot;
 
-        let obsolete_commits = CommitSet::from_iter(
-            observed_commits
-                .iter()
-                .copied()
-                .filter(|commit_oid| {
-                    match event_replayer
-                        .get_cursor_commit_activity_status(event_cursor, *commit_oid)
-                    {
-                        CommitActivityStatus::Active | CommitActivityStatus::Inactive => false,
-                        CommitActivityStatus::Obsolete => true,
-                    }
-                })
-                .map(CommitVertex::from)
-                .map(Ok)
-                .collect_vec(),
-        );
+        let obsolete_commits: CommitSet = observed_commits
+            .iter()
+            .copied()
+            .filter(|commit_oid| {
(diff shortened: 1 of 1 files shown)


arxanas:
Thanks for cleaning this up!

> My understanding ATM is that they're roughly equivalent ways of nudging the compiler towards the desired type, right?

Yep, they're pretty much the same. Generally, I prefer `let x: CommitSet = (...).collect();` just because there's less line noise (as `::<>` can be a bit hard to read). I would say that Rust is actually missing the relevant feature, which is the ability to annotate the type of an arbitrary expression, and not just a pattern binding. Using the same `:` syntax, it might hypothetically look like this:

```
let x = foo
    .iter()
    .collect(): CommitSet;
```

I would probably prefer to do it that way if the feature existed.

martinvonz:
> I would say that Rust is actually missing the relevant feature, which is the ability to annotate the type of an arbitrary expression, and not just a pattern binding. Using the same `:` syntax, it might hypothetically look like this:
> 
> ```
> let x = foo
>     .iter()
>     .collect(): CommitSet;
> ```
> 
> I would probably prefer to do it that way if the feature existed.

It's might be less hypothetical than you think: https://github.com/rust-lang/rust/issues/23416. At least I think that feature will allow you to write what you wanted.

arxanas:
@martinvonz I knew the feature was in in RFC (after all, OCaml has it), but given that it's been 7+ years, I don't have any hope for it coming out anytime soon 😅.

martinvonz:
> @martinvonz I knew the feature was in in RFC (after all, OCaml has it), but given that it's been 7+ years, I don't have any hope for it coming out anytime soon 😅.

Wow, I didn't realize it had been there for that long (or even very long at all)

PR #382, opened 2022-04-29
Working copy snapshotting

arxanas:


git-branchless-lib/src/core/check_out.rs
```diff
@@ -9,6 +9,7 @@ use cursive::utils::markup::StyledString;
 use tracing::instrument;
 
 use crate::git::{CategorizedReferenceName, GitRunInfo, MaybeZeroOid, Repo, WorkingCopySnapshot};
+use crate::util::ExitCode;
```
epage, line 16: Check out [proc_exit](https://docs.rs/proc-exit/).  I ran into a similar situation and created it to help with cases like this

epage:
Not seeing a description of this.  Is this the branchless equivalent of [git branch-stash](https://github.com/epage/git-branch-stash) (backs up and restores what your branches point to)?

arxanas:
@epage `git-branchless` has already had a kind of equivalent to `git-branch-stash` for a while, as `git undo` will undo branch moves. Technically speaking, it stores branch-move *diffs* instead of *snapshots*, which I think is fair to say has turned out to be an inferior approach, but it mostly works in practice.

This stack (and #357, which it continues) is to implicitly make commits out of the working copy contents, jj-style (although these commits don't update any branches, and so are largely invisible to the user).

Motivation:

- Currently, commands like `git amend` will modify state in a non-undoable way. You can amend the current working copy changes into the current commit, but there's no easy way to get them out again. By making a working copy snapshot before carrying out the amend, it becomes possible to rewind to the state immediately before the amend.
- This will simplify the implementations of future commands. For example, then we can just have a single implementation shared for a command that selects only certain changes for a commit, and a command which splits a commit into multiple commits. jj and hg already represent things internally this way, if i understand correctly.

Some other notes:

- Unlike jj, we currently only snapshot the working copy when we issue a command that would query the working copy status@ anyways. That way, we don't incur unnecessary overhead on every command, which is a problem in practice for large repositories. (Ideally both git-branchless and jj would defer to a solution like Watchman to query the working copy status efficiently.) 
- We only snapshot tracked files, to avoid committing secrets or large binary blobs.
- We don't store snapshots anywhere where they could be accidentally pushed, to ensure that we don't leak secrets. At present, they'll be garbage-collected by Git eventually.

martinvonz:
> * jj and hg already represent things internally this way, if i understand correctly.

jj does, yes, but hg mostly does not. There are abstractions that let you treat the working copy like a commit, but there's no snapshotting, so each command still need to check if the working copy is clean, or create its own temporary commit.

> * (Ideally both git-branchless and jj would defer to a solution like Watchman to query the working copy status efficiently.)

Yes. For jj (and maybe git-branchless too?), we'll probably want to subscribe to changes and automatically commit the working copy. That way we can skip committing the working copy from the commands themselves (as if the user always runs `jj --no-commit-working-copy`).

git-branchless-lib/src/core/check_out.rs
arxanas: @epage `proc_exit::Code` isn't `#[must_use]` ☹️

PR #388, opened 2022-05-05
feat(amend): support undoing working copy changes from before the amend

arxanas:


arxanas:
Oops, already merged this.

PR #390, opened 2022-05-05
ci: don't override nixpkgs version (use the lock file's version)

martinvonz:
This should avoid surprises like martinvonz/jj#299, I hope.


arxanas:
Thanks! (This thanks supersedes https://github.com/martinvonz/jj/pull/300#pullrequestreview-963750363)

Issue #371, opened 2022-04-24
reword invokes vi instead of nano


AaronLieberman:
### Description of the bug

When I run git reword, `vi` comes up. When I run other git commands that invoke an editor (e.g. git amend), nano comes up. I didn't have core.editor set on this machine, when I set it to nano, nano shows up as expected. 

### Expected behavior

- no core.editor set
- `git amend` nano shows up
- `git reword 44087e1` nano shows up

### Actual behavior

- no core.editor set
- `git amend` nano shows up
- `git reword 44087e1` vi shows up

### Version of `git-branchless`

git-branchless 0.3.12

### Version of `git`

git version 2.25.1

### Version of `rustc`

rustc 1.57.0

#### Software version

git-branchless 0.3.12

#### Operating system

Linux 5.10.16.3-microsoft-standard-WSL2

#### Command-line

```bash
/home/aaron/.cargo/bin/git-branchless bug-report 
```

#### Environment variables

```bash
SHELL=/bin/bash
EDITOR=<not set>
```

#### Git version

```
> git version 
git version 2.25.1
```

#### Events

<details>
<summary>Show 5 events</summary>

##### Event ID: 11, transaction ID: 24

1. `CommitEvent { timestamp: 1650765291.0, event_tx_id: EventTransactionId(24), commit_oid: NonZeroOid(8cc55d6085aaeceef2a1e5ac4c87db6a51e6fdb9) }`
1. `RewriteEvent { timestamp: 1650765292.4729295, event_tx_id: EventTransactionId(24), old_commit_oid: f148352480d981f078b609f5a49870baa2e6d107, new_commit_oid: 8cc55d6085aaeceef2a1e5ac4c87db6a51e6fdb9 }`
```
:
O 79d815c 13d xxx xxxxxxx
|\
| @ 8cc55d6 6m xxx xx xxxxxxx xxxxxx
|
X f148352 13d (rewritten as 8cc55d60) (remote origin/main) xxx xx xxxxxxx xxxxxx
|\
| o 6f6a4ae 17m xxx
|
o 44087e1 17m xxxx
```
##### Event ID: 9, transaction ID: 21

1. `UnobsoleteEvent { timestamp: 1650765273.3941846, event_tx_id: EventTransactionId(21), commit_oid: NonZeroOid(6f6a4aebcb2212ade943d71e0f2b014bcfe8bbb8) }`
[...]

PR #397, opened 2022-05-10
feat(reword): Omit marker line when editting exactly 1 commit

claytonrcarter:
From https://github.com/arxanas/git-branchless/issues/371#issuecomment-1121540003:

> That being said, @claytonrcarter, maybe it would make sense to remove the header if only one commit is being reworded? In addition to the user surprise here, I also personally have the problem that I want to immediately start typing on the focused line, but then I accidentally overwrite the reword header.

Thanks for the nudge. It had been bugging me, too! I believe that this change handles this case, though I haven't hammered on it. It's passed the "fiddle around with it" test for the past hour, though. 😄 

cc @AaronLieberman

### DIFF
--- a/git-branchless/src/commands/reword.rs
+++ b/git-branchless/src/commands/reword.rs
@@ -404,7 +404,12 @@ fn prepare_messages(
             original_message
         };
 
-        message.push_str(format!("++ reword {}\n{}\n\n", oid, msg).as_str());
+        let msg = if commits.len() == 1 {
+            format!("{}\n\n", msg)
+        } else {
+            format!("++ reword {}\n{}\n\n", oid, msg)
+        };
+        message.push_str(msg.as_str());
     }
 
     message.push_str(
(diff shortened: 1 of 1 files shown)


git-branchless/src/commands/reword.rs
```diff
@@ -522,6 +527,14 @@ fn parse_bulk_edit_message(
         commits_oids.insert(commit.get_short_oid()?, commit.get_oid());
     }
 
+    let message = if commits.len() == 1 && !message.contains("++ reword") {
+        // For single commits, add the marker line, but only if the user hasn't already done so.
```
claytonrcarter, line 531: If the user has added their own marker line (however unlikely that may be), adding another one will generate a less-direct error message. So this check for `contains` is pretty edge-casey, but will end up producing a nicer error message if it ever does happen.

git-branchless/src/commands/reword.rs
```diff
@@ -522,6 +527,14 @@ fn parse_bulk_edit_message(
         commits_oids.insert(commit.get_short_oid()?, commit.get_oid());
     }
 
+    let message = if commits.len() == 1 && !message.contains("++ reword") {
```
arxanas, line 530: To avoid the `.unwrap` here, we can write it something like this:

```
let message = match commits.as_slice() {
    [only_commit] if !message.contains("++ reword") => format!(...),
    _ => message,
};
```

arxanas (approved):

arxanas:
Typo in commit message: `editting` -> `editing`

git-branchless/src/commands/reword.rs
claytonrcarter: Oh ... that's yummy! Thank you! I've made this change. `match` is really exciting to learn more about. It's been eye opening. 👀

claytonrcarter:
> Typo in commit message: editting -> editing

🤦 I just can't stop making that typo. Fixed! Thanks.

PR #398, opened 2022-05-10
Implement revsets

### LINKED ISSUE #175: Implement support for revsets
See discussion at https://github.com/arxanas/git-branchless/discussions/157.

arxanas:
Closes #175.


claytonrcarter:
This is really, really great, and much more robust than what I had worked up for parsing ranges, which was based on https://github.com/rust-lang/git2-rs/blob/master/examples/rev-list.rs

I know this is still draft/preliminary (so take these with a grain of salt) but a couple of things that popped up while I was playing with this:
- UX: many of the command args have been updated from `commits` or `hashes` to `revsets`. This then shows up in the output of `git command -h` as eg 
```
USAGE:
    git-branchless hide [OPTIONS] [REVSETS]...

ARGS:
    <REVSETS>...    Zero or more commits to hide
```
Though I know what revsets are (er, at least, I think I do!), I wonder if this might be potentially confusing to new users. I wonder if it might be preferrable to leave this option names as `COMMITS` and supplement the description to include `Zero or more commits to hide. (Also accepts revsets.)`
- Parsing of `..` ranges appears to differ from normal git `..` ranges:
```
❯ git rev-list master..@
... prints 6 commits, including ...
527f3e31f9cbe84d372ebce7aa514f3c875d9d2c

❯ cargo run -- query master..@
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query 'master..@'`
527f3e31f9cbe84d372ebce7aa514f3c875d9d2c
(only prints this 1 commit)

❯ git rev-list master..527f3e3
... prints 6 commits ...

❯ cargo run -- query master..527f3e3
    Finished dev [unoptimized] target(s) in 0.15s
     Running `target/debug/git-branchless query master..527f3e3`
Evaluation error for expression 'master..527f3e3': name is not defined: 'master..527f3e3'

❯ git rev-list master..
... prints 6 commits ...

❯ cargo run -- query master..
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query master..`
Evaluation error for expression 'master..': name is not defined: 'master..'
```
- parent selectors are a little off: `@^`, `@^^`, etc seem to work fine, but `@^1`, etc does not. `~` does not work at all. For a first pass, I think it would be nice to at least have `~` refer to parents like `^`, and then -- if desired -- these can be solidified to better match `git` at a later time.
- When a commit can't be found, the error message might be a little difficult for new users to understand: 
```
❯ cargo run -- query abc123
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query foobarbaz`
Evaluation error for expression 'abc123': name is not defined: 'abc123'
```
I wonder if something like "Error evaluating 'abc123'. No matching commits or branches found, and 'abc123' is not a valid revset.` or something along those lines. 

I don't mean these to be nitpicky; they just came up while I was playing with it. I think this is really exciting!

claytonrcarter:
This is really, really great, and much more robust than what I had worked up for parsing ranges, which was based on https://github.com/rust-lang/git2-rs/blob/master/examples/rev-list.rs

I know this is still draft/preliminary (so take these with a grain of salt) but a couple of things that popped up while I was playing with this:
- UX: many of the command args have been updated from `commits` or `hashes` to `revsets`. This then shows up in the output of `git command -h` as eg 
```
USAGE:
    git-branchless hide [OPTIONS] [REVSETS]...

ARGS:
    <REVSETS>...    Zero or more commits to hide
```
Though I know what revsets are (er, at least, I think I do!), I wonder if this might be potentially confusing to new users. I wonder if it might be preferrable to leave this option names as `COMMITS` and supplement the description to include `Zero or more commits to hide. (Also accepts revsets.)`
- Parsing of `..` ranges appears to differ from normal git `..` ranges:
```
❯ git rev-list master..@
... prints 6 commits, including ...
527f3e31f9cbe84d372ebce7aa514f3c875d9d2c

❯ cargo run -- query master..@
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query 'master..@'`
527f3e31f9cbe84d372ebce7aa514f3c875d9d2c
(only prints this 1 commit)

❯ git rev-list master..527f3e3
... prints 6 commits ...

❯ cargo run -- query master..527f3e3
    Finished dev [unoptimized] target(s) in 0.15s
     Running `target/debug/git-branchless query master..527f3e3`
Evaluation error for expression 'master..527f3e3': name is not defined: 'master..527f3e3'

❯ git rev-list master..
... prints 6 commits ...

❯ cargo run -- query master..
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query master..`
Evaluation error for expression 'master..': name is not defined: 'master..'
```
- parent selectors are a little off: `@^`, `@^^`, etc seem to work fine, but `@^1`, etc does not. `~` does not work at all. For a first pass, I think it would be nice to at least have `~` refer to parents like `^`, and then -- if desired -- these can be solidified to better match `git` at a later time.
- When a commit can't be found, the error message might be a little difficult for new users to understand: 
```
❯ cargo run -- query abc123
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/git-branchless query foobarbaz`
Evaluation error for expression 'abc123': name is not defined: 'abc123'
```
I wonder if something like "Error evaluating 'abc123'. No matching commits or branches found, and 'abc123' is not a valid revset.` or something along those lines. 

I don't mean these to be nitpicky; they just came up while I was playing with it. I think this is really exciting!

arxanas:
@epage given a type `Revset` which appears as a positional argument, is there a way to surface contextual information for each parameter of that type which appears in a command? Something that would let me explain what a "revset" is without copying the same text everywhere. I was looking through the clap API, but I didn't see anything obviously appropriate.

> Parsing of .. ranges appears to differ from normal git .. ranges:

For this syntax, you would use `foo:bar` instead of `foo..bar`, which (kind of) comes from the Mercurial syntax. (It's a little different due to complications in the grammar.)

> parent selectors are a little off: @^, @^^, etc seem to work fine, but @^1, etc does not. ~ does not work at all. For a first pass, I think it would be nice to at least have ~ refer to parents like ^, and then -- if desired -- these can be solidified to better match git at a later time.

Sounds reasonable. I'll try to get that working before merging. The `lalrpop` grammar I copied doesn't seem to have an easy way to handle operators which might be either infix or pre/postfix. @quark-zju do you have any idea if that's possible with `lalrpop`?

We might also want to decide if we want to copy the Jujutsu approach of using `+` and `-` for these instead.

> When a commit can't be found, the error message might be a little difficult for new users to understand:

I'll see what I can do 👍

git-branchless/src/revset/grammar.lalrpop
```diff
@@ -106,5 +106,5 @@ Symbol2: String = {
 }
 
 Symbol1: String = {
-    r"[a-zA-Z0-9/_$@.]+" => <>.to_string(),
+    r"[a-zA-Z0-9/_$@.-]+" => <>.to_string(),
```
martinvonz, line 109: In what contexts will `-` be interpreted as an operator (for set difference) after this change?

git-branchless/src/revset/grammar.rs
```diff
@@ -0,0 +1,2625 @@
+// auto-generated: "lalrpop 0.19.8"
```
martinvonz, line 1: Did you try generating this at build time instead? Was it just too slow? For incremental builds, you can do something like https://github.com/martinvonz/jj/blob/6483aeefeaeb22462901ba0201b9b3a2f99052f4/lib/build.rs#L40-L43

martinvonz:
> The `lalrpop` grammar I copied doesn't seem to have an easy way to handle operators which might be either infix or pre/postfix. @quark-zju do you have any idea if that's possible with `lalrpop`?

I think that's the issue I ran into with lalrpop, which made me switch to pest. But you may want to wait for a more reliable answer from Jun before you make a decision :) Even *if* lalrpop doesn't let you have the same operator as infix and postfix, you can of course decide that it's worth it for the better ergonomics lalrpop provides (it does look much simpler than [what I've done](https://github.com/martinvonz/jj/blob/6483aeefeaeb22462901ba0201b9b3a2f99052f4/lib/src/revset.rs#L424-L563))

arxanas:
> Sounds reasonable. I'll try to get that working before merging. The lalrpop grammar I copied doesn't seem to have an easy way to handle operators which might be either infix or pre/postfix. @quark-zju do you have any idea if that's possible with lalrpop?

The fix I implemented was to first attempt to parse the expression as a regular Git revision expression, which would handle things like `~3`, and only then fall back to the revset language.

FYI I was experimenting and I did find a way to handle operators which are "missing"-fix on one side. You do need to resolve the ambiguity in an expression like `1:2:3`. This can be done the same way as in the original LALRPOP grammar, by defining precedence levels and not allowing a `:`-expression as an operand to a `:`.

git-branchless/src/revset/grammar.lalrpop
arxanas: It looks like it's only interpreted as an operator when there is a space to disambiguate, which is what I was hoping for.

git-branchless/src/revset/grammar.rs
arxanas: I got rid of the `build.rs` lately and didn't want to add it back so that I could keep build times down... but I might end up just adding it anyways, since it's a pain to manually regenerate the grammar each time.

PR #399, opened 2022-05-11
feat(hide): Display (and optionally delete) abandoned branches

claytonrcarter:
Changes in this PR:
- Display a list of branches abandoned during a `git hide` operation
- Add `-D/--delete-branches` flag to delete branches abandoned during a `git hide` operation
  - The `-D` short flag was chosen for parity w/ `git branch -D <branch name>`

See discussion at https://github.com/arxanas/git-branchless/discussions/283#discussioncomment-2298871

### DIFF
--- a/git-branchless/src/commands/hide.rs
+++ b/git-branchless/src/commands/hide.rs
@@ -14,11 +14,18 @@ use lib::core::effects::Effects;
 use lib::core::eventlog::{CommitActivityStatus, Event};
 use lib::core::eventlog::{EventLogDb, EventReplayer};
 use lib::core::formatting::{printable_styled_string, Glyphs, Pluralize};
-use lib::git::{CategorizedReferenceName, Repo};
+use lib::core::rewrite::move_branches;
+use lib::git::{CategorizedReferenceName, GitRunInfo, MaybeZeroOid, Repo};
 
 /// Hide the hashes provided on the command-line.
 #[instrument]
-pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Result<ExitCode> {
+pub fn hide(
+    effects: &Effects,
+    git_run_info: &GitRunInfo,
+    hashes: Vec<String>,
+    delete_branches: bool,
+    recursive: bool,
+) -> eyre::Result<ExitCode> {
     let now = SystemTime::now();
     let glyphs = Glyphs::detect();
     let repo = Repo::from_current_dir()?;
@@ -89,6 +96,22 @@ pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Re
         }
     }
 
+    if delete_branches {
+        // Delete any branches pointing to any of the hidden commits by "moving" them from their
+        // current OID to a Zero OID.
+        let abandoned_branches = commits
+            .iter()
+            .map(|commit| (commit.get_oid(), MaybeZeroOid::Zero))
+            .collect();
+        move_branches(
+            effects,
+            git_run_info,
+            &repo,
+            event_tx_id,
+            &abandoned_branches,
+        )?;
+    }
+
     let mut abandoned_branches: Vec<String> = commits
         .iter()
         .filter_map(|commit| {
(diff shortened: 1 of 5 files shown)


git-branchless/src/commands/hide.rs
```diff
@@ -89,14 +113,62 @@ pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Re
         }
     }
 
+    if delete_branches {
+        // Delete the abandonded branches by "moving" them from their current OID to a Zero OID.
+        let abandoned_branches = abandoned_branches
+            .values()
+            .map(|oid| (*oid, MaybeZeroOid::Zero))
+            .collect();
```
arxanas, line 121: I think it should be safe to pass OIDs which don't point to a branch to `move_branches`. Then we can reduce the scope of `abandoned_branches` by moving its construction below and writing something like `commits.iter().map(|commit| (commit.get_oid(), MaybeZeroOid::Zero)).collect()` here.

git-branchless/src/commands/hide.rs
```diff
@@ -89,14 +113,62 @@ pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Re
...
+        move_branches(
+            effects,
+            git_run_info,
+            &repo,
+            event_tx_id,
+            &abandoned_branches,
+        )?;
+    }
+
+    let abandoned_branches: Vec<&str> = abandoned_branches.into_keys().collect();
```
arxanas, line 131: As per above, move the construction of branches from above to down here, maybe something like this:

```
commits
    .iter()
    .flat_map(|commit| branch_oid_to_names.get(commit.get_oid()).unwrap_or_default())
    .collect();
```

You should also sort the branch names before rendering for determinism.

To render the branch name, do something like this rather than manually stripping the suffix above:

```
CategorizedReferenceName::new(branch_name).render_suffix()
```

git-branchless/src/opts.rs
```diff
@@ -157,6 +157,10 @@ pub enum Command {
         /// Can either be hashes, like `abc123`, or ref-specs, like `HEAD^`.
         commits: Vec<String>,
 
+        /// Also delete any branches that are abandoned as a result of this hide.
+        #[clap(short = 'D', long = "delete-branches")]
+        delete_branches: bool,
```
arxanas, line 162: [non-blocking] I had personally imagined `-b`/`--branches` as the option, but I don't feel strongly about it.

git-branchless/tests/command/test_hide.rs
```diff
@@ -168,7 +168,14 @@ fn test_branches_always_visible() -> eyre::Result<()> {
     git.run(&["branch", "test"])?;
     git.run(&["checkout", "master"])?;
 
-    git.run(&["hide", "test", "test^"])?;
+    let (stdout, _stderr) = git.run(&["hide", "test", "test^"])?;
```
arxanas, line 171: Let's add a test case which hides more than one branch. Otherwise we don't get to show off all of your string-rendering logic 😉

arxanas (approved):

git-branchless/src/commands/hide.rs
```diff
@@ -89,14 +113,62 @@ pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Re
...
+    if !abandoned_branches.is_empty() {
+        // This message will look like either of these:
+        // Abandoned X branches: <list of branches>
+        // Abandoned and deleted X branches: <list of branches>
```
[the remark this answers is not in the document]
claytonrcarter: Good call! I've changed it.

git-branchless/src/commands/hide.rs
```diff
@@ -89,14 +113,62 @@ pub fn hide(effects: &Effects, hashes: Vec<String>, recursive: bool) -> eyre::Re
         }
     }
 
+    if delete_branches {
+        // Delete the abandonded branches by "moving" them from their current OID to a Zero OID.
```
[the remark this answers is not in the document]
claytonrcarter: Fixed, though I ended up rewriting the whole comment.

git-branchless/src/commands/hide.rs
claytonrcarter: Yup, this is great, thank you! It really tightened up the changes and cleaned things up nicely.

git-branchless/tests/command/test_hide.rs
claytonrcarter: 😄 Yeah, that makes sense. I added a test for hiding/deleting 2 branches, and tried to set it up confirm the sorting of the branch names, too.

git-branchless/src/opts.rs
claytonrcarter: Well, if you really don't feel strongly about it, then I would prefer it as is, but I haven't thought too much about it past the symmetry with `git branch -D`. Ultimately, it's up to you and I'll be happy to change it if you do feel strongly about it!

arxanas: Nope!

PR #400, opened 2022-05-11
feat(move): Add `--insert` flag 

claytonrcarter:
This adds an `-i/--insert` flag to `git move` to attempt to insert the moved commits between the destination and it's children. 

- This is still draft b/c I still need to confirm a few things, and I don't like my error messages.
- This only works if the the moved subtree is a "stick", ie the subtree root commit (ie source commit) has exactly 1 parent and the subtree has exactly 1 head.
  - I have a suspicion that the "exactly 1 parent" restriction is technically unnecessary, but it made the code easier to write and reason about.
- This also updates the rebase planner to support moving commit ranges instead of just entire subtrees, to handle the case where the moved commits are already descendants of the destination. This works, but I see that there has been some recent work in https://github.com/arxanas/git-branchless/tree/move-individual-commits that might be a more elegant/less-hacky solution. I'm happy to take guidance on how to improve this!


arxanas:
Haven't looked at this PR at all yet, but I recommend using `-I` instead of `-i` for the short form. For example, we might want an equivalent to `git commit -i` to interactively select changes and `git commit -I` to insert the commit into the subgraph.

claytonrcarter:
> I recommend using -I instead of -i for the short form

Sure thing. Done.

claytonrcarter:
> I see that there has been some recent work in https://github.com/arxanas/git-branchless/tree/move-individual-commits 

So, I went ahead and updated/rebased this work onto that existing work just for the sake of exploration. I went ahead and implemented `Constraint::ImmediateChild` as well as `plan::move_commit`, but I have not wired up an `--exact` flag.

claytonrcarter:
FYI I just pushed a "wip" commit that is my first attempt at a struct/impl. *If* you like the general initial approach of treating the `constraint_graph` as a `HashMap<NonZeroOid, HashMap<NonZeroOid, Option<NonZeroOid>>>`, then perhaps this refactor may provide a more ergonomic way to manipulate those constraints instead of directly, repeated querying the keys/values of the Map.

Emphasis on WIP. 😄 😴

claytonrcarter:
I've chopped up the [wip changes I mentioned](https://github.com/arxanas/git-branchless/pull/400#issuecomment-1124505069) and pushed them back up into the commit graph so the changes flow a bit better. I know this isn't done yet, but I think it's ready for an initial review to point out all of the things that are hiding right in front of me. 😄  I'll update the PR description when I get some more time in the next few days, but the summary of the changes are:
1. pick up the changes from @arxanas `move-individual-commits` branch
2. extend those forward and extract a `ConstraintGraph` struct in the rebase planner. (This is the bulk of the changes, but it wasn't rocket science; it was just a straightforward refactoring to hide some code behind a more expressive (maybe? I think?) API.)
3. updating the planner to support moving ranges of commits
4. connecting a few wires to implement `--insert`

arxanas (commented):
@claytonrcarter Nice work so far.

I envisioned there being only two kinds of constraints:

- `ImmediateChild`
- `ChildAndDescendants`

The third kind which you have, `ChildAndSomeDescendants`, can be expressed in terms of the other two.

In general, when a commit X is moved, what should happen to its child commit Y? There are two possible answers:

- The child commit Y should still remain the child of X, and therefore should be moved along with it.
- The child commit Y should be moved to be a child of the nearest unmoved ancestor commit of X.

Any unconstrained commits will retain their usual child-parent relationships.

Suppose we have this graph:

```mermaid
graph RL
  E --> D
  D --> C
  C --> B
  B --> A
```

If we create the constraint `ChildAndDescendants` to move `C` on top of `A` (which is the only kind of constraint which we have now) and execute the rebase, we should get this graph:

```mermaid
graph RL
  E --> D
  D --> C
  C -.-> A
  B --> A
```

But if we create a constraint `ImmediateChild` to move `C` on top of `A`, we should resolve `C`'s child commit `D` by moving it on top of the nearest unconstrained parent (and `E`'s parent will remain `D`, unchanged):

```mermaid
graph RL
  E --> D
  D -.-> B
  C -.-> A
  B --> A
```

To move the *range* of commits including only `C` and `D`, we'll issue *two* contraints: a `ChildAndDescendants` from `C` to `A` and a `ChildAndDescendants` from `E` to `B`:


```mermaid
graph RL
  E -.-> B
  D --> C
  C -.-> A
  B --> A
```

You can also implement insertion this way. To undo the above move, you can issue constraints `ChildAndDescendants` from `C` to `B` and `ChildAndDescendants` from `E` to `D`.

This should also handle the case of moving non-stick ranges (although we should probably check that moved commits have the same number of parents, if we don't already, i.e. a non-merge commit should not become a merge commit because it was rebased on top of a subtree with two heads).

claytonrcarter:
Thank you for this detailed guidance; that makes a ton of sense! I've played with it a bit and I was able to remove `ChildAndSomeDescendants` w/o issue. I have some more error handling to work on before I want to push, though. 🙈 

While removing `ChildAndSomeDescendants`, I started wondering if we even need `ImmediateChild`. I am assuming that I'm just not seeing something that you have in mind for it, but it strikes me that we could treat it like a range with only 1 commit. That is, in your example of moving C on top of A w/ `ImmediateChild`, couldn't that also be accomplished with a `ChildAndDescendants` from C to A and a `ChildAndDescendants` from D to B?

arxanas:
> While removing ChildAndSomeDescendants, I started wondering if we even need ImmediateChild. I am assuming that I'm just not seeing something that you have in mind for it, but it strikes me that we could treat it like a range with only 1 commit. That is, in your example of moving C on top of A w/ ImmediateChild, couldn't that also be accomplished with a ChildAndDescendants from C to A and a ChildAndDescendants from D to B?

I think you're right, it could be all implemented with one operation. The main concerns in my eyes are:

- When we query the DAG. Currently, we don't query it at all when building the initial set of constraints. If we wanted to issue two constraints to move a single commit, we'd have to query for the children commits at the time that we insert the constraint into the set.
- Error-reporting. I forget what we do currently. Is it an error to issue another constraint for a commit which is already constrained? If we decide we want to move a child commit somewhere else in the graph, we have to be sure that the new constraint overrides the old one properly.

Feel free to proceed however you think is cleaner, whether it's having either one or multiple types of constraint.

By the way, you might be interested in contributing the same feature either instead or in addition to [Jujutsu](https://github.com/martinvonz/jj). In the long term, I think it will make most sense to concentrate efforts there.

claytonrcarter:
I'll try to pick this up again soon (sorry, I've been distracted). And thanks for clarifying about querying the DAG while building the initial constraints vs building the actual plan. I hadn't been thinking about the performance penalty of DAG queries since most of my repos are relatively small. I'll keep tinkering and see what I come up with.

> By the way, you might be interested in contributing the same feature either instead or in addition to [Jujutsu](https://github.com/martinvonz/jj). In the long term, I think it will make most sense to concentrate efforts there.

I'm looking forward to trying out `jj` sometime, but for the time being I'm still really liking the way that `branchless` just augments my existing workflow. Once I land this PR here, maybe that's a great excuse to finally dig into jj and learn more about it, though!

claytonrcarter:
This now includes my latest WIP. The main API change is removing `Constraint::ImmediateChild` and adding `Constraint::ChildrenOfToParentOf`.[^1] (Terrible name ... any suggestions?) Doing so allows moving single commits and ranges w/o doing any dag queries while collecting the initial constraints. There *are* some dag queries needed in `r#move` to figure some things out, but I need to think about those some more before finishing this. I also haven't addressed your question about error reporting yet.

[^1]: I also expanded the scope of the new `ConstraintGraph` struct a bit, moving many `fn`s that were only related to populating the constraint graph from `RebasePlanBuilder` into `ConstraintGraph`.

claytonrcarter:
OK, I think I'm ready for some more guidance on this. I have some questions based on your comments:

> When we query the DAG. Currently, we don't query it at all when  building the initial set of constraints. If we wanted to issue two  constraints to move a single commit, we'd have to query for the children commits at the time that we insert the constraint into the set.

I interpreted this as meaning that we should _avoid DAG queries while collecting the initial constraints_. Is that what you intended? If so, *how strongly* should be we be avoiding that? I ask b/c I'm finding it very difficult to handle some cases. I think that doing some DAG queries while collecting these constraints might be helpful, but I don't want to dig into that if it's a blocker related to performance or something.

> Error-reporting. I forget what we do currently. Is it an error to  issue another constraint for a commit which is already constrained? If  we decide we want to move a child commit somewhere else in the graph, we have to be sure that the new constraint overrides the old one properly.

Currently, no, it is not an error to "reconstrain" a commit. As I understand it, in fact, doing so will just duplicate the constrained subtree. I have code locally that handles this better, and I'm just working on the details of some specific cases (like the aforementioned adjacent commits).



Here's an example that I'm struggling with. Consider this graph:

```mermaid
flowchart RL
    E --> D --> C --> B --> A
```

Now consider these rebase instructions:

```rust
builder.move_commit(C, A);
builder.move_commit(D, A);
```

I would expect this to produce the following rebased graph, which just moves the exact commits `C` and `D` out of the `B-C-D-E` lineage:
```mermaid
flowchart RL
    E --> B --> A
    D --> A
    C --> A
```

The problem I'm facing is that, w/o doing any DAG queries initially, it's very difficult (impossible?) to know whether `E` should end up on `B` or `C`, because of how the first `move_commit` affects the second: 

- the first call says "move `C` to `A`" *and* "move `D` to `B`" 
  - As you know, moving `D` also implies it's descendent `E`
- the second then says "move `D` to `A`" *and* "move `E` to `C`"
  - However, as `C` is already constrained, I've opted to keep looking at it's ancestors until we find an unconstrained commit and then using that.

The issue I'm facing is that I would expect those 2 `move_commit` calls to be order-independent (so we could do them in any order), but because of the need to handle child commits, doing these in the reverse order isn't working:

 ```rust
 builder.move_commit(D, A);
 builder.move_commit(C, A);
 ```

- now the first says "move `D` to `A`" *and* "move `E` to `C`"
- the second call says "move `C` to `A`" *and* "move `D` to `B`" 
  - Note how the first order (where move `C` then `D`) doesn't constrain `E` until the second call. In this order, though, `E` *is* constrained first, and the second call **doesn't/can't know that it should cancel the constraint**.

I *think* that doing some more interrogation of the DAG and existing constraints while the initial constraints are being collected could help with this, but I want a sense of how careful to be w/ DAG queries before proceeding. Thank you!



Phew! This has proven to be much more complicated than I expected. The basic cases (ie handling the needs of `move --insert`) seem to be easy, but generalizing things to be able to handle possibly arbitrary range constraints is proving to be tricky. I guess that leads me to another question: **is it even worth it to support arbitrary range constraints at this point?** If we limited the algorithm to only accept 1 range constraint, it would delay much of the aforementioned complexity until we really need/want it.

What do you think?

arxanas:
> I interpreted this as meaning that we should avoid DAG queries while collecting the initial constraints. Is that what you intended? If so, how strongly should be we be avoiding that? I ask b/c I'm finding it very difficult to handle some cases. I think that doing some DAG queries while collecting these constraints might be helpful, but I don't want to dig into that if it's a blocker related to performance or something.

This isn't a blocker for performance. After all, we'd have to do the same DAG queries immediately afterward when we want to execute the rebase plan anyways. It's just a stylistic matter, so we don't need to restrict ourselves here if it makes something more difficult to implement, but see below.

> The problem I'm facing is that, w/o doing any DAG queries initially, it's very difficult (impossible?) to know whether E should end up on B or C, because of how the first move_commit affects the second

The problem is that you're trying to do all the work in one pass, but since later constraints can affect earlier constraints, this won't work. Really, you want to break this up into two steps:

- Pass 1, apply direct constraints:
  - Set D's parent to A.
  - Set C's parent to A.
- Pass 2, apply implied constraints:
  - Find D's immediate unconstrained children, then find D's nearest unconstrained ancestor. Its immediate parent C is already constrained (by pass 1), so we skip it and go to B, which is unconstrained. So we'll set E's parent to B.
  - Find C's immediate unconstrained children. D is its only child and it's constrained (by pass 1), so we have no more work to do.

Since we're doing a two-step constraint resolution process, I think it also works out that we don't need to do any DAG queries when issuing the initial constraints with `move_*`. We have to do the second pass later anyways to make sure we're constraining child nodes correctly.

> I guess that leads me to another question: is it even worth it to support arbitrary range constraints at this point? If we limited the algorithm to only accept 1 range constraint, it would delay much of the aforementioned complexity until we really need/want it.

It looks like you the single-range case working, so I'd be happy merging this as-is. There would be immediate use-cases, such as `git move`ing ranges of commits (obviously) but also for a `git commit --insert` command and possibly also a `git split` command.

Thanks for all your work so far! 😊

git-branchless-lib/src/core/rewrite/plan.rs
```diff
@@ -1056,4 +1156,594 @@ mod tests {
...
+        })?;
+
+        let (stdout, _stderr) = git.run(&["smartlog"])?;
+
+        // FIXME This output is correct mathematically, but feels like it should
+        // be incorrect. What *should* we be doing if the user moves 2 ranges w/
+        // the same source/root oid but different end oids??
+        //
+        // NOTE See also the next test for the other case: where the end of the
+        // range is moved again. That *does* feel correct.
```
claytonrcarter, line 1548: I want to bring you attention to this case/test. I understand why it's correct, but it feels a little surprising. ATM there's no way (that I know) to handle it any differently.

claytonrcarter:
Thank *you* for your guidance on this! These comments were very helpful. And you're correct, handling the constraints in 2 passes means we still don't need any DAG queries while collecting the initial constraints. Also, I was able to get everything working w/o limiting us to moving a single-range, so no worries there either.

The only piece that I needed to add was that we need to handle the "ancestor search" for constrained ranges (and commits) differently than constrained subtrees. (I'm open to other ways to handle this, of course, but this is what I've come up with.) Basically, in the above example, moving the 2 commits `C` and `D` will produce different results than if we were to move the subtree @ `C` and then commit `D` (in either order). In the former, we want `E` to end up @ `B` (as covered in previous comments), but in the latter we want it to end up @ `C`, even though `C` is also being moved. I *think* that I've covered the cases in all of the tests that I wrote, but I would not be surprised if you find some that I didn't cover.

Speaking of tests, I added a bunch of tests for the rebase planner. You'll note that they're very similar to some tests for the `move` command in that they directly manipulate a repo and test against `smartlog` output. If that's to "end-to-end-y", I can redo them, but I found the `smartlog` output very easy to work with in those tests. Since the core of this feature is really the ability to add numerous, arbitrary ranges to the rebase planner, I tried to be exhaustive in the tests. If they're too much (or not enough!) I can add or remove as desired.

I've updated the main PR description w/ some more overview details.

git-branchless-lib/src/core/rewrite/plan.rs
arxanas: This looks fine to me as-is. There are a few options:

- Produce an error in this case.
- Convert these commits to be merge commits (multi-parent commits). It would probably be better to to make `move_range` accept a set of end OIDs, rather than have the caller call the same function multiple times.

git-branchless/src/commands/move.rs
```diff
@@ -169,7 +170,57 @@ pub fn r#move(
...
+            for dest_child in commit_set_to_vec_unsorted(&dest_children)? {
+                if dag
+                    .query()
+                    .descendants(CommitSet::from(dest_child))?
+                    .contains(&source_oid.into())?
+                {
+                    // If this child subtree actually contains the source
+                    // subtree being moved, then we should only move the commit
+                    // range *up to* the source subtree, not the entire child
+                    // subtree.
```
arxanas, line 207: [no action required] I think there's an argument to be made that this logic ought to be in the constraint solver itself, i.e. transform any `move_subtree` constraints into `move_range` constraints if they would overlap without another `move_subtree` constraint. That being said, it sounds hard, and isn't necessary right now.

arxanas (approved):
Everything looks great! The tests are pretty thorough, too. I like how you set up the `create_and_execute_plan` fixture to test the rebase planner directly.

I found one bug for moving a subtree with merge commits and added a couple of failing tests to document it, but I don't think it's worth addressing. The behavior before this stack was that it would exit with an error about a cycle (confusingly, because there is no obvious cycle in that case). Now it leaves the merge commit unmoved, so it needs to be restacked manually, but I think that's an improvement overall, actually.

arxanas:
> Basically, in the above example, moving the 2 commits C and D will produce different results than if we were to move the subtree @ C and then commit D (in either order). In the former, we want E to end up @ B (as covered in previous comments), but in the latter we want it to end up @ C, even though C is also being moved.

I actually never thought of this case, so I'm glad you discovered it and figured it out 😅.

PR #431, opened 2022-06-28
fix(sync): fix performance regression from #418

arxanas:


martinvonz (commented):
Which performance do the refs impact? I've noticed that they make fetch slow. Anything else?

arxanas:
@martinvonz it's not the refs per se. The problem is that it's an O(n) operation in the size of the operation log. It would preserve any commit, even hidden ones. This commit reduces it to O(n) in the size of visible gears. Syncing the DAG happens before every DAG operation and added 1.5s to every smartlog, etc. invocation on my small repo.

PR #448, opened 2022-07-06
`git move`: Add `--range` and `--exact` options (closes #176)

### LINKED ISSUE #176: Support moving individual commits without their descendants
This functionality could be exposed via something like `git move --exact <commit> --dest <commit>`. More important, this is necessary infrastructure to implement some form of interactive rebase.

claytonrcarter:
Still a WIP, but seems to be working. I've rebased my work onto #447 (which is also WIP) and I'll update this as that PR develops. I'll probably add some more tests, too, as I wrote all of these w/ a pre-revset mindset, so I was only thinking about moving single ranges/commits. 

- Adds a `git move --range <revset>` option to move ranges of commits.
- Adds a `git move --exact <revset>` option to move single commits.
- Updates CHANGELOG for #400, #447 and #448

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -18,6 +18,14 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Forbid nocommit string
+        run: |
+          if results=$(git grep --column --line-number --only-matching '@''nocommit'); then
+            echo "$results"
+            awk <<<"$results" -F ':' '{ print "::error file=" $1 ",line=" $2 ",col=" $3 "::Illegal string: " $4 }'
+            exit 1
+          fi
+
       - name: Set up Rust
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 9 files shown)


arxanas:
Do you think it's possible to support just an `--exact` flag without the `--range` flag? That is, can we preserve the ancestor relationships in the revset given by `--exact`? That seems like it would be the most general solution and would support situations which aren't precise ranges. For example, if you have

```mermaid
graph RL
  E --> D --> C --> B --> A
  F --> A
```

and you want to issue `git move --exact (B | D | E) --dest F`, it would ideally end up looking like this:

```mermaid
graph RL
  C --> A
  E --> D --> B -->  F --> A
```

In doing so, all ancestor relationships have been preserved: B was an ancestor of D before the move, so it remains one afterwards as well.

Also, what do you think about `-x` as the short version of the flag `--exact`, rather than `-e`? (For me, it also has connotations from Vim's `x` to cut/delete 😄.)

arxanas:
Oops, it looks like this PR was closed because the base branch `arxanas/revset` no longer exists. (It would have been better if it updated the PR with the new base branch instead...) I can't update the base branch on this PR myself, so you'll have to re-open or re-create the PR.

PR #450, opened 2022-07-07
`git move`: Add `--exact` option (closes #176)

### LINKED ISSUE #176: Support moving individual commits without their descendants

claytonrcarter:
_[reopening #448]_

Still a WIP, but seems to be working. I've rebased my work onto #447 (which is also WIP) and I'll update this as that PR develops. I'll probably add some more tests, too, as I wrote all of these w/ a pre-revset mindset, so I was only thinking about moving single ranges/commits. 

- ~~Adds a `git move --range <revset>` option to move ranges of commits.~~
- Adds a `git move --exact <revset>` option (shorthand: `-x`) to move single commits and/or ranges of commits.
- Updates CHANGELOG for #400, #447 and #448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - EXPERIMENTAL: (#382) Working copy snapshots are created before potentially-destructive operations in order to improve the capabilities of `git undo`.
   - Working copy snapshots are taken by default. Disable by setting `branchless.undo.createSnapshots` to `false`.
 - EXPERIMENTAL: (#391) created the `git record` command, which opens an interactive change selector for committing.
+- Numerous updates to `git branchless move`:
+  - (#400) Added `--insert` option to to insert the moved subtree or commits into the commit tree between the destination commit and its children, if any.
+  - (#450) Added `--exact` option to move a specific set of commits elsewhere in the commit tree. They will be removed from their current location and any unmoved children (if any) will be moved to their closest unmoved ancestor. The moved commits will maintain their overall ancestry structure.
+  - (#447) `--source`, `--base` and `--exact` options can all be provided multiple times to move multiple subtress, ranges and/or commits to the destination.
 - Added `--yes` option to `git undo` to skip confirmation.
 - (#366) Added bulk edit mode to `git reword` to allow updating multiple, individual commit messages at once.
 - (#398) Added support for [revset expressions](https://github.com/arxanas/git-branchless/wiki/Reference:-Revsets) in most commands.
--- a/git-branchless-lib/src/core/dag.rs
+++ b/git-branchless-lib/src/core/dag.rs
@@ -421,6 +421,51 @@ impl Dag {
         };
         Ok(Some(path))
     }
+
+    /// Given a CommitSet, return a list of CommitSets, each representing a
+    /// connected component of the set.
+    ///
+    /// For example, if the DAG contains commits A-B-C-D-E-F and the given
+    /// CommitSet contains `B, C, E`, this will return 2 `CommitSet`s: 1
+    /// containing `B, C` and another containing only `E`
+    #[instrument]
+    pub fn get_connected_components(&self, commit_set: &CommitSet) -> eyre::Result<Vec<CommitSet>> {
+        let mut components: Vec<CommitSet> = Vec::new();
+        let mut component = CommitSet::empty();
+        let mut commits_to_connect = commit_set.clone();
+
+        // FIXME: O(n^2) algorithm (
+        // FMI see https://github.com/arxanas/git-branchless/pull/450#issuecomment-1188391763
+        for commit in commit_set_to_vec_unsorted(commit_set)? {
+            if commits_to_connect.is_empty()? {
+                break;
+            }
+
+            if !commits_to_connect.contains(&commit.into())? {
+                continue;
+            }
+
+            let mut commits = CommitSet::from(commit);
+            while !commits.is_empty()? {
+                component = component.union(&commits);
+                commits_to_connect = commits_to_connect.difference(&commits);
+
+                let parents = self.query().parents(commits.clone())?;
+                let children = self.query().children(commits.clone())?;
+                commits = parents.union(&children).intersection(&commits_to_connect);
+            }
+
+            components.push(component);
+            component = CommitSet::empty();
+        }
+
+        let connected_commits = union_all(&components);
+        assert_eq!(commit_set.count()?, connected_commits.count()?);
+        let connected_commits = commit_set.intersection(&connected_commits);
+        assert_eq!(commit_set.count()?, connected_commits.count()?);
+
+        Ok(components)
+    }
 }
 
 impl std::fmt::Debug for Dag {
--- a/git-branchless/src/commands/move.rs
+++ b/git-branchless/src/commands/move.rs
@@ -3,6 +3,7 @@
 //! Under the hood, this makes use of Git's advanced rebase functionality, which
 //! is also used to preserve merge commits using the `--rebase-merges` option.
 
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt::Write;
 use std::time::SystemTime;
@@ -16,7 +17,7 @@ use tracing::instrument;
 use crate::opts::{MoveOptions, Revset};
 use crate::revset::resolve_commits;
 use lib::core::config::get_restack_preserve_timestamps;
-use lib::core::dag::{commit_set_to_vec_unsorted, union_all, CommitSet, Dag};
+use lib::core::dag::{commit_set_to_vec_unsorted, sorted_commit_set, union_all, CommitSet, Dag};
 use lib::core::effects::Effects;
 use lib::core::eventlog::{EventLogDb, EventReplayer};
 use lib::core::rewrite::{
@@ -59,10 +60,12 @@ pub fn r#move(
     sources: Vec<Revset>,
     dest: Option<Revset>,
     bases: Vec<Revset>,
+    exacts: Vec<Revset>,
     insert: bool,
     move_options: &MoveOptions,
 ) -> eyre::Result<ExitCode> {
-    let should_sources_default_to_head = sources.is_empty() && bases.is_empty();
+    let should_sources_default_to_head =
+        sources.is_empty() && bases.is_empty() && exacts.is_empty();
 
     let repo = Repo::from_current_dir()?;
     let head_oid = repo.get_head_info()?.oid;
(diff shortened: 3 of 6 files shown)


claytonrcarter:
🤔 I could've sworn I commented on this already, but it looks like I forgot to press the big green submit button.

This is still WIP, and the code is full of untold horrors that I need to go back through and rethink/clean up (and remove my local working commits, like adding `roots()` to revsets, etc), but I think it's implementing the gist of what you asked about in https://github.com/arxanas/git-branchless/pull/448#issuecomment-1176984865

It's been a *long* time since I've seriously thought about big O notation, so I'm hesitant to make any guesses about my `get_connected_components` implementations, but I have 2 of them just for fun. I *think* the second one is more efficient, and may operate in `O(n)`, but if you factor in all of the DAG queries then it becomes `O(over my head)`. 😄 Any comments you have on these would be great; thank you. (I figure we'd just keep 1, but I put in 2 in case 1 was more correct or performant or something.)

Working on this w/ connected components vs my previous approach (that only considered semi-linear "ranges") turned up a *bunch* of interesting edge cases, many of them involving merge commits. I've tried to document some or all of them in tests, and they may need further attention as this PR solidifies.

There are a few more edge cases that I need to code for (eg moving and *inserting* multiple components with various "non-stick" structures) but at this point I think I need to just dogfood it for a while to see how it feels and works, updating this as I do. 

Thank you!

git-branchless/src/commands/move.rs
```diff
@@ -168,16 +217,163 @@ pub fn r#move(
...
+                    }
+                } else {
+                    // TODO this should be unreachable
+                    dest_oid
+                }
+            };
+
+            let component_parent = match dag.get_only_parent_oid(*component_root) {
+                Ok(oid) => oid,
+                Err(_) => {
```
arxanas, line 321: In this case, `dag.get_only_parent_oid` could have had an I/O error, which we would want to propagate directly. Instead, either make `get_only_parent_oid` return an `eyre::Result<Option<_>>`, or get all of the parents and do a pattern match.

git-branchless/src/commands/move.rs
```diff
@@ -168,16 +217,163 @@ pub fn r#move(
...
+                    )?;
+                    return Ok(ExitCode(1));
+                }
+            };
+
+            let component_children: CommitSet = dag
+                .query()
+                .children(component.clone())?
+                .difference(&dag.obsolete_commits)
+                .difference(component);
```
arxanas, line 334:
```suggestion
                .difference(component)
                .difference(&dag.obsolete_commits);
```

It probably doesn't matter too much in this case, but let's do this in the opposite order. Generally, we want to operate on the smallest sets first for performance.

git-branchless/src/commands/move.rs
```diff
@@ -187,33 +383,13 @@ pub fn r#move(
             let dest_children: CommitSet = dag
                 .query()
                 .children(CommitSet::from(dest_oid))?
-                .difference(&dag.obsolete_commits);
+                .difference(&dag.obsolete_commits)
+                // In case a source OID is already a child of dest_oid.
+                .difference(&source_oids)
+                .difference(&exact_oids);
```
arxanas, line 389:
```suggestion
                // In case a source OID is already a child of dest_oid.
                .difference(&source_oids)
                .difference(&exact_oids)
                .difference(&dag.obsolete_commits);
```

Same regarding performance and ordering.

git-branchless/tests/command/test_move.rs
```diff
@@ -361,51 +454,1397 @@ fn test_move_insert_tree() -> eyre::Result<()> {
...
-            |
-            @ bf0d52a create test4.txt
-            |\
-            | o 44352d0 create test2.txt
-            |
-            o 0a4a701 create test3.txt
-            "###);
-        }
+        // insta::assert_snapshot!(stderr, @r"");
+
```
arxanas, line 463:
```suggestion

```

git-branchless/tests/command/test_move.rs
```diff
@@ -361,51 +454,1397 @@ fn test_move_insert_tree() -> eyre::Result<()> {
...
+        |
+        @ 8fcf7dd create test4.txt
+        "###);
+    }
+
+    Ok(())
+}
+
+#[test]
+fn test_move_exact_insert_swap() -> eyre::Result<()> {
```
arxanas, line 831: It's super cool that you can do this now!

git-branchless/tests/command/test_move.rs
```diff
@@ -361,51 +454,1397 @@ fn test_move_insert_tree() -> eyre::Result<()> {
...
+    |
+    o 96d1c37 create test2.txt
+    |\
+    | o 70deb1e create test3.txt
+    |
+    @ f57e36f create test4.txt
+    "###);
+
+    // --on-disk
+    {
```
arxanas, line 937: Uncomment this test?

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - EXPERIMENTAL: (#382) Working copy snapshots are created before potentially-destructive operations in order to improve the capabilities of `git undo`.
   - Working copy snapshots are taken by default. Disable by setting `branchless.undo.createSnapshots` to `false`.
 - EXPERIMENTAL: (#391) created the `git record` command, which opens an interactive change selector for committing.
+- Numerous updates to `git branchless move`:
+  - (#400) Added `--insert` option to to insert the moved subtree or commits into the commit tree between the destination commit and it's children, if any.
```
arxanas, line 18:
```suggestion
  - (#400) Added `--insert` option to to insert the moved subtree or commits into the commit tree between the destination commit and its children, if any.
```

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - (#450) Added `--exact` option to move a specific set of commits elsewhere in the commit tree. They will be removed from their current location and any unmoved children (if any) will be moved to their closest unmoved ancestor. The moved commits will maintain their overall ancestry structure.
+  - (#447) `--source`, `--base`, `--range` and `--exact` options can all be provided multiple times and in any combination to move multiple subtress, ranges and/or commits to the destination.
```
arxanas, line 20: - I think we no longer have the `--range` flag.
- I think these flags can be provided multiple times, but not in arbitrary combinations.

arxanas (commented):
Impressive work!

git-branchless/tests/command/test_move.rs
claytonrcarter: Yeah, sorry, there were a lot of WIP artifacts still lingering in here. I've removed a bunch of them, but still need to give this another proof read to make sure I've gotten all of them.

git-branchless/src/commands/move.rs
claytonrcarter: Ah, OK, thank you. That makes sense!

git-branchless/src/commands/move.rs
claytonrcarter: OK, done, thank you. 

Do you have tips on where to look to know if something will fail due to I/O vs just failing for normal use situations?

git-branchless/src/commands/move.rs
```diff
@@ -168,16 +217,163 @@ pub fn r#move(
...
+                        if component
+                            .contains(&(*component_root).into())
+                            .unwrap_or(false)
+                        {
+                            return false;
+                        }
+                    } else {
+                        // TODO uh oh component not found??
+                    }
+
```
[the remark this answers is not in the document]
claytonrcarter: Done, but heads up: this whole block has changed.

git-branchless-lib/src/core/dag.rs
```diff
@@ -421,6 +421,108 @@ impl Dag {
...
+                    parents => {
+                        // This root has multiple parents, so find each
+                        // component that contains any of the parents, remove
+                        // each from the list of components, combine them all
+                        // and then add that a new component.
+                        let mut new_component = CommitSet::empty();
+                        for component in &mut components {
+                            // FIXME if a parent is in a component, should we
+                            // remove it from our search list? Like, can it
+                            // exist in multiple components? I don't think so...
```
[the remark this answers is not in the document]
claytonrcarter: Heads up: I had included 2 implementations of `get_connected_components` in this PR b/c I couldn't tell which I preferred. I've settled on one, but it's not the one you commented on, so you'll probably want to double check the remaining implementation.

git-branchless-lib/src/core/dag.rs
```diff
@@ -421,6 +421,108 @@ impl Dag {
...
+                            .into_iter()
+                            .filter(|c| c.is_empty().unwrap_or(false)) // FIXME should Err be false?
+                            .collect();
+                    }
+                }
+            }
+            commits_to_connect = commits_to_connect.difference(&roots);
+        }
+
+        Ok(components)
```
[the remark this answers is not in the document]
claytonrcarter: Done! Please let me know what you think of how I went about it.

git-branchless/src/commands/move.rs
```diff
@@ -168,16 +217,163 @@ pub fn r#move(
...
+                // from component_roots, which has been sorted topologically; so
+                // each included component should "come after" the previous
+                // component.
+                let nearest_dest_index = possible_destinations.len() - 1;
+                if possible_destinations.len() > 1
+                    && !dag.query().is_ancestor(
+                        possible_destinations[nearest_dest_index - 1].into(),
+                        possible_destinations[nearest_dest_index].into(),
+                    )?
+                {
```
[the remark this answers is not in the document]
claytonrcarter: Funny. I had originally been checking all of them, but then removed it to just check the last 2; but I hadn't thought of this case apparently. Done!

CHANGELOG.md
claytonrcarter: Yup, I totally missed that. Sorry, and thanks for catching it!

claytonrcarter: Not for this PR, of course, but I do wonder if the current implementation would actually support "arbitrary combinations" of these flags if we just removed the clap `conflicts_with` attributes. (Or if it could do so w/ minimal changes.) I don't know that it would be *useful* to do so, though...

claytonrcarter:
Sorry for the false start and all of the WIP artifacts you had to wade through on this. I still consider this WIP pending some more dogfooding and proofreading, but I think I've addressed all of your initial comments.

This has been a joy to use locally, but I have mostly been moving single commits. I haven't found myself moving multiple commits, ranges or "range trees" yet. I mean ... the tests say that it works, but I'll be interested in feedback on what happens IRL. 😄

arxanas:
I've been using this for a bit and I haven't noticed any issues so far. What do you say we merge it and fix any issues in follow-up commits?

> I think the second one is more efficient, and may operate in O(n)

I never answered this. I think both implementations are O(n^2) because you have one loop over the commits and a nested loop over the existing components to merge any intersecting components. You can do better with a data structure which can merge intersecting components efficiently (disjoint set), but otherwise I think you're likely to have O(n^2) complexity. Mostly we're hoping that we won't have to compute connected components for a large number of commits. If that happens, we can update it to use a better data structure in the future.

claytonrcarter:
Thanks for the nudge. I've read through this again, pushed some clean up and I think it's good enough for now. A couple of notes on what I just pushed:

- Mostly formatting and comments.
- Moved some error checking earlier in the process (checking for 1 parent/component is now consistent w/ checking for 1 root and 1 head).
- I've added a test for moving non-connected components, such as `D`&`F` in the following example.
I'm still 😬 about merging my own code, but since you've signed off on this already, and what I pushed are just minor fixes, I've set it to rebase/merge when CI finishes. Thank you!

One thing that sticks out at me that I'd like to fix in a follow up is that I think the restriction to "only move components with 1 head" is lazy. I don't see any reason we can't `move --exact` the pink commits in something like this:

```mermaid
graph RL
D --> C --> B --> A
F --> E --> B

classDef pink fill:#f9f;
class B,C,E pink;
```

I think that restriction predates the component approach and will be straightforward to support. I'll see what I can do. *\*cracks knuckles\**

PR #451, opened 2022-07-10
`git record` optimizations

arxanas:


arxanas:
I think I'll have to rewrite the `git record` component anyways for both performance and correctness reasons, so further development here is moot.

PR #466, opened 2022-07-19
feat(move --exact): Remove single HEAD restriction

claytonrcarter:
#450 included a restriction that each component or range of commits being moved had to have a single HEAD. This restriction was from an earlier implementation of `move --exact` that didn't use connected components, and -- as far as I can tell -- isn't actually needed. See comment https://github.com/arxanas/git-branchless/pull/450#issuecomment-1188497505 for an example.

@arxanas can you take a peek at [this test fixture that I updated](https://github.com/arxanas/git-branchless/pull/466/files#diff-40097149f604160704079190c192ae81be5fbfc762b3d6e6e703be79e82a2e07L2032-L2043). It's hard to follow in the diff, but basically it looks like it changed from `test3 <-- test6` to `test2 <-- test6` ... which I *think* is actually correct. (Everything else appears to have stayed the same.) Can you take a peek at it and let me know what you think?

### DIFF
--- a/git-branchless/src/commands/move.rs
+++ b/git-branchless/src/commands/move.rs
@@ -142,18 +142,6 @@ pub fn r#move(
                     return Ok(ExitCode(1));
                 };
 
-                let component_heads = dag.query().heads(component.clone())?;
-                if component_heads.count()? != 1 {
-                    writeln!(
-                        effects.get_error_stream(),
-                        "The --exact flag can only be used to move ranges with exactly 1 head.\n\
-                         Received range with {} heads: {:?}",
-                        component_heads.count()?,
-                        component_heads
-                    )?;
-                    return Ok(ExitCode(1));
-                };
-
                 components.insert(component_root, component);
             }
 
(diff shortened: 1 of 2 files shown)


git-branchless/src/commands/move.rs
```diff
@@ -142,18 +142,6 @@ pub fn r#move(
...
-                    writeln!(
-                        effects.get_error_stream(),
-                        "The --exact flag can only be used to move ranges with exactly 1 head.\n\
-                         Received range with {} heads: {:?}",
-                        component_heads.count()?,
-                        component_heads
-                    )?;
-                    return Ok(ExitCode(1));
-                };
-
```
claytonrcarter, line 156: It's worth noting that `--insert` still has a 1 HEAD restriction, but that's already handled on line 364 and it's not unique to `--exact` (ie it also applies to subtrees).

git-branchless/src/commands/move.rs
```diff
@@ -338,7 +322,7 @@ pub fn r#move(
                 }
             }
 
-            builder.move_range(component_root, component_head, component_dest_oid)?;
+            builder.move_subtree(component_root, component_dest_oid)?;
```
claytonrcarter, line 325: The children of each component are already being moved (in the `for` loop that immediately precedes this statement), so the call to `move_range` was just unnecessary.

arxanas:
The test output looks right to me. Thanks for the enhancement!

PR #467, opened 2022-07-19
feat(revsets): Add `roots`, `heads` and `children` functions

claytonrcarter:
This is a simple one: it just adds `roots`, `heads` and `children` functions to the revset evaluator. In particular, the addition of `children()` would allow the alias for `git drop` (as [described here](https://github.com/arxanas/git-branchless/discussions/48#discussioncomment-2360697)) to work via the git alias `drop='!f() { git move -x "children($1)" -d "$1"^; git hide "$1"; }; f'`

And `heads` and `roots` were just added while I was there. It felt like it could be handy to calculate the first commit in a stack (eg `roots(stack())`) and such.


arxanas (approved):

claytonrcarter:
Thank you. I also updated the revset wiki at https://github.com/arxanas/git-branchless/wiki/Reference:-Revsets

PR #472, opened 2022-07-26
Support passing revsets to `git smartlog`

arxanas:


git-branchless/src/revset/grammar.lalrpop
```diff
@@ -1,110 +1,97 @@
...
-    // "not " => "negate",
+    // For Mercurial users' familiarity.
+    <lhs:Expr3> "::"  <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("range"),       vec![lhs, rhs]),
+    <lhs:Expr3> "::"              =>  Expr::FunctionCall(Cow::Borrowed("descendants"), vec![lhs,    ]),
+                "::"  <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("ancestors"),   vec![     rhs]),
 
-    "::" => "ancestors",
+    <lhs:Expr3> ".." <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("only"), vec![lhs, rhs]),
+    <lhs:Expr3> ".."             =>  Expr::FunctionCall(Cow::Borrowed("only"), vec![lhs, Expr::Name(Cow::Borrowed("."))]),
+                ".." <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("only"), vec![Expr::Name(Cow::Borrowed(".")), rhs]),
```
claytonrcarter, line 37: Would you consider reversing the `lhs, rhs` (ie `vec![rhs, lhs]`)  so that `foo..bar` would mean `only(bar, foo)`? This would allow revset's evaluation of `..` to match (as I understand it) how `git` handles double dot notation, [which is a pretty common notation in standard `git`](https://www.git-scm.com/docs/gitrevisions#_dotted_range_notations). 

To illustrate my concern, here's some example output from this branch:
```
❯ git co arxanas/revset
branchless: creating working copy snapshot
branchless: running command: git checkout arxanas/revset
Already on 'arxanas/revset'
Your branch is up to date with 'origin/arxanas/revset'.
branchless: processing checkout
⋮
◇ 0f06699 16h (master, remote origin/master) build: run `cargo upgrade`
┃
◯ 7e512e6 15h build(revset): build parser as part of `build.rs`
┃
◯ 87bc6fd 15h tests(eval): create `eval_and_sort` function
┃
◯ 96b891e 15h feat(revset): add `not` function
┃
◯ c9ec768 15h feat(revset): produce friendlier error messages
┃
◯ 3793bfd 15h refactor(revset): rewrite grammar to support `:` as a mixfix operator
┃
● bc0c8b7 15h (ᐅ arxanas/revset) feat(smartlog): support passing an arbitrary revset expression

❯ git rev-list master..@
bc0c8b750f514883c8abca09246fa72637b9be08
3793bfdb721ccb2c6e0b2bbed0bd36ceced3174c
c9ec76842ca33839b9770fb98f3f51185594e8cc
96b891edb5ab4dbda1154ea6f22c38317a518598
87bc6fdc6dd3fb50dc05e5b11c81067adea502f4
7e512e620b19925129669b5f11538c40b09c319e

❯ cargo run -- query master..@
    Finished dev [unoptimized] target(s) in 0.20s
     Running `target/debug/git-branchless query 'master..@'`
# no output

❯ cargo run -- query @..master
    Finished dev [unoptimized] target(s) in 0.90s
     Running `target/debug/git-branchless query '@..master'`
bc0c8b750f514883c8abca09246fa72637b9be08
3793bfdb721ccb2c6e0b2bbed0bd36ceced3174c
c9ec76842ca33839b9770fb98f3f51185594e8cc
96b891edb5ab4dbda1154ea6f22c38317a518598
87bc6fdc6dd3fb50dc05e5b11c81067adea502f4
7e512e620b19925129669b5f11538c40b09c319e

❯ git rev-list @..master
# no output

```

git-branchless/src/revset/grammar.lalrpop
```diff
@@ -1,110 +1,97 @@
...
-    // negation unimplemented for now
-    // "!" => "negate",
-    // "not " => "negate",
+    // For Mercurial users' familiarity.
+    <lhs:Expr3> "::"  <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("range"),       vec![lhs, rhs]),
+    <lhs:Expr3> "::"              =>  Expr::FunctionCall(Cow::Borrowed("descendants"), vec![lhs,    ]),
+                "::"  <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("ancestors"),   vec![     rhs]),
 
-    "::" => "ancestors",
+    <lhs:Expr3> ".." <rhs:Expr4> =>  Expr::FunctionCall(Cow::Borrowed("only"), vec![lhs, rhs]),
```
claytonrcarter, line 35: This is out of scope for this PR, but it's similar to my other comment and may be easy to add while you're here: would you consider adding a three dot notation, again for familiarity/parity with `git`s native `...` notation? If I'm not mistaken, it would be something like `union(only(lhs, rhs), only(rhs, lhs))`.

To be fair, though, I'm personally ambivalent to triple dot ... I don't think I've ever used it. It just strikes me as something that a `git` might user might expect to work out of the box when first learning `git branchless`.

claytonrcarter:
I'm really excited about this! I know it's still in draft, but what do you think about rendering **only** the commits in the revset when a revset is passed in? Or perhaps the _intersection_ of the default observed commits and the given revset, to avoid something like `git smartlog :@` rendering everything since the beginning of time?

As is, it looks like all commits between `<main branch>` and `HEAD`, plus all branches, are rendered regardless of what is included in the revset. This seems counterintuitive when doing something like `smartlog master::`, `smartlog HEAD~5:HEAD~3`, because, though those commits _are_ included, so is a bunch of other stuff that I didn't really want.

Aside from these contrived cases, I can see myself wanting to do something like:
- `smartlog master::` - just show my work off of master
- `smartlog stack()` - just show my current stack
- `smartlog 'roots(stack())::` - just show the work based on my current stack

I may be misunderstanding the use case, though. If that's the case, please feel to ignore me, of course. 😄

git-branchless/src/revset/grammar.lalrpop
arxanas: Oops, thanks for catching! Fixed.

arxanas:
> Or perhaps the *intersection* of the default observed commits and the given revset

> As is, it looks like all commits between <main branch> and HEAD, plus all branches, are rendered regardless of what is included in the revset

The unfortunate problem is that all of those commits are in the set of "default observed commits" 😅, which is why the problem is happening. I've been trying to fix this, but I think some tests are failing because they try to render the smartlog in the past, while the revset queries aren't properly equipped to deal with that right now. You can try checking out https://github.com/arxanas/git-branchless/pull/474 to see if the behavior is more intuitive.

Unfortunately, I haven't come up with a way to support not showing the ancestor commits up to the main branch. Currently, it's a hard-coded assumption.

git-branchless/src/revset/grammar.lalrpop
arxanas: I'm open to it, but I've never actually used `...` either, so I'm going to defer it for now 😄.

claytonrcarter: Makes sense to me; thanks for considering!

Issue #355, opened 2022-04-13
git branchless sync --pull not pulling main branch


tp-woven:
### Description of the bug

When I run `git branchless sync --pull`, it fetches the "main" branch from the remote and pulls it into my "feature" branches, but it does not update the "main" branch itself.

For example: (irrelevant branches and logs were removed for simplicity)

```
❯ git sl
⋮
◆ f6217c4 1d (remote origin/main, ᐅ main) Merge pull request #XXX from XXXXXXX
❯ git branchless sync --pull
...
branchless: running command: git checkout main
Switched to branch 'main'
Your branch is behind 'origin/main' by 39 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
...
❯ git sl
⋮
◆ f6217c4 1d (ᐅ main) Merge pull request #XXX from XXXXXXX
⋮
◇ 3aafdf4 39m (remote origin/main) Merge pull request #YYY from YYYYYYY
```

### Expected behavior

The main branch should be "pulled" and updated to reference the latest remote ref. In the example above, I would have expected `main` to also be moved to `3aafdf4` (same as `origin/main`).

### Actual behavior

The main branch is unaffected by `sync` and needs to be `pull`ed manually.

### Version of `git-branchless`

git-branchless 0.3.12

### Version of `git`

git version 2.25.1

### Version of `rustc`

rustc 1.60.0-nightly (bd3cb5256 2022-01-16)

#### Software version

git-branchless 0.3.12

#### Operating system

Linux 5.4.0-1045-aws

#### Command-line

```bash
/home/devec2/.cargo/bin/git-branchless bug-report
```

#### Environment variables

```bash
SHELL=/usr/bin/zsh
EDITOR=<not set>
```

#### Git version

```
> git version
git version 2.25.1
```

#### Events

<details>
<summary>Show 5 events</summary>

##### Event ID: 179, transaction ID: 188

1. `RefUpdateEvent { timestamp: 1649820518.0892587, event_tx_id: EventTransactionId(188), ref_name: "refs/heads/redacted-ref-0", old_oid: a1ac01f8528c336b64f41e54baeb3d430f6f5dc5, new_oid: 0602d383e956d811831245264a8d3c13a755ffc3, message: None }`
[...]

PR #485, opened 2022-08-02
Fix `git sync` deleting main branch

arxanas:


arxanas:
This works for the case of the main branch, but I don't think the general behavior for remote-tracking branches is correct. We should probably rethink it. I suspect that the best solution will also handle #355 somehow.

PR #505, opened 2022-08-19
docs: fix inline code markup

kidonng:
Replace markdown code markup with HTML as it can't be recognized when interleaved with the latter.

Also changed `git rebase -i` to `git rebase --interactive` to be more clear and consistent with the content above (`git log --graph`).


arxanas:
Hi @kidonng, thanks for the fixes! 😄

PR #508, opened 2022-08-21
feat(revset): Add `exactly` function

claytonrcarter:
As discussed in discord, this adds an `exactly` revset function. This returns the given revset only if it contains the expected number of commits. Otherwise it errors out. This was very straightforward to implement and includes tests.

**Notes**

- I considered creating a more general usage `count` function that would behave like the text/date matchers, for example:
  - `count(<revset>, exactly:1)`
  - `count(<revset>, greater_than:2)`
  - `count(<revset>, less_than_or_equal_to:3)`, etc. 
  - While that would be cool, it was more than I needed at this time.
- At first I had also added a `sole()` function, but I removed that in favor of making it an alias for `exactly($1, 1)` using the revset aliases PR.
- I have already found this useful in scripting to give me added confidence that some automated task won't target more than I expect it to. I didn't expect this, but I like where it's going; for example I might write a script guard like `if ! git query exactly(parents($INPUT), 1) ; then exit ; fi` to prevent myself from accidentally operating on a merge commit.
- The `ExpectedSingletonSet` appears to be unused. Should it be removed or would you like to keep it around?

### DIFF
--- a/git-branchless/src/revset/eval.rs
+++ b/git-branchless/src/revset/eval.rs
@@ -10,6 +10,7 @@ use once_cell::unsync::OnceCell;
 use thiserror::Error;
 
 use lib::core::dag::{CommitSet, Dag};
+use lib::core::formatting::Pluralize;
 use lib::git::{Repo, ResolvedReferenceInfo};
 use tracing::instrument;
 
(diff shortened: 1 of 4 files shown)


git-branchless/src/revset/eval.rs
```diff
@@ -107,6 +108,20 @@ pub enum EvalError {
...
 
+    #[error(
+        "expected '{expr}' to evaluate to {}, but got {actual_len}",
+        Pluralize {
+            determiner: None,
+            amount: *expected_len,
+            unit: ("element", "elements"),
+        }
+    )]
+    ExpectedLengthSet {
```
arxanas, line 119: Maybe this should be called `ExpectedLengthOfSet` or `ExpectedLengthSet`? I don't think it's quite grammatical right now.

arxanas (approved):
> At first I had also added a sole() function, but I removed that in favor of making it an alias for exactly($1, 1) using the revset aliases PR.

Sounds good.

> The ExpectedSingletonSet appears to be unused. Should it be removed or would you like to keep it around?

Oops, yeah, we can remove that. I forget what I wanted to use that for.

git-branchless/src/revset/eval.rs
claytonrcarter: Sure thing. I think I was trying (too hard) for parity with `ExpectedSingletonSet`. I've changed this to `UnexpectedSetLength`.

claytonrcarter:
FYI I've added `exactly` to the revset wiki. ✅

PR #509, opened 2022-08-21
feat(revset): Support custom revset aliases

claytonrcarter:
This was fun to implement, and was pretty easy. Considering how flexible and powerful it could be ... I'm sort of concerned I'm overlooking something! 😄  As always, I'm 100% open to suggestions here.

- This does nothing to check for -- let alone prevent -- recursion in aliases. If someone defines a `foo` alias that itself calls `foo()` ... they're out of luck.
- This only allows aliasing functions; I did not consider allowing aliases for names, though it would be easy to add. 
  - I kind of feel like it's not necessary, though. Aliasing names risks conflicting w/ commits and refs, while functions have no such conflict. And aliases feel a little more "functiony" anyway.
- I made no attempt to include the defined aliases in the `EvalError::UnboundFunction` output. It wasn't immediately obvious to me how to list all config keys defined in the section, but I should admit that I didn't look very hard. :smile: Including them would certainly be kinder and would aid in discoverability.
- This has me wondering about `git branchless` shipping w/ some predefined aliases of it's own, which would be installed via `init`. Doing so would allow us to use Rust for functions that need some intricate logic, while aliases could be used for queries composable from the Rust-based "building block" functions. I'm thinking of something like `siblings=children(parents($1)) - $1`, `onlyParent=exactly(parents($1), 1)`, `sole=exactly($1, 1)`, etc
  - One downside to this, though, is that errors might end up being surprising. For example, if I call `siblings(foo)` but get an error about `children` or `parents`, perhaps that could be confusing?
  - Anyway, just an idea, not in scope for this PR.
- I chose `branchless.revsets.alias` as the config key, and this shows up in the `.config/git/config` file like so:
```
[branchless "revsets.alias"]
    grandChildren = children(children($1))
    oldest = sole(roots(intersection(ancestors($1), descendants($1))))
    sole = exactly($1, 1)
    onlyChild = sole(children($1))
```
  - I don't have strong feelings about the config key, and I'm happy to change to match whatever you'd prefer. ~~As I write this, though, I see that I should drop the `git-` and just use `branchless.revsets.alias` ... I'll have to fix that tomorrow~~ Fixed.

Thank you!

### DIFF
--- a/git-branchless/src/revset/ast.rs
+++ b/git-branchless/src/revset/ast.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::collections::HashMap;
 use std::fmt::Display;
 
 /// A node in the parsed AST.
@@ -22,6 +23,26 @@ impl Display for Expr<'_> {
                     write!(f, "{}", arg)?;
                 }
                 write!(f, ")")
+            }
+        }
+    }
+}
+
+impl<'input> Expr<'input> {
+    /// Replace names in this expression with arbitrary expressions.
+    ///
+    /// Given a HashMap of names to Expr's, build a new Expr by crawling this
+    /// one and replacing any names contained in the map with the corresponding
+    /// Expr.
+    pub fn replace_names(&self, map: &HashMap<String, Expr<'input>>) -> Expr<'input> {
+        match self {
+            Expr::Name(name) => match map.get(&name.to_string()) {
+                Some(expr) => expr.clone(),
+                None => self.clone(),
+            },
+            Expr::FunctionCall(name, args) => {
+                let args = args.iter().map(|arg| arg.replace_names(map)).collect();
+                Expr::FunctionCall(name.clone(), args)
             }
         }
     }
--- a/git-branchless/src/revset/eval.rs
+++ b/git-branchless/src/revset/eval.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::fmt::Display;
 use std::num::ParseIntError;
 use std::sync::Arc;
@@ -11,10 +12,11 @@ use thiserror::Error;
 
 use lib::core::dag::{CommitSet, Dag};
 use lib::core::formatting::Pluralize;
-use lib::git::{Repo, ResolvedReferenceInfo};
+use lib::git::{ConfigRead, Repo, ResolvedReferenceInfo};
 use tracing::instrument;
 
 use super::builtins::FUNCTIONS;
+use super::parser::{parse, ParseError};
 use super::pattern::{Pattern, PatternError};
 use super::Expr;
 
@@ -119,6 +121,9 @@ pub enum EvalError {
         actual_len: usize,
     },
 
+    #[error("failed to parse alias expression '{alias}'\n{source}")]
+    ParseAlias { alias: String, source: ParseError },
+
     #[error("not an integer: {from}")]
     ParseInt {
         #[from]
@@ -216,14 +221,36 @@ pub(super) fn eval_name(ctx: &mut Context, name: &str) -> EvalResult {
 }
 
 pub(super) fn eval_fn(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
-    let function = FUNCTIONS
-        .get(name)
-        .ok_or_else(|| EvalError::UnboundFunction {
-            name: name.to_owned(),
-            available_names: FUNCTIONS.keys().sorted().copied().collect(),
-        })?;
+    if let Some(function) = FUNCTIONS.get(name) {
+        return function(ctx, name, args);
+    }
 
-    function(ctx, name, args)
+    let alias_key = format!("branchless.revsets.alias.{}", name);
+    let alias_template: Option<String> = ctx
+        .repo
+        .get_readonly_config()
+        .map_err(EvalError::OtherError)?
+        .get(alias_key)
+        .map_err(EvalError::OtherError)?;
+    if let Some(alias_template) = alias_template {
+        let alias_expr = parse(&alias_template).map_err(|err| EvalError::ParseAlias {
+            alias: alias_template.clone(),
+            source: err,
+        })?;
+        let arg_map: HashMap<String, Expr> = args
+            .iter()
+            .enumerate()
+            .map(|(i, arg)| (format!("${}", i + 1), arg.clone()))
+            .collect();
+        let alias_expr = alias_expr.replace_names(&arg_map);
+        let commits = eval_inner(ctx, &alias_expr);
+        return commits;
+    }
+
+    Err(EvalError::UnboundFunction {
+        name: name.to_owned(),
+        available_names: FUNCTIONS.keys().sorted().copied().collect(),
+    })
 }
 
 #[instrument]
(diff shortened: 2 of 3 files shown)


claytonrcarter:
Oh dear! I just realized that I pushed these three recent branches directly to this repo, and not to my fork. I'm sorry about that! (It was a misfire w/ `git stack`.) I'll be more careful in the future. 🤦

git-branchless/src/revset/eval.rs
```diff
@@ -204,14 +206,33 @@ pub(super) fn eval_name(ctx: &mut Context, name: &str) -> EvalResult {
...
+        for (i, arg) in args.iter().enumerate() {
+            alias_template =
+                alias_template.replace(format!("${}", i + 1).as_str(), format!("{}", arg).as_str());
+        }
+        let expr = parse(&alias_template)
+            .wrap_err("Parsing alias expression")
+            .map_err(EvalError::OtherError)?;
+        let commits = eval_inner(ctx, &expr);
+        return commits;
+    }
```
claytonrcarter, line 230: It's worth calling attention to the ... hacky? ... crude? ... expeditious? way that I implemented this: When an unrecognized fn is encountered, we check to see if a corresponding alias is defined. If so, we serialize each element of `args` (from `Expr` back into `Strings`) and replace each placeholder in the alias "template" w/ the corresponding serialized arg. Then we re-parse the alias (w/ the placeholders replaced) and send the result back through `eval_inner`. 

I suspect that this will be fine for most cases, but if someone creates some complex aliases that also include other complex aliases, which include complex aliases, etc, etc, then this will result in a lot of serializing and re-parsing.

A more robust option might be to just parse the alias as is (w/ placeholders *not* replaced) but then pass some sort of "unbound arg resolver" callback through the eval system, and that callback could be used to resolve things like `$n` placeholders into parsed `Expr`s. I suspect that this might introduce some significant complexity into the eval system, though.

As I said, I'm assuming that the current approach will be OK for many or most use cases, but if you dislike it, I could look into an alternative approach. Or, we could go w/ this for now as an experiment and revisit it if we encounter performance issues.

claytonrcarter: For the sake of discussion, I've pushed a fixup commit w/ an alternate approach. Instead of serializing and reparsing expressions, this latest version works by:
1. Parsing the alias template as is, w/o replacing placeholders.
2. Crawls through the parsed alias template `Expr` and replaces any names in the AST that "match" the input args (ie `$1` from the template) w/ a copy of the corresponding `Expr` from the input (ie `myAlias(foo)` would end up replacing `$1` with `Expr:Name(foo)`)
3. Evaluating the new `Expr`

I'm not familiar w/ `Cow`, so there's a lot of `.clone()` happening in the new `replace_names()` fn, and I'm not sure if that's really how it's intended to be used. 100% open to pointers, as usual. If this over complicates things, I'm happy to revert it ... or expand/alter it if you'd prefer.

Thanks for considering!

git-branchless/src/revset/eval.rs
```diff
@@ -204,14 +207,35 @@ pub(super) fn eval_name(ctx: &mut Context, name: &str) -> EvalResult {
...
+        return function(ctx, name, args);
+    }
+
+    let alias_key = format!("branchless.revsets.alias.{}", name);
+    let alias_template = ctx
+        .repo
+        .get_readonly_config()
+        .map_err(EvalError::OtherError)?
+        .get::<String, String>(alias_key)
+        .map_err(EvalError::OtherError)?;
```
arxanas, line 220:
```suggestion
    let alias_template: String = ctx
        .repo
        .get_readonly_config()
        .map_err(EvalError::OtherError)?
        .get(alias_key)
        .map_err(EvalError::OtherError)?;
```

If you add a type annotation, you should be able to skip annotating the generic arguments manually.

git-branchless/src/revset/eval.rs
```diff
@@ -204,14 +207,35 @@ pub(super) fn eval_name(ctx: &mut Context, name: &str) -> EvalResult {
...
+    if let Some(alias_template) = alias_template {
+        let alias_expr = parse(&alias_template)
+            .wrap_err("Parsing alias expression")
+            .map_err(EvalError::OtherError)?;
```
arxanas, line 224: Can you add a test to track the behavior for when an alias has a parse expression? In particular, we should probably surface the alias value in the error message.

git-branchless/src/revset/eval.rs
```diff
@@ -204,14 +207,35 @@ pub(super) fn eval_name(ctx: &mut Context, name: &str) -> EvalResult {
...
+        let arg_map: HashMap<String, Expr> = args
+            .iter()
+            .enumerate()
+            .map(|(i, arg)| (format!("${}", i + 1), arg.clone()))
+            .collect();
+        let alias_expr = alias_expr.replace_names(&arg_map);
+        let commits = eval_inner(ctx, &alias_expr);
+        return commits;
```
arxanas, line 232: This is pretty elegant in my opinion 😃.

git-branchless/src/revset/ast.rs
```diff
@@ -26,3 +27,23 @@ impl Display for Expr<'_> {
...
     }
 }
+
+impl<'input> Expr<'input> {
+    /// Replace names in this expression with arbitrary expressions.
+    ///
+    /// Given a HashMap of names to Expr's, build a new Expr by crawling this
+    /// one and replacing any names contained in the map with the corresponding
+    /// Expr.
+    pub fn replace_names(&self, map: &HashMap<String, Expr<'input>>) -> Expr<'input> {
```
arxanas, line 37: One interesting topic is [macro hygiene](https://en.wikipedia.org/wiki/Hygienic_macro): whether names like `$1`/`$2`/etc. "leak" into the surrounding contexts and can be evaluated by code than the macro they're defined in. Can we add a test for macro hygiene? That is, if you have nested macros, can an inner one access the `$1`/`$2`/etc. from the outer one? (It probably shouldn't be able to, and I think it doesn't in this implementation.)

arxanas (approved):
> This only allows aliasing functions; I did not consider allowing aliases for names, though it would be easy to add.

That seems sensible. After all, the user can just call `foo()` with no arguments anyways.

> I made no attempt to include the defined aliases in the EvalError::UnboundFunction output. It wasn't immediately obvious to me how to list all config keys defined in the section, but I should admit that I didn't look very hard. 😄 Including them would certainly be kinder and would aid in discoverability.

That's probably fine for now. If we wanted to implement this, we would probably add variable/function lookup contexts and iterate through those. To list all config, I think we would want the [`Config::entries`](https://docs.rs/git2/0.13.17/git2/struct.Config.html#method.entries) function (which currently isn't wrapped by git-branchless's Git APIs).

> This has me wondering about git branchless shipping w/ some predefined aliases of it's own, which would be installed via init.

If we do, won't it be more or less equivalent to having defined them in the code?

> It's worth calling attention to the ... hacky? ... crude? ... expeditious? way that I implemented this: When an unrecognized fn is encountered, we check to see if a corresponding alias is defined. If so, we serialize each element of args (from Expr back into Strings) and replace each placeholder in the alias "template" w/ the corresponding serialized arg. Then we re-parse the alias (w/ the placeholders replaced) and send the result back through eval_inner.

To speak more accurately about this, you've implemented a user-defined *macro* system for revset aliases (as opposed to a user-defined *function* system). In this case, your macro system is entirely textual. 

> For the sake of discussion, I've pushed a fixup commit w/ an alternate approach. Instead of serializing and reparsing expressions

And this is a syntactic macro system, in which you find the AST node and you replace it with some other nodes.

As you've noticed, there are some downsides to the macro approach:

- The same expression might be evaluated more than once.
- The user might get confused because error messages might refer to macro-expanded code that they can't see.

> A more robust option might be to just parse the alias as is (w/ placeholders not replaced) but then pass some sort of "unbound arg resolver" callback through the eval system, and that callback could be used to resolve things like $n placeholders into parsed Exprs. I suspect that this might introduce some significant complexity into the eval system, though.

The other implementation approach would be to define lookup contexts for variables/functions, rather than using macros. This is how `gitrevset` does it: https://github.com/quark-zju/gitrevset/blob/50234203913331641aae1d479b5f861869f39cda/src/eval.rs#L21-L31

At startup, you would populate the contexts with the set of variables (initially none) and functions (the predefined ones and then add all aliases that the user has defined). For each user entry, you would need to analyze it and construct some kind of function object (currently we use `type FnType = &'static (dyn Fn(&mut Context, &str, &[Expr]) -> EvalResult + Sync)`, but that might not be suitable if we need to close over more state). During the evaluation of the function in question, you would add `$1`/`$2`/etc. to the variable lookup context based on the number of actual arguments, and you would want to make sure that these are *not* inherited by any functions called by the alias.

The advantages would be:

- We can prevent the evaluation of the arguments more than once.
- We might be able to produce better error messages since we can traverse the logical call stack and say something like

```
In expression `sole(foo())`
               ^^^^^^^^^^^
with `sole = exactly($1, 1)`
                     ^^
error: expected `$1` to evaluate to 1 element, but got 3
```

That would indeed be a lot of complexity and I don't think the advantages are that important here, so we can stick with the macro approach indefinitely (unless you want a challenge 😉).

git-branchless/src/revset/eval.rs
claytonrcarter: Thank you for this suggestion; the output is much more helpful now. Here's an example:
```
❯ git config branchless.revsets.alias.foo 'bar('
❯ cargo run -- query 'foo()'
Evaluation error for expression 'foo()': Failed to parse alias expression 'bar('
parse error: Unrecognized EOF found at 4
Expected one of "(", ")", "..", ":", "::", a commit/branch/tag or a string literal
```

Note that I used `eyre` to manually wrap and format the `ParseError` ... sorry to be dense about this, but I'm still feeling shaky about how to properly handle errors in some cases. Does this sound/look OK to you?

git-branchless/src/revset/ast.rs
claytonrcarter: OK, I've made an attempt at this, but would like your confirmation that I'm on the right path. The relevant test (and explanation) is at https://github.com/arxanas/git-branchless/pull/509/files#diff-ce578e640c3ce91a995bf4a577a8a4afcbdd9208a518df3b28998cc2c7d472e6R1161-R1201

claytonrcarter:
Thank you for the review. I've combined the previous fixup commit into the base commit, and I've added a new fixup commit w/ today's changes. If you don't mind, I'd like you to double check my work on a couple of pieces that I'm a little uncertain on. (I've noted those in other comments.)



> That seems sensible. After all, the user can just call `foo()` with no arguments anyways.

My thinking exactly. :wink:

> If we do, won't it be more or less equivalent to having defined them in the code?

Yup. I don't see any real upsides, I guess, aside from perhaps providing demonstration on aliases/macros.

> unless you want a challenge 

Not right now. :smile: I'm happy to use this implementation, and then change it if that becomes necessary or prudent.

I think I see what you're saying though: the text-macro system (as originally implemented here) leads to repeated parsing and evaluation of sets/arguments, the syntactic-macro system (currently implemented here) avoids repeated parsing but will still do repeated evaluation of sets, while the system in use by `gitrevset` parses *and* evaluates everything only once.

git-branchless/src/revset/eval.rs
arxanas: I don't think there's any need to use `Cow` here. Most of the places I use it are for practice and to document that in principle the type might borrow e.g. from the source text, to future-proof it a little.

arxanas: The approach re. syntactic macros looks good to me 👍.

arxanas (approved):

git-branchless/src/revset/eval.rs
arxanas: Ideally we'll create new error types for reach failure case. You can see the documentation for [`thiserror`] for how to do it. In this case, we could create `EvalError::ParseRevsetError(ParseError)` to wrap the underlying type directly, rather than create a new `eyre::Error`.

claytonrcarter: Ah, of course. Yes, that makes sense. Done! Thank you!

claytonrcarter:
I've added a basic write up about aliases to the revset docs at https://github.com/arxanas/git-branchless/wiki/Reference:-Revsets#Aliases

Issue #515, opened 2022-08-28
`git reword` on a public commit loses the message edits, wrong hint in error message


ilyagr:
### Description of the bug

1. Be on a public commit (main branch).
2. Do `git reword`.

### Expected behavior

You get an error message telling you to use the `--force-rewrite` option.

### Actual behavior

The editor opens, lets you edit the commit message. After saving the message, you get an error message and the edits are lost.

Moreover, the message tells you to use an `--force` option, which doesn't exist.

The exact error I got:

```
You are trying to rewrite 1 public commit, such as: 51298b4 Instead of c_restow, try c.restow
It is generally not advised to rewrite public commits, because your
collaborators will have difficulty merging your changes.
Retry with -f/--force to proceed anyways.
```

#### Software version

git-branchless 0.4.0 (v0.3.6-nixos.0-494-gedb6e91)

#### Operating system

Linux 5.10.120-16471-g167ad2e5dcba

#### Command-line

```bash
git-branchless_dev bug-report 
```

#### Environment variables

```bash
SHELL=/usr/bin/fish
EDITOR=vim
```

#### Git version

```
> git version 
git version 2.30.2
```

#### Hooks

<details>
<summary>Show 6 hooks</summary>

##### Hook `post-commit`

```
#!/bin/sh
## START BRANCHLESS CONFIG

git branchless hook-post-commit "$@"

## END BRANCHLESS CONFIG
```
##### Hook `post-merge`

```
#!/bin/sh
## START BRANCHLESS CONFIG

git branchless hook-post-merge "$@"

## END BRANCHLESS CONFIG
```
##### Hook `post-rewrite`

```
#!/bin/sh
## START BRANCHLESS CONFIG

git branchless hook-post-rewrite "$@"

## END BRANCHLESS CONFIG
```
##### Hook `post-checkout`

```
#!/bin/sh
## START BRANCHLESS CONFIG

git branchless hook-post-checkout "$@"

## END BRANCHLESS CONFIG
```
##### Hook `pre-auto-gc`

```
#!/bin/sh
## START BRANCHLESS CONFIG

git branchless hook-pre-auto-gc "$@"

## END BRANCHLESS CONFIG
```
##### Hook `reference-transaction`

```
#!/bin/sh
## START BRANCHLESS CONFIG

# Avoid canceling the reference transaction in the case that `branchless` fails
# for whatever reason.
git branchless hook-reference-transaction "$@" || (
echo 'branchless: Failed to process reference transaction!'
[...]

PR #518, opened 2022-08-29
Display correct flag name about rewriting public commits (#515)

claytonrcarter:
I think this fixes (one of) the issue raised in #515: the error message displayed when attempting to rewrite public commits was out of sync w/ the actual name of the flag.


arxanas (approved):
Thanks for fixing!

PR #526, opened 2022-09-09
Add support for interactively checking out a branch

bcspragu:
This is a follow-up from [the discussion here](https://github.com/arxanas/git-branchless/discussions/520#discussioncomment-3603617). It updates the `git co` command to check if the selected commit is associated with a single branch, and checks that out if so.

### DIFF
--- a/git-branchless-lib/src/core/check_out.rs
+++ b/git-branchless-lib/src/core/check_out.rs
@@ -7,8 +7,10 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use cursive::theme::BaseColor;
 use cursive::utils::markup::StyledString;
 use eyre::Context;
+use itertools::Itertools;
 use tracing::instrument;
 
+use crate::core::config::get_auto_switch_branches;
 use crate::git::{
     update_index, CategorizedReferenceName, GitRunInfo, MaybeZeroOid, NonZeroOid, ReferenceName,
     Repo, Stage, UpdateIndexCommand, WorkingCopySnapshot,
@@ -19,6 +21,7 @@ use super::config::get_undo_create_snapshots;
 use super::effects::Effects;
 use super::eventlog::{Event, EventLogDb, EventTransactionId};
 use super::formatting::printable_styled_string;
+use super::repo_ext::{RepoExt, RepoReferencesSnapshot};
 
 /// An entity to check out.
 #[derive(Clone, Debug)]
@@ -54,6 +57,40 @@ impl Default for CheckOutCommitOptions {
     }
 }
 
+fn maybe_get_branch_name(
+    current_target: Option<String>,
+    oid: Option<NonZeroOid>,
+    repo: &Repo,
+) -> eyre::Result<Option<String>> {
+    let RepoReferencesSnapshot {
+        head_oid,
+        branch_oid_to_names,
+        ..
+    } = repo.get_references_snapshot()?;
+    if (head_oid.is_some() && head_oid == oid) || current_target == head_oid.map(|o| o.to_string())
+    {
+        // Don't try to checkout the branch if we aren't actually checking anything new out.
+        return Ok(current_target);
+    }
+
+    // Determine if the oid corresponds to exactly a single branch. If so,
+    // check that out directly.
+    match oid {
+        Some(oid) => match branch_oid_to_names.get(&oid) {
+            Some(branch_names) => match branch_names.iter().exactly_one() {
+                Ok(branch_name) => {
+                    // To remove the `refs/heads/` prefix
+                    let name = CategorizedReferenceName::new(branch_name);
+                    Ok(Some(name.remove_prefix()?))
+                }
+                Err(_) => Ok(current_target),
+            },
+            None => Ok(current_target),
+        },
+        None => Ok(current_target),
+    }
+}
+
 /// Checks out the requested commit. If the operation succeeds, then displays
 /// the new smartlog. Otherwise displays a warning message.
 #[instrument]
@@ -71,20 +108,25 @@ pub fn check_out_commit(
         render_smartlog,
     } = options;
 
-    let target = match target {
-        None => None,
+    let (target, oid) = match target {
+        None => (None, None),
         Some(CheckoutTarget::Reference(reference_name)) => {
             let categorized_target = CategorizedReferenceName::new(&reference_name);
-            Some(categorized_target.remove_prefix()?)
+            (Some(categorized_target.remove_prefix()?), None)
         }
-        Some(CheckoutTarget::Oid(oid)) => Some(oid.to_string()),
-        Some(CheckoutTarget::Unknown(target)) => Some(target),
+        Some(CheckoutTarget::Oid(oid)) => (Some(oid.to_string()), Some(oid)),
+        Some(CheckoutTarget::Unknown(target)) => (Some(target), None),
     };
 
     if get_undo_create_snapshots(repo)? {
         create_snapshot(effects, git_run_info, repo, event_log_db, event_tx_id)?;
     }
 
+    let target = if get_auto_switch_branches(repo)? {
+        maybe_get_branch_name(target, oid, repo)?
+    } else {
+        target
+    };
     let args = {
         let mut args = vec![OsStr::new("checkout")];
         if let Some(target) = &target {
--- a/git-branchless/tests/command/test_navigation.rs
+++ b/git-branchless/tests/command/test_navigation.rs
@@ -600,7 +600,7 @@ fn test_checkout_pty_branch() -> eyre::Result<()> {
         let (stdout, _stderr) = git.run(&["smartlog"])?;
         insta::assert_snapshot!(stdout, @r###"
         :
-        @ 62fc20d (master) create test1.txt
+        @ 62fc20d (> master) create test1.txt
         |\
         | o 96d1c37 create test2.txt
         |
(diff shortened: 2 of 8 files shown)


arxanas:
Thanks for looking into this! Can we try instead implementing this logic in `core::check_out::check_out_commit`? That way, we can benefit from this behavior in other places too.

> My Rust is gross (lots of copying/mut) and any pointers are greatly

Generally, rather than write

```rust
let mut foo = 1;
if cond {
    foo = 2;
}
```

you would write

```rust
let foo = 1;
let foo = if cond { 2 } else { foo };
```

Similarly, to return multiple values from a conditional statement, you can return a tuple (or even a `struct` if it gets to be particularly complicated):

```rust
let (foo, bar) = if cond {
    (true, "hello")
} else {
    (false, "world")
}
```

> instead of checking out <branch name>, it checks out refs/heads/<branch name>.

I think `check_out_commit` has a fix for this anyways (where it calls `remove_prefix`).

You may need to update tests for this change. If necessary, to review snapshot tests interactively, you can install `cargo-insta` and run `cargo insta review`.

arxanas:
Also, I added a refactoring for the checkout target in https://github.com/arxanas/git-branchless/pull/514 which might help, so you may want to rebase your work on top of that.

bcspragu:
Thanks! Pulled in your changes, and moved the code into `core check_out`. Made some of the code less gross, but I'm sure there's plenty of room for improvement.

Currently, I'm just looking for all local branches then checking for an `oid` match, which may be expensive for large repos. If there's an cheaper way to say "give me all the branches with this oid", we should probably do that.

Looks like the tests are passing, but it seemed like lots of tests required updates. Some of them made sense, we're no longer checking out a detached commit, so the hash is replaced with the branch name, but others were more puzzling (e.g. why does `test_reword` need to change?)

git-branchless-lib/src/core/check_out.rs
```diff
@@ -71,20 +71,36 @@ pub fn check_out_commit(
...
-        Some(CheckoutTarget::Unknown(target)) => Some(target),
+        Some(CheckoutTarget::Oid(oid)) => (Some(oid.to_string()), Some(oid)),
+        Some(CheckoutTarget::Unknown(target)) => (Some(target), None),
     };
 
     if get_undo_create_snapshots(repo)? {
         create_snapshot(effects, git_run_info, repo, event_log_db, event_tx_id)?;
     }
 
+    let target = if let Some(oid) = oid {
```
arxanas, line 88: For consistency with the rest of the codebase, use `match oid { Some(oid) => { ... }, None => { ... } }`.

git-branchless-lib/src/core/check_out.rs
```diff
@@ -71,20 +71,36 @@ pub fn check_out_commit(
...
+        let all_branches = repo.get_all_local_branches()?;
```
arxanas, line 89: To avoid scanning the repo for branches again, we'll want to use the field in `ReferencesSnapshot` called something like `branch_oid_to_names`. Since it's not a parameter to this function, you'll want to refactor this function to pass a `ReferencesSnapshot` from each caller. Then it should also be straightforward to get the matching branches, by just indexing into that `HashMap`.

Also, `ReferencesSnapshot` should be valid for repo states in the past (for `git undo`), unlike the functions on `Repo` which only operate on the present repo state.

git-branchless-lib/src/core/check_out.rs
```diff
@@ -71,20 +71,36 @@ pub fn check_out_commit(
...
+        let matching_branches: Vec<&Branch> = all_branches
+            .iter()
+            .map(|b| (b, b.get_oid().unwrap().unwrap()))
+            .filter(|t| t.1 == oid)
+            .map(|t| t.0)
+            .collect();
```
arxanas, line 95: We'll use `branch_oid_to_names` instead, but for future reference, you'd want to write this pattern `.filter_map(|(branch, branch_oid)| if branch_oid == oid { Some(branch) } else { None })`

git-branchless-lib/src/core/check_out.rs
```diff
@@ -71,20 +71,36 @@ pub fn check_out_commit(
...
+        match matching_branches.len() {
+            1 => Some(String::from(matching_branches.get(0).unwrap().get_name()?)),
+            _ => target,
+        }
```
arxanas, line 99: You can do a cool pattern-match to avoid the `unwrap` here, something like this:

```suggestion
        match matching_branches.as_slice() {
            &[branch] => Some(branch.get_name()?.to_owned()),
            _ => target,
        }
```

For consistency with the rest of the codebase, use `to_owned()` here to convert the `&str` into a `String`.

git-branchless/tests/command/test_reword.rs
```diff
@@ -49,7 +49,7 @@ fn test_reword_current_commit_not_head() -> eyre::Result<()> {
     let (stdout, _stderr) = git.run(&["smartlog"])?;
     insta::assert_snapshot!(stdout, @r###"
     :
-    @ 62fc20d (test1) create test1.txt
+    @ 62fc20d (> test1) create test1.txt
```
arxanas, line 52: I think this is happening because `git reword` (and other rebasing-related functions?) happens to implicitly check out the current commit after it's done. Maybe we can update `check_out_commit` to only check out the branch if the target isn't the current commit? Or even skip the rest of the function body altogether?

git-branchless/tests/command/test_navigation.rs
```diff
@@ -600,7 +600,7 @@ fn test_checkout_pty_branch() -> eyre::Result<()> {
         let (stdout, _stderr) = git.run(&["smartlog"])?;
         insta::assert_snapshot!(stdout, @r###"
         :
-        @ 62fc20d (master) create test1.txt
+        @ 62fc20d (> master) create test1.txt
```
arxanas, line 603: I was also hoping that `git next` and `git prev` would automatically check out the relevant branch when moving onto a certain commit. Do you know why they're not doing that? Maybe there's actually just no test for a `git next`/`git prev` which moves onto a commit with a branch pointing to it?

git-branchless-lib/src/core/check_out.rs
bcspragu: Looks like I can get `get_branch_oid_to_names` on the existing repo parameter directly by importing `RepoExt`, which I've done here. Let me know if that's the wrong approach

git-branchless-lib/src/core/check_out.rs
bcspragu: Good to know, thanks!

git-branchless-lib/src/core/check_out.rs
bcspragu: This code has changed a bit. It's replacement also uses unwrap, but I think it's far less unwieldy this time

git-branchless/tests/command/test_navigation.rs
bcspragu: They do automatically work! Looks like there are just no tests covering that case (I can add some if that's useful)

Here's an asciinema of it in action: 
[asciicast](https://asciinema.org/a/M1emyaWr05JWRyK0gH5HrRIPG)

git-branchless-lib/src/core/check_out.rs
```diff
@@ -54,6 +55,40 @@ impl Default for CheckOutCommitOptions {
...
+        Some(oid) => match branch_oid_to_names.get(&oid) {
+            Some(branch_names) => match branch_names.len() {
+                1 => {
+                    // To remove the `refs/heads/` prefix
+                    let name = CategorizedReferenceName::new(branch_names.iter().next().unwrap());
+                    Ok(Some(name.remove_prefix()?))
+                }
+                _ => Ok(current_target),
+            },
+            None => Ok(current_target),
```
arxanas, line 86:
```suggestion
            Some(branch_names) => match branch_names.iter().exactly_one() {
                Ok(branch_name) => {
                    // To remove the `refs/heads/` prefix
                    let name = CategorizedReferenceName::new(branch_name);
                    Ok(Some(name.remove_prefix()?))
                }
                Err(_) => Ok(current_target),
            },
```

You can use the `itertools` library's `exactly_one` function here to make it a little safer.

git-branchless/tests/command/test_reword.rs
arxanas: Oh, I guess that this is happening because the commit *is* different after the reword, i.e. it has a different commit hash. I guess we could work around this by looking at the old and new commits and seeing if the new version of the commit is a rewrite of the old one, and, if so, inhibiting the auto-check-out behavior.

That being said, I think it's a lot of complexity, and I don't envision users needing to do that. I expect that either they'll want to always have the one branch associated with the current commit checked out (and can use e.g. `git record --detach` to create a commit without affecting the current branch), or they would like the previous behavior where we just don't adjust the current branch.

I think the rest of this PR is good to ship. Could you add a configuration option to enable this behavior (we can enable it by default) and a test to confirm that the behavior doesn't happen when the behavior is disabled?

- Update the Wiki here: https://github.com/arxanas/git-branchless/wiki/Configuration
- Add the configuration option to this file: https://github.com/arxanas/git-branchless/blob/605d5dc641687cd87dfa41716ad8a005fc5aa9b6/git-branchless-lib/src/core/config.rs

I guess for the configuration variable name, we could call it something like `branchless.navigation.autoSwitchBranches`, or maybe you have a better idea? (Adopting the new Git terminology of "switch" for branches, since "check out" is overloaded.)

(There's also the existing configuration variable `branchless.next.interactive`, but I think that is a misnomer, and we should eventually move it to this new `branchless.navigation` section, because it also affects the `git prev` command.)

git-branchless-lib/src/core/check_out.rs
bcspragu: Good call, done

git-branchless/tests/command/test_reword.rs
bcspragu: Added the `branchless.navigation.autoSwitchBranches` configuration option, updated the documentation, and added two tests, one with the functionality turned on, and the other with it turned off.

bcspragu:
Not sure why the tests are failing, they work fine locally and produce no diffs in need of review. Given the failing output, looks like the feature might be broken in CI, I'll take a look at this later today

arxanas:
My bet is that you just need to apply this patch:

```diff
diff --git a/git-branchless/tests/command/test_navigation.rs b/git-branchless/tests/command/test_navigation.rs
index 3aa0d90..5eb0153 100644
--- a/git-branchless/tests/command/test_navigation.rs
+++ b/git-branchless/tests/command/test_navigation.rs
@@ -859,6 +859,7 @@ fn test_checkout_auto_switch_interactive() -> eyre::Result<()> {
             PtyAction::WaitUntilContains("> "),
             PtyAction::Write("test1"),
             PtyAction::WaitUntilContains("> test1"),
+            PtyAction::WaitUntilContains("> 62fc20d"),
             PtyAction::Write(CARRIAGE_RETURN),
         ],
     )?;
@@ -902,6 +903,7 @@ fn test_checkout_auto_switch_interactive_disabled() -> eyre::Result<()> {
             PtyAction::WaitUntilContains("> "),
             PtyAction::Write("test1"),
             PtyAction::WaitUntilContains("> test1"),
+            PtyAction::WaitUntilContains("> 62fc20d"),
             PtyAction::Write(CARRIAGE_RETURN),
         ],
     )?;
```

The failure case on CI (or a slow machine in general) is that the process can exit before the UI has a chance to select the given item, which means that nothing will happen when you exit the process because you haven't selected a switch target. It's not clear at all from just the test case, but there are *two* carets of interest in the UI. Example screenshot:

Screen Shot 2022-09-29 at 11 05 59

One of them is the prompt where you're typing, and the other is the actual selection. You need to wait for the actual selection to complete before exiting. (I guess it would be possible to not have the wait for `> test1` at all; I don't remember if there's a reason that the other test cases have that.)

arxanas:
Other than the failing tests, everything looks good to me. When you're done, please squash and force-push your changes and I'll merge this PR. Thanks for working on this!

arxanas:
We have some failing Nix tests already for PTY-related stuff, so I just added the new failing tests introduced in this PR to the list of skipped tests. @bcspragu thanks for your contribution!

PR #533, opened 2022-09-12
Support rewording merge commits

arxanas:
Also addresses #515.


git-branchless-lib/src/core/rewrite/execute.rs
```diff
@@ -585,18 +594,109 @@ mod in_memory {
...
+                        for parent in commits_to_merge {
+                            let parent = match parent {
+                                OidOrLabel::Oid(oid) => repo.find_commit_or_fail(*oid)?,
+                                OidOrLabel::Label(label) => {
+                                    let reference = repo
+                                        .find_reference(&ReferenceName::from(label.clone()))?
+                                        .ok_or_else(|| {
+                                            eyre::eyre!("Could not find label {label}")
+                                        })?;
+                                    reference.peel_to_commit()?.ok_or_else(|| {
```
arxanas, line 660: This should probably be a lookup in `labels` instead.

claytonrcarter:
This looks good to me, FWIW. I like the clean up in `reword.rs`!

PR #538, opened 2022-09-13
feat(reword): Add `--fixup` flag to `git branchless reword`

claytonrcarter:
This adds a `--fixup` option to `git reword`, providing a convenient way to turn existing commits into `fixup!` commit. With this, instead of doing something like `git reword <commits to reword> -m 'fixup! <paste title of commit to fixup>'`, you can now do `git reword <commits to reword> --fixup <commit to fixup>` and `reword` will handle the rest, including some error checking.

I've been using this for a few weeks and it's been working well. My only remaining concern about this is that the CLI option docs and error messages feel a little clunky because it's hard to talk to about "the commit being fixed up" vs "the commits that will become `fixup!` commits". As usual, suggestions are welcome!

### DIFF
--- a/git-branchless/src/commands/reword.rs
+++ b/git-branchless/src/commands/reword.rs
@@ -119,7 +119,7 @@ pub fn reword(
                     writeln!(
                         effects.get_error_stream(),
                         "--fixup expects exactly 1 commit, but '{}' evaluated to {}.\nAborting.",
-                        revset.0,
+                        revset,
                         commits.len()
                     )?;
                     return Ok(ExitCode(1));
--- a/git-branchless/src/commands/reword.rs
+++ b/git-branchless/src/commands/reword.rs
@@ -11,7 +11,7 @@ use std::fs::File;
 use std::io::Write as OtherWrite;
 use std::time::SystemTime;
 
-use bstr::ByteSlice;
+use bstr::{ByteSlice, ByteVec};
 use chrono::Local;
 use dialoguer::Editor;
 use eden_dag::DagAlgorithm;
@@ -40,6 +40,9 @@ use crate::revset::resolve_commits;
 pub enum InitialCommitMessages {
     /// The user wants to start with an empty (or template) message.
     Discard,
+
+    /// The user wants to fixup a commit.
+    FixUp(Revset),
 
     /// The user provided explicit messages.
     Messages(Vec<String>),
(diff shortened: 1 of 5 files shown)


git-branchless/src/commands/reword.rs
```diff
@@ -90,6 +93,43 @@ pub fn reword(
...
+            let commits_to_fixup =
+                resolve_commits_from_hashes(&repo, &mut dag, effects, vec![revset.clone()])?
+                    .unwrap_or_default();
+            let commit_to_fixup = match commits_to_fixup.as_slice() {
+                [commit_to_fixup] => {
+                    if commits.iter().any(|commit| {
+                        !dag.query()
+                            .is_ancestor(commit_to_fixup.get_oid().into(), commit.get_oid().into())
+                            .expect("Failed checking ancestry of commit.")
+                    }) {
```
claytonrcarter, line 108: I suppose that instead of `iter().any(...)` I could instead just use a single dag query; perhaps like
```rust
commits.intersection(
    dag.query.descendents(
        commit_to_fixup.get_oid().into()
    )
) == commits
```

git-branchless/src/opts.rs
```diff
@@ -399,8 +399,6 @@ pub enum Command {
     /// Reword commits.
     Reword {
         /// Zero or more commits to reword. If not provided, defaults to "HEAD".
-        ///
-        /// Can either be hashes, like `abc123`, or ref-specs, like `HEAD^`.
```
claytonrcarter, line 403: This is related, but I removed this b/c we don't include a similar explanation in, eg, `git move`, and I also think it's outdated. (eg, it doesn't mention revsets)

git-branchless/src/commands/reword.rs
claytonrcarter: Nevermind this; I thought that `NameSet` allowed for `==/!=` but it doesn't. I'll leave it as is unless you have a suggestion for a more intuitive way to check this.

arxanas: I think you could write this as `dag.query().common_ancestors(CommitSet::from(commits))?.contains(commit_to_fixup)?` or something like that. I don't know why there isn't a `.equals()` method or similar.

git-branchless/src/commands/reword.rs
```diff
@@ -90,6 +93,43 @@ pub fn reword(
...
+                        )?;
+                        return Ok(ExitCode(1));
+                    }
+                    commit_to_fixup
+                }
+                commits => {
+                    writeln!(
+                        effects.get_error_stream(),
+                        "--fixup expects exactly 1 commit, but '{}' evaluated to {}.\nAborting.",
+                        revset.0,
```
arxanas, line 121: Feel free to implement `Display` for `Revset` instead of accessing `.0`.

git-branchless/src/opts.rs
```diff
@@ -420,6 +418,11 @@ pub enum Command {
         /// that; otherwise, the editor starts empty.
         #[clap(action, short = 'd', long = "discard", conflicts_with("messages"))]
         discard: bool,
+
+        /// A commit to "fix up". The reworded commits will become `fixup!` commits (suitable for
+        /// use with `git rebase --autosquash`) targetting the supplied commit.
```
arxanas, line 423:
```suggestion
        /// A commit to "fix up". The reworded commits will become `fixup!` commits (suitable for
        /// use with `git rebase --autosquash`) targeting the supplied commit.
```

I think this message is pretty much fine as-is; either you're familiar with `--autosquash` and you know what this option does, or you'll need to read more about it anyways.

arxanas (approved):
I'm okay with merging this as-is for now, but maybe we should invent a more straightforward method to fixup commits? For example, it would be more direct if you could write `git amend <target> --from <revs>` or something like that. Now that we can move exact commits, it should be possible to move all of the `<revs>` right on top of `<target>`, and then create a `RebaseCommand::Fixup` to handle the fixing up. The implementation for in-memory rebases doesn't seem too bad: you would basically do a `pick` but call `amend_commit` instead of `cherry_pick_fast`.

arxanas:
I guess a theoretical fixup command should be part of `git move`; `git amend` should probably be for working copy changes only.

git-branchless/src/commands/reword.rs
claytonrcarter: > I think you could write this as `dag.query().common_ancestors(...`

Oh, yes ... much nicer, thank you!

> Feel free to implement `Display` for `Revset`

Done!

PR #539, opened 2022-09-13
fix(revset): Fix ParseError for single quotes in complex nested expresssions

claytonrcarter:
This includes 2 trivial changes related to revsets: 
- a nitpick on an error message missing a single character
- a minor fix to the revset grammar that include single quoted strings

I ran into an issue trying to use the revset `message('wip') - roots(message('wip'))`. This look valid to me, but triggers a `ParseError` for unexpected `)`. Interestingly enough, many simpler, related revsets seem to work just fine:
- `message('wip')`
- `message('wip') - message('wip')`
- `roots(message('wip'))`
... all work.

But adding the `-` (or any bitwise op, FWIW) was causing the error. I have not checked if this breaks when using the long form of `intersection(...)` etc.

Anyway, I looked at the lalrpop grammar and noticed some asymmetry in the regex for parsing names. This change allows the example revset to parse and evaluate just fine. I've included a test for this edge case, which confirms that it parses. That test fails when run against `master`.

Thanks!

### DIFF
--- a/git-branchless/src/revset/grammar.lalrpop
+++ b/git-branchless/src/revset/grammar.lalrpop
@@ -63,8 +63,8 @@ Name: Cow<'input, str> = {
     // \x22 = double-quote.
     // \x27 = single-quote.
     // \x5c = backslash
-    <s:r"\x22([^\x22]|\x5c.)*\x22|\x27([^\x22]|\x5c.)*\x27"> => {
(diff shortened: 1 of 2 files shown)


arxanas (approved):
Nice catch, thanks! In principle, we can write our own tokenizer for LALRPOP that doesn't use regexes, which might have avoided this situation, but it's a bit of a pain (and our current tokenization scheme *can* be implemented with a regex).

PR #546, opened 2022-09-18
fix(submit): Prefer remote.pushDefault

claytonrcarter:
Hi there. I'm just opening this for discussion's sake. I tried out the sweet new `git submit` command for #545 and found that either I'm misunderstanding how to use it, or it might not be set up correctly to handle the "triangle" or forked remote workflow, as it pushed my branch directly to `origin` (this repo) instead of my fork as I had intended.

In my case, `master` branch is tracking `origin/master`, but I've got `remote.pushDefault` set to `claytonrcarter`, which is my fork. As I understand the current code, `git submit` will prefer whatever remote is being tracked by the main branch, in my case `origin`, and will only use `pushDefault` if there is no main tracking branch. Unless I'm misreading this, this seems like it might make `pushDefault` useful only before a remote is added ... maybe?

Anyway, I'm suggesting that we prefer `pushDefault` over whatever the main branch is using. I tried this out to push this change, and it worked as expected.

If this is not correct, please feel free to close and, if you don't mind, give me a little hint about how/why I'm misusing and/or misunderstanding this. Thanks!

### DIFF
--- a/git-branchless/src/commands/submit.rs
+++ b/git-branchless/src/commands/submit.rs
@@ -181,6 +181,10 @@ To create and push them, retry this operation with the --create option."
 }
 
 fn get_default_remote(repo: &Repo) -> eyre::Result<Option<String>> {
+    if let Some(push_default_remote) = repo.get_readonly_config()?.get("remote.pushDefault")? {
+        return Ok(Some(push_default_remote));
+    }
+
     let main_branch_reference = repo.get_main_branch_reference()?;
     let main_branch_name = main_branch_reference.get_name()?;
     match CategorizedReferenceName::new(&main_branch_name) {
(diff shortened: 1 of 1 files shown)


arxanas:
The intention was to only use `remote.pushDefault` for branches which don't track any branch yet, and set `branch.<name>.pushRemote` for branches which need to pull and push from different places. So for your main branch, you should set the push-remote to create the "triangle". See https://git-scm.com/docs/git-config#Documentation/git-config.txt-branchltnamegtpushRemote. But I don't remember if I actually implemented `pushRemote` in `git submit` or not, so that might remain to be done.

claytonrcarter:
> The intention was to only use remote.pushDefault for branches which don't track any branch yet, and set branch.<name>.pushRemote for branches which need to pull and push from different places

Thank you, that is what I was missing ... and in fact I failed to read all the way through. It looks like `branch.{branch_name}.pushRemote` is implemented already, and I was just ignorant of it.

Thank you and sorry for the distraction!

PR #561, opened 2022-09-23
Fix lint error

pokey:
I believe this is what was intended.  I think the linter caught a genuine error here, because I believe `0` actually would refer to `commit` here.  At least that's what the linter suggests would be equivalent.  Guessing maybe the map gets sorted alphabetically, so then `0` would refer to `commit`?

Looks like there's no tests for this error message, though, as the tests passed whether I put `{commit}` or `{source}` here


arxanas:
Thanks! Not sure why this wasn't caught in CI previously.

pokey:
I think it may be caused by a recent rust upgrade; I wasn't getting it locally either and then I upgraded rust and started seeing it locally as well

PR #571, opened 2022-09-27
feat(revset): Add `rewritten` revset function

claytonrcarter:
This is my first take on a `rewritten()` revset that resolves possibly-outdated commit hashes into their most current version. First discussed here https://github.com/arxanas/git-branchless/discussions/48#discussioncomment-3718889

The primary use case is being able to do things like `git move -x abc123 -d deb456 && git reword 'rewritten(abc123)'`, ie to avoid having to lookup the new SHA of the moved commit. If I understand the docs of `find_rewrite_target()` correctly, this should be able to work across many rebase/move/reword/restack/etc operations, so long as the specified SHA hasn't been gc'd.

I have only exercised this a little bit, so I'm not 100% certain of the implementation. I also note that there's *a lot* of `.wrap_err("...").map_err(EvalError::OtherError)` going on in here. They seemed like they should maybe "other errors", but I'd be happy to change those if another error type makes more sense.

### DIFF
--- a/git-branchless/src/revset/builtins.rs
+++ b/git-branchless/src/revset/builtins.rs
@@ -1,7 +1,9 @@
 use bstr::ByteSlice;
 use eden_dag::DagAlgorithm;
 use lib::core::dag::CommitSet;
-use lib::git::{Commit, Repo};
+use lib::core::eventlog::{EventLogDb, EventReplayer};
+use lib::core::rewrite::find_rewrite_target;
+use lib::git::{Commit, MaybeZeroOid, NonZeroOid, Repo};
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::convert::TryFrom;
@@ -50,6 +52,7 @@ lazy_static! {
             ("committer.email", &fn_committer_email),
             ("committer.date", &fn_committer_date),
             ("exactly", &fn_exactly),
+            ("current", &fn_current),
         ];
         functions.iter().cloned().collect()
     };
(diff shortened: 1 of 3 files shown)


git-branchless/src/revset/builtins.rs
```diff
@@ -398,3 +401,46 @@ fn fn_exactly(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+                MaybeZeroOid::NonZero(new_commit_oid) => {
+                    // commit rewritten as new_commit_oid
+                    result.push(new_commit_oid);
+                }
+                MaybeZeroOid::Zero => {
+                    // commit deleted, skip
+                }
+            },
+            // commit not rewritten
+            None => result.push(oid),
```
arxanas, line 442: Is it accurate to call this function "rewritten" if it sometimes returns versions of the commits which haven't been rewritten?

Mercurial calls the concept itself `successors`:

> `successors(set)`
>     All successors for set, including the given set themselves

But I'm not really sure when you would want to get the successors of a set but also the set itself.

For most tasks, I think you would want either:

* For each commit, if it's been rewritten, produce the latest rewritten version, or else drop the the commit.
* For each commit, if it's been rewritten, produce the latest rewritten version, or else produce the commit itself (which is what you have here).

Maybe we should call this function `latest` instead?

arxanas (approved):
> If I understand the docs of find_rewrite_target() correctly, this should be able to work across many rebase/move/reword/restack/etc operations, so long as the specified SHA hasn't been gc'd.

Yep, that sounds right.

> I also note that there's a lot of .wrap_err("...").map_err(EvalError::OtherError) going on in here. They seemed like they should maybe "other errors"

They seem fine as they are. `OtherError` is being used for generic `eyre::Error`s which don't have a more specific error type. There actually are some cases in the existing code in this file where we have an `eden_dag::Error` but we wrap it in an `eyre::Error` anyways to get a better stack trace.

git-branchless/src/revset/builtins.rs
claytonrcarter: Ah, yes, thanks for pointing this out. I hadn't considered it, but I agree that `rewritten` is not accurate and could me misleading. I agree it should be changed, but I'm not *quite* set on `latest`, which feels like it could also mean "the commit in this set that was most recently created/commited/etc" as well as "latest version of the commits in this set". 

How would you feel about `current` instead of `latest`?

arxanas: Sure, `current` sounds fine to me!

claytonrcarter: Great, thank you!

PR #576, opened 2022-10-02
Create `git test` command

arxanas:
Like this one: https://github.com/spotify/git-test

Ideally, we'll be able to use this command to be able to run a linter/formatter and apply changes to an entire stack of commits.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1193,6 +1193,8 @@ dependencies = [
  "rayon",
  "regex",
  "rusqlite",
+ "serde",
+ "serde_json",
  "skim",
  "thiserror",
  "tracing",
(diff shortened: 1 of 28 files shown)


claytonrcarter:
This is probably only relevant if/when this command gets support for committing changes to each commit, such as when trying to run a formatter on a stack. In that case, it might be nice to test/format each commit in reverse order. 

From the few times I've had to reformat a stack of commits, I've found it much easier to start w/ the head/most recent commits and move backwards. Doing so oldest->newest generated the worst possible merge conflicts for every commit. Based on what I read hear, though, it looks like some sort of `--reverse` flag should be easy enough to add and implement, and I'd be happy to help out and push such an update, if desired.

arxanas:
> In that case, it might be nice to test/format each commit in reverse order.

This is what I noticed too when applying formatting changes manually, but for `git test`, I think we can do better. We just have to replace each commit's tree with the result of applying the formatter/linter, without replacing the children commits/attempting to rebase them. We should actually be able to run all the commits in parallel with that strategy.

PR #618, opened 2022-11-01
build: run `cargo update`

arxanas:


arxanas:
@hmenke Any idea what the problem is here? It looks like `zlib.h` is not found at build-time.

arxanas:
This isn't blocking anything in particular, it would just be good to resolve for the next time we want to run `cargo update`.

hmenke:
Looks like a regression in the `pkg-config` crate, which received an upgrade in this PR.

https://github.com/arxanas/git-branchless/actions/runs/3370883274/jobs/5592357455#step:4:2255

arxanas:
@hmenke Thanks. I tried locking the version of `pkg-config` for now; we'll see if that works.

arxanas:
Hmm, I can see that the version of `pkg-config` is fixed to be the same as that of the parent commit, but the issue is still occurring. `pkg-config` itself doesn't seem to have any (explicit) dependencies.

PR #620, opened 2022-11-05
refactor(dag): unify definition of commit visibility

arxanas:


git-branchless-lib/src/core/dag.rs
```diff
@@ -362,36 +372,61 @@ impl Dag {
...
+        self.draft_commits.get_or_try_init(|| {
+            let visible_commits = self.query_visible_commits()?;
+            let public_commits = self.query_public_commits()?;
+            Ok(visible_commits.difference(public_commits))
+        })
+    }
+
+    /// Determine the default set of commits that is shown in the smartlog when
+    /// no revset is passed.
+    pub fn query_default_smartlog_commits(&self) -> eyre::Result<&CommitSet> {
```
claytonrcarter, line 416: This is minor, but maybe clarify this comment: passed to what? As I read it, this is not the set used when no revset is passed to `git smartlog` by the user. It looks like it's used for smartlogs rendered for undo, nav, etc event; so maybe it's more like an internal or system default?

Otherwise this PR looks good to me. Love the improved consistency!

arxanas: It's basically codifying the system default, since we automatically print the smartlog in several places. Actually, I like the default-from-config/specified system you were working on in one of your commits. We can switch to using that and delete this function then.

PR #626, opened 2022-11-07
feat(switch): change `--branch` to `--create`

mlcui-google:
`git branch` uses `--branch` to create a new branch, but `git switch` uses `--create`. This PR changes `git-branchless switch` to match `git switch`.

### DIFF
--- a/git-branchless/src/opts.rs
+++ b/git-branchless/src/opts.rs
@@ -138,7 +138,7 @@ pub struct SwitchOptions {
 
     /// When checking out the target commit, also create a branch with the
     /// provided name pointing to that commit.
-    #[clap(value_parser, short = 'b', long = "branch")]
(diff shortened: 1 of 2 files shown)


mlcui-google:
By the way, feel free to close this if users are used to `-b` / `--branch`! `git switch` has quite a different interface to `git checkout`, and I feel like it's probably best to match what `git switch` does instead of `git checkout` for `git-branchless switch`.

arxanas:
@mlcui-google thanks for the fix! Originally, this command was `git branchless checkout`, so it matched the interface of `git checkout`. Now it's better to match `git switch`, like you noticed.

It's slightly unfortunately because `git branchless switch` also lets you switch to commits, not just branches, so the flag `--create` becomes ambiguous (is it creating a commit or a branch?). But I doubt anyone will be confused.

PR #636, opened 2022-11-10
feat(smartlog): show number of omitted commits in sparse smartlog

arxanas:


arxanas:
@claytonrcarter this is what it looks like, thoughts?

Screen Shot 2022-11-09 at 21 04 08

It uses a bit of vertical space. I tried annotating the vertical ellipses themselves, rather than inserting virtual nodes, but it seemed harder to read to me.

claytonrcarter:
Yes, I think this makes sense and I really like your implementation. I can definitely see myself wanting to add a config option to disable these or something, but that can wait for another day. 😄

claytonrcarter (approved):

Issue #308, opened 2022-03-24
Populate the DAG on `git branchless init`


arxanas:
If you run `git branchless init` and then `git smartlog` in a big repository like https://github.com/mozilla/gecko-dev, it takes ~20 seconds to build the initial internal DAG. It would be best to do this during the `init` step, since that's the time you would expect any heavy-lifting operations to be carried out.

We just need to add a call to sync the DAG in the `init` function here:

https://github.com/arxanas/git-branchless/blob/45ace45579804697f8daeb551750424bf8e8538b/src/commands/init.rs#L550-L556

The call is to `Dag::open_and_sync`, and looks like this:

https://github.com/arxanas/git-branchless/blob/45ace45579804697f8daeb551750424bf8e8538b/src/commands/restack.rs#L242-L251

To test this:

1. Clone [`mozilla/gecko-dev`](https://github.com/mozilla/gecko-dev) locally.
2. Delete any existing DAG like this: `rm -r /path/to/gecko-dev/.git/branchless/dag/`
3. Run `cargo run -- -C /path/to/gecko-dev init` and confirm that it produces a progress meter like this:

```
Updating commit graph (5.1s)
```

and that it terminates after a few seconds.

It would also be cool if you could make the message say "Initializing commit graph" instead of "Updating commit graph" if this is the first time the DAG is being initialized, but it's not important. (That would be a more involved change.)

PR #667, opened 2022-11-22
Populate DAG during initialization

melko:
This addresses #308 calling the open_and_sync function at initialization.

### DIFF
--- a/git-branchless/src/commands/init.rs
+++ b/git-branchless/src/commands/init.rs
@@ -11,7 +11,10 @@ use tracing::{instrument, warn};
 
 use crate::opts::write_man_pages;
 use lib::core::config::{get_core_hooks_path, get_default_branch_name};
+use lib::core::dag::Dag;
 use lib::core::effects::Effects;
+use lib::core::eventlog::{EventLogDb, EventReplayer};
+use lib::core::repo_ext::RepoExt;
 use lib::git::{BranchType, Config, ConfigRead, ConfigWrite, GitRunInfo, GitVersion, Repo};
 
 pub const ALL_HOOKS: &[(&str, &str)] = &[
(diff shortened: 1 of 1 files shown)


arxanas:
Thanks! For some reason, the CI jobs didn't run. I pushed another commit to run `cargo fmt` and re-run the CI jobs. If it works out, then I'll squash the commits together and merge.

arxanas (changes requested):
It looks like a test is failing because it can't find the main branch. Perhaps, because it was just configured above, the appropriate configuration is not yet reflected in the current `repo` object? If so, you could reinitialize it from the repository and use that to sync the DAG.

melko:
I think the problem is that in a fresh repo the master branch is created at the first commit, so it does not exist yet when `git branchless init` is run in that test.
You can get the same error by doing something like this:
```
git init
git branchless init
git sl
```

melko:
A possible solution could be to not forward the Result<> from the `open_and_sync` called inside `init`,  so that if it fails the behavior should be the old one, something like this:
(not sure if a closure is the most ideomatic way to handle this, let me know)
```rust
    let _ = || -> eyre::Result<()> {
        let conn = repo.get_db_conn()?;
        let event_log_db = EventLogDb::new(&conn)?;
        let references_snapshot = repo.get_references_snapshot()?;
        let event_replayer = EventReplayer::from_event_log_db(effects, &repo, &event_log_db)?;
        let event_cursor = event_replayer.make_default_cursor();
        Dag::open_and_sync(
            effects,
            &repo,
            &event_replayer,
            event_cursor,
            &references_snapshot)?;
        Ok(())
    }();
```
what do you think about this?

arxanas:
> I think the problem is that in a fresh repo the master branch is created at the first commit, so it does not exist yet when git branchless init is run in that test.

Actually, I think in this case it's fine to change the requirements for `git branchless init` and mandate that the main branch be created before invoking. Then we can just update the test to match. I wasn't able to push my change directly to your branch, so instead I opened https://github.com/arxanas/git-branchless/pull/686. Thanks for investigating this!

melko:
Agreed, requiring the main branch to exist seems reasonable.
I fear you couldn't push it here because in the meanwhile I amended my PR to try the other solution, sorry.

I guess this PR can be closed then?

arxanas:
@melko it turns out there are a few other cases which don't work well, as per the failing tests in PR #686. For example, `git record` might plausibly be used to *create* the first commit on the main branch (and therefore the main branch itself), so we don't want to require that the main branch exist before then. I ended up returning to your approach, but instead using `if let Ok(_) = ... {}` to handle the possible error case.

arxanas:
Merged https://github.com/arxanas/git-branchless/pull/686, thanks again 😄.

PR #674, opened 2022-11-24
tests: explicitly set an initial git branch during init

mlcui-google:
Some tests assume a "master" branch exists during tests, which is assumed to be created on the git repo initialization. This assumption might be incorrect - from the `git-init` man page:

> the default [branch] name [is] currently `master`, but this is subject
> to change in the future; the name can be customized via the
> `init.defaultBranch` configuration variable

We cannot rely on the default branch name as it may be different - `init.defaultBranch` could be set in the global git config to something other than "master", or it could change in a git version increase.

This commit explicitly sets the initial branch so tests relying on a "master" branch will work, even if the default branch name changes or `init.defaultBranch` is set to something different.

### DIFF
--- a/git-branchless-lib/src/testing.rs
+++ b/git-branchless-lib/src/testing.rs
@@ -292,7 +292,10 @@ stderr:
     /// with it.
     #[instrument]
     pub fn init_repo_with_options(&self, options: &GitInitOptions) -> eyre::Result<()> {
-        self.run(&["init"])?;
+        // Ensure that the branch "master" is created for tests, regardless of
+        // the `init.defaultBranch` configuration variable, or git defaults.
+        self.run(&["init", "--initial-branch=master"])?;
+
         self.run(&["config", "user.name", DUMMY_NAME])?;
         self.run(&["config", "user.email", DUMMY_EMAIL])?;
 


mlcui-google:
Two questions:

- Is the comment necessary? I feel like we could remove it, but it might be worthwhile explaining why we explicitly create a "master" branch.
- How should I update this PR? I believe amending commits on GitHub results in a not-so-great review experience, so my assumption is "create more commits on the branch" instead of "amend the commit".

mlcui-google:
Hm, some old versions of git might not support this argument - see https://github.com/arxanas/git-branchless/actions/runs/3537660920/jobs/5937840248.

We could possibly take inspiration from `test_dont_install_existing_aliases` and set a fake `HOME` and global git config for all tests, which has `init.defaultBranch` set to "master" (so older Git versions that don't support `init.defaultBranch` will also use "master", as they'd be too old for the new default branch name). This would effectively make my three test-related PRs (#675, this one and https://github.com/mlcui-google/git-branchless/commit/8912dd60da3abba99c1e2fb542f5509eb25d1f62, which is a follow-up to this one) redundant.

Let me do that instead and close these PRs - I feel like that'd be much cleaner.

PR #675, opened 2022-11-24
tests: ignore global config in tests

mlcui-google:
Global configs such as `merge.conflictStyle = diff3` interfere with tests. This uses `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` introduced in Git 2.32, but keeps the prior use of `GIT_CONFIG_NOGLOBAL` as older git
versions might not support these new environment variables. See https://stackoverflow.com/a/67512433 for more details.

While the documentation for this feature both mention setting the environment variables to "/dev/null" to ignore the system and global configs, this is not portable as Windows does not have a /dev/null file. An empty string works the same way and should be portable.

### DIFF
--- a/git-branchless-lib/src/testing.rs
+++ b/git-branchless-lib/src/testing.rs
@@ -166,6 +166,8 @@ impl Git {
 
         let new_path = self.get_path_for_env();
         let envs = vec![
+            ("GIT_CONFIG_GLOBAL", OsString::from("")),
+            ("GIT_CONFIG_SYSTEM", OsString::from("")),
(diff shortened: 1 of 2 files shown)


mlcui-google:
Closing due to a better way of doing this - see https://github.com/arxanas/git-branchless/pull/674#issuecomment-1325960671.

mlcui-google:
For future readers, the above commit message is totally incorrect - global configs shouldn't be applying in tests as `$HOME` and `$XDG_CONFIG_HOME` isn't set. This only happens when using a wrapper around `git` which sets `$HOME`.

PR #685, opened 2022-11-29
feat(submit): Apply styles to branch names

claytonrcarter:
This is a small patch to `git submit` to style the pushed/skipped branches with colors. My goal is to improve on the default output, which doesn't make it very easy to quickly identify the pushed/skipped branches w/o reading the output carefully. 

Screen Shot 2022-11-28 at 9 42 16 PM

- This was mostly lifted from `git test` and how it styles passing/skipped/failed tests.

### DIFF
--- a/git-branchless/src/commands/submit.rs
+++ b/git-branchless/src/commands/submit.rs
@@ -1,17 +1,26 @@
 use std::fmt::Write;
 use std::time::SystemTime;
 
+use cursive_core::theme::{BaseColor, Effect, Style};
 use itertools::{Either, Itertools};
+use lazy_static::lazy_static;
 use lib::core::dag::{commit_set_to_vec, Dag};
 use lib::core::effects::{Effects, OperationType};
 use lib::core::eventlog::{EventLogDb, EventReplayer};
-use lib::core::formatting::Pluralize;
+use lib::core::formatting::{Pluralize, StyledStringBuilder};
 use lib::core::repo_ext::RepoExt;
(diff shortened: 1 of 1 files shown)


arxanas (approved):
Thanks, this looks better! I think for the `git test` output, we use bold for both the green and the yellow. I think we should remain internally consistent in this case and either use bold for both sets of branches or neither. What do you think?

claytonrcarter:
I can totally add the bold back in for consistency. I removed it b/c it felt a little heavy handed in this case, but I definitely don't feel strongly about it.

New issue, though: it looks like #650 removed `cursive` from `git-branchless/Cargo.toml`, but this PR adds a dep on `cursive` to `git submit`, which was not extracted to it's own crate. Which course of action would you prefer:
1. Add `cursive` back to `git-branchless/Cargo.toml` just for this change
2. Extract `git submit` as it's own crate a la #650 (I think I can do that by following along to the changes in that PR)
3. Abandon this change as not worth it ... no hard feelings; I promise. 😄

arxanas:
@claytonrcarter does it work to add a dependency on just `cursive_core`?

> Extract git submit as it's own crate

This would be good to do at some point, but `cursive_core` isn't as big of a dependency (although I'm surprised that we don't already need it for output rendering anywhere else in the crate).

claytonrcarter:
Aha! I hadn't done my homework. `cursive_core` *is* already a dependency, and yes it worked to just use that instead. Updated and pushed. Thank you!

arxanas:
I'm liking the coloring so far!

PR #723, opened 2023-01-01
fix(init): warn when using dynamically-calculated hooks path

### LINKED ISSUE #720: Git branchless doesn't detect dynamically set hooks path
### Description of the bug

My company dynamically sets some configuration for the company repos by having uses add this to their ~/.gitconfig

```
[includeIf "hasconfig:remote.*.url:git@github.com:company/**"]
        path = ~/work/company/gitconfigcompany
```

this gitconfigcompany file contains

```
[core]
hooksPath = ~/work/company/hooks
```
where we install custom hooks.

Branchless will not detect this hooks path, and will still install in $GIT_DIR/hooks. But Git will detect this hooks path, and use ~/work/company/hooks, meaning git-branchless's hooks won't work. Even worse, git branchless init doesn't report any error and the aliases still work, meaning you could easily lose a commit if you try to use detached head.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Version of `rustc`

_No response_

### Automated bug report

_No response_

git-branchless-opts 0.7.0-pre.1

### Version of `git`

git version 2.39.0

arxanas:
Closes #720.

### DIFF
--- a/git-branchless-lib/src/core/config.rs
+++ b/git-branchless-lib/src/core/config.rs
@@ -13,12 +13,41 @@ use crate::core::formatting::StyledStringBuilder;
 use crate::git::{ConfigRead, GitRunInfo, GitRunOpts, Repo};
 
 use super::effects::Effects;
+use super::eventlog::EventTransactionId;
+
+/// Get the expected hooks dir inside `.git`, assuming that the user has not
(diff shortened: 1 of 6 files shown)


slymon99:
fixes it 👍

PR #739, opened 2023-01-11
feat(revset): add "abandoned()" revset

john-h-k:
WIP - I don't know enough about `find_abandoned_children` yet to understand the rest of the parameters

### DIFF
--- a/git-branchless-revset/src/builtins.rs
+++ b/git-branchless-revset/src/builtins.rs
@@ -23,6 +23,7 @@ type FnType = &'static (dyn Fn(&mut Context, &str, &[Expr]) -> EvalResult + Sync
 lazy_static! {
     pub(super) static ref FUNCTIONS: HashMap<&'static str, FnType> = {
         let functions: &[(&'static str, FnType)] = &[
+            ("abandoned", &fn_abandoned),
             ("all", &fn_all),
             ("none", &fn_none),
             ("union", &fn_union),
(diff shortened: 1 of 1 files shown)


git-branchless-revset/src/builtins.rs
```diff
@@ -60,6 +61,21 @@ lazy_static! {
...
 }
 
+#[instrument]
+fn fn_abandoned(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
+    eval0(ctx, name, args)?;
+
+    let obsolete_commits = ctx.dag.query_obsolete_commits();
+    let abandoned_children = find_abandoned_children(
+        obsolete_commits,
+        // ?
```
arxanas, line 71: The `draft_commits` parameter is actually unused, so I opened https://github.com/arxanas/git-branchless/pull/744 to remove it.

git-branchless-revset/src/builtins.rs
```diff
@@ -60,6 +61,21 @@ lazy_static! {
...
+        // ?
+        // ?
+        // ?
```
arxanas, line 74: For these three parameters, you can copy the lines here:

https://github.com/arxanas/git-branchless/blob/c6292d28cbca375785720e46c86270fbbbccf7ff/git-branchless-revset/src/builtins.rs#L493-L500

Ideally, we would change it to lazily construct these objects, but that's harder to do in Rust.

john-h-k: What is the use of the `oid` param? My guess was that it was the "base"(?) commit to start finding from but thought I would check before trying to supply it

arxanas: Oops, that's an important parameter to have forgotten to describe. It's probably best to check the source in case I misremember. I think it's the OID of the old base commit, which should have been rewritten by now, whose children are currently abandoned.

john-h-k: So in this context (looking for all abandoned commits), should it be the root commit of the repo?

arxanas: This only checks the immediate children of the provided commit, not all descendants recursively. It would be best to follow the logic in `restack.rs`. Here, we use the list of all obsolete commits as the starting point:

https://github.com/arxanas/git-branchless/blob/ac8a1876cfd7c14ff4ecc99e3ca54d7edc989180/git-branchless/src/commands/restack.rs#L100

arxanas:
Closing this PR for now since it's been inactive for a while. Feel free to re-open later.

PR #740, opened 2023-01-11
feat(revset): add "obsolete()" revset 

john-h-k:
Adds `obsolete()` reset

### DIFF
--- a/git-branchless-revset/src/builtins.rs
+++ b/git-branchless-revset/src/builtins.rs
@@ -28,6 +28,7 @@ lazy_static! {
             ("union", &fn_union),
             ("intersection", &fn_intersection),
             ("difference", &fn_difference),
+            ("obsolete", &fn_obsolete),
             ("only", &fn_only),
             ("range", &fn_range),
             ("not", &fn_not),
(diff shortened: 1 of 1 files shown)


arxanas:
Thanks! This might require writing some tests, because there's some logic in revset evaluation that discards commits which aren't visible, so it's possible that this doesn't do what we want it to do in practice.

My expectation is that `obsolete()` would return commits which are obsolete but still visible, i.e. have visible children, but I could be wrong. Incidentally, this might suffice for your purposes: `children(obsolete()): - obsolete()` might be the same as the set of abandoned commits? It depends on whether you only want to include commits which can be restacked, as it's possible to abandon a commit manually with `git hide`, for which there would be no restack target.

It would be best to add a new test function to `eval.rs` with a commit graph with obsolete/rewritten commits, and test this there (and your possible future `abandoned()` revset function).

arxanas (changes requested):
Requesting tests as per the previous comment.

arxanas:
Closing this PR for now since it's been inactive for a while. Feel free to re-open later.

PR #758, opened 2023-01-20
style: diff.deltas() is already an iterator

samueltardieu:
Calling .into_iter() in this context is a no-op and is flagged by clippy.


martinvonz (approved):

samueltardieu:
(incidentally this is a check I added to clippy a few weeks ago, I'm happy to see it in action)

arxanas:
Thanks!

PR #777, opened 2023-02-03
feat(test): create `git test run --search` option

arxanas:


.github/workflows/lint.yml
```diff
@@ -26,6 +26,14 @@ jobs:
             exit 1
           fi
 
+      - name: Forbid dbg string
+        run: |
+          if results=$(git grep --column --line-number --only-matching 'dbg''!'); then
+            echo "$results"
+            awk <<<"$results" -F ':' '{ print "::error file=" $1 ",line=" $2 ",col=" $3 "::Illegal string: " $4 }'
+            exit 1
+          fi
```
epage, line 35: Any reason not to use `#[warn(clippy::dbg_macro)]`?

arxanas: @epage TIL, thanks! Do you know if there's a way to set the enabled clippy lints for the entire workspace? So far I've just been annotating each `lib.rs` with the same clippy settings.

epage: You can set it in your [`.cargo/config.toml`](https://github.com/EmbarkStudios/rust-ecosystem/blob/main/lints.toml) but that will only works by running  commands in the correct directories.  For example, I run all of my own commands at HEAD via shell scripts wrapping `cargo run` which meant they used  my current project's `.cargo/config.toml` rather than their own which isn't too bad but it causes a lot of force rebuilds.

PR #790, opened 2023-02-11
feat(smartlog): add --reverse option

wabain:
This pull request adds a flag for reverse-chronological smartlog output. 😈 As far as I can tell, all that's needed is to swap the `split` and `merge` glyphs and then to reverse the order when outputting the smartlog line by line.

I've continued to emit hints at the end of the output, mostly just because that was the path of least resistance. I'm agnostic on whether it would make sense to move them to the top.

### DIFF
--- a/git-branchless-smartlog/src/lib.rs
+++ b/git-branchless-smartlog/src/lib.rs
@@ -749,6 +749,10 @@ mod render {
 
         /// The options to use when resolving the revset.
         pub resolve_revset_options: ResolveRevsetOptions,
+
+        /// Reverse the ordering of items in the smartlog output, list the most
+        /// recent commits first.
+        pub reverse: bool,
     }
 
     impl Default for SmartlogOptions {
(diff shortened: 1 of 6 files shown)


git-branchless-smartlog/src/lib.rs
```diff
@@ -851,8 +857,13 @@ pub fn smartlog(
             &mut DifferentialRevisionDescriptor::new(&repo, &Redactor::Disabled)?,
             &mut CommitMessageDescriptor::new(&Redactor::Disabled)?,
         ],
-    )?;
-    for line in lines {
+    )?
+    .into_iter();
+    while let Some(line) = if reverse {
```
wabain, line 862: Using an `if` in the iteration expression here is a bit clunky, but all the alternatives I could think of added more code than seemed necessary. (E.g.: splitting out a separate function, adding a local `&mut dyn Iterator<...>` variable backed by one of two conditionally initialized iterator types, ...)

arxanas: A classic Rust ergonomics problem! Looks fine to me.

arxanas (approved):
Nice work, thanks!

PR #791, opened 2023-02-11
feat(reword): Use distinctive name for tempfile

wabain:
Name the tempfile used by `git reword` `COMMIT_EDITMSG-*.txt`, aligning with the naming of `.git/COMMIT_EDITMSG`. This makes it easier to configure editors to apply syntax highlighting without false positives.


arxanas:
Nice, thanks!

PR #825, opened 2023-02-28
Limit fetch in submit to relevant branches

jakejx:
This is a follow up from my question in https://github.com/arxanas/git-branchless/discussions/671. I have made the changes to only fetch branches that are being pushed. I am testing it locally now and so far it seems to behave as expected but I am not sure if there are other edge cases to consider. I will continue to use this build for a few days and see if I encounter any other issues, but I figured it would be good to just open the MR for feedback first.

Thank you.

EDIT: not sure why the tests are failing for windows. They are passing on my machine.
EDIT2: I was not running tests for the entire workspace. Fixed.

### DIFF
--- a/git-branchless-submit/src/lib.rs
+++ b/git-branchless-submit/src/lib.rs
@@ -159,12 +159,13 @@ These remotes are available: {}",
     };
 
     // TODO: explain why fetching here.
-    let remote_names = remotes_to_branches.keys().sorted().collect_vec();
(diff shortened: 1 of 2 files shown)


arxanas:
Hi @jakejx, thanks for the fix! I want to release version 0.7.0 with the current version of the code and then I'll merge this.

jakejx:
Thanks for merging this! Fully understand the decision to delay the merge till after the release.

PR #829, opened 2023-03-04
tests: set an explicit value for core.abbrev config

melko:
Users can have a value for core.abbrev different than the default one in their ~/.gitconfig, this makes some tests fail because when comparing the output, they expect hashes to be 7 chars long. Fix by setting an explicit value of 7 when initializing a repo inside a test.


arxanas:
Thanks!

Issue #88, opened 2021-08-26
List of Commands?


AriaFallah:
Hey! Love the idea for this project. I'm completely new to it, and I can't really find a list of commands (+flags they can take) anywhere in the docs. Is there something like that?

If not, is there somewhere in the source code I could just read in the meantime? For example, I'd love to learn if `git sl` takes a flag that can show more history than just the past few commits or if there's a flag to put the most recent commit at the top instead of at the bottom.

arxanas:
@AriaFallah Thanks again for your feedback. I've started work on a command overview  in the [Wiki](https://github.com/arxanas/git-branchless/wiki). Please check the articles in the sidebar and let me know if you have any feedback. For example, here's the documentation on `git move`: https://github.com/arxanas/git-branchless/wiki/Command:-git-move

I've also opened #112 to attempt to fix the usage of `--help` under Git for at least some cases.

AriaFallah:
I love it. Thanks a lot :)

martinvonz:
Agreed, great documentation!

A bit off topic, but I have a comment about what to do about the `git move -s 87b20a4b -d xyzzy` at the end (instead of erroring out because it's a cycle). Another reasonable behavior would be to leave the commits in the range between those commits in place but move any side branches on top of the destination. You could then do something like `git move -s <old origin/main> -d <new origin/main>` to rebase all the commits you had off of `<old origin/main>`.

arxanas:
@martinvonz That's a good idea. It's also currently the case that `git move` without any parameters fails because it takes on the default behavior of attempting to rebase the current commit on top of itself, which would form a cycle. Adopting your suggestion would cleanly make it a no-op.

PR #830, opened 2023-03-06
feat(init): create `install-man-pages` subcommand

### LINKED ISSUE #828: No obvious way to generate man pages
### Wiki page

_No response_

### Subcommand

_No response_

### Description

I've been packaging `git-branchless` in the Arch Linux AUR (user repository) for a while now and am considering migrating it to official package repositories. I just bumped the packaging to the new v0.7.0 release, but I'm having trouble figuring out where the rumored man pages are.

I see from #112 the discussion about the topic, but no conclusion. It kind of punts to "disto packager support", but I *am* the distro packager and our tooling does support it, I just can't figure out where to get them *from*. I see in #114 some tooling was added suggesting man pages might be generated by `build.rs` and stuffed in the output dir, but I don't see them there either. In more current code I see a function being run as part of `git branchless init` that suggests it might be stuffing man pages locale to a repository, but I tried that too and didn't find any.

What is the status of having man pages and where can I get them from to add to our packaging?

arxanas:
Closes #828.

### DIFF
--- a/git-branchless-init/src/lib.rs
+++ b/git-branchless-init/src/lib.rs
@@ -22,7 +22,7 @@ use lib::util::ExitCode;
 use path_slash::PathExt;
 use tracing::{instrument, warn};
 
-use git_branchless_opts::{write_man_pages, InitArgs};
+use git_branchless_opts::{write_man_pages, InitArgs, InstallManPagesArgs};
 use lib::core::config::{get_default_branch_name, get_default_hooks_dir, get_hooks_dir};
 use lib::core::dag::Dag;
 use lib::core::effects::Effects;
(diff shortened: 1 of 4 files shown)


arxanas:
> Will the existing init subcommand still try to install man pages locally? I would suggest it not do that by default.

No, it currently doesn't install man pages at all (and it still won't).

> Will the man page install subcommand be hidden from --help? I typically suggest it my a hidden subcommand, but document it in the project readme or install instructions so distro folks in particular catch it but it doesn't clutter up usage info for end users that may not have access to the root system anyway and shouldn't be stuffing stuff there behind their package manager's back unless they know what they are doing.

The majority of git-branchless users install from source and won't have any way to install man pages (which breaks `git <command> -h`, see https://github.com/arxanas/git-branchless/issues/88#issuecomment-906861160), so I think I'll leave it as visible in case someone does want to install man pages on their own (at their own risk).

PR #857, opened 2023-03-17
docs: Suggest checking package repositories before cargo install

alerque:
With v0.7.1 I am migrating `git-branchless` [to the official repositories for Arch Linux](https://archlinux.org/packages/community/x86_64/git-branchless/). That makes quite a few platforms with prebuilt packages now. I would suggest that *if* the user is on a system with system packages available that is almost certainly a better path for them than directly compiling and installing via `cargo` (completions, access from other users, automatic upgrades, etc.). As such I suggest suggesting it in the "short version" of how to install.


arxanas (approved):
Thanks!

PR #859, opened 2023-03-18
refactor: create `EyreExitOr` type

arxanas:


arxanas:
This seems to be difficult because `ExitCode` implicitly gets converted into an `eyre::Error`.

PR #900, opened 2023-04-17
fix(git): fix "not a git repository" when running in `repo` projects

mlcui-google:
This works around https://github.com/libgit2/libgit2/issues/6401 by using the current working directory, instead of `libgit2`'s incorrect repo working copy path, when a `repo` project is detected.


arxanas:
Thanks!

PR #907, opened 2023-04-18
tests(git): add test for running commands in `repo` tool projects

mlcui-google:
I forgot to add a test in #900, oops!

This test is Windows only, as creating symlinks on Windows requires administrator / developer privileges.


arxanas (approved):
Thanks also!

PR #909, opened 2023-04-19
`scm-diff-editor` binary

arxanas:
This PR creates the `scm-diff-editor` binary, a TUI application for selecting subsets of changes in a diff. It can be used with `jj` as the `diff-editor` to facilitate commands like `jj move` and `jj split`.

To install the `scm-diff-editor` binary:

```sh
$ cargo install --git https://github.com/arxanas/git-branchless --branch scm-diff-editor scm-record --features scm-diff-editor
```

Then set the following in your `jjconfig.toml`:

```toml
[ui]
diff-editor = ["scm-diff-editor", "--dir-diff", "$left", "$right"]
```

### DIFF
--- a/scm-record/src/bin/scm-diff-editor.rs
+++ b/scm-record/src/bin/scm-diff-editor.rs
@@ -298,7 +298,7 @@ mod render {
         change_type: ChangeType,
     ) -> Vec<SectionChangedLine<'static>> {
         contents
-            .lines()
+            .split_inclusive('\n')
             .map(|line| SectionChangedLine {
                 is_toggled: false,
                 change_type,
@@ -903,7 +903,7 @@ qux2
                             SectionChangedLine {
                                 is_toggled: false,
                                 change_type: Added,
-                                line: "right",
+                                line: "right\n",
                             },
                         ],
                     },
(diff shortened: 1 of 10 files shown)


scm-record/src/bin/scm-diff-editor.rs
```diff
@@ -0,0 +1,606 @@
...
+            scm_record::SelectedContents::Absent => match fs::remove_file(&file_path) {
+                Ok(()) => {}
+                Err(err) if err.kind() == io::ErrorKind::NotFound => {}
+                Err(err) => {
+                    return Err(Error::RemoveFile {
+                        path: file_path,
+                        source: err,
+                    });
+                }
+            },
```
martinvonz, line 472: For added files that have not been selected, I guess `get_selected_contents()` will not mention the file, so we won't get an `Absent` match here even though we want to remove the file.

scm-record/src/bin/scm-diff-editor.rs
```diff
@@ -0,0 +1,606 @@
+//! Supporting library for
```
arxanas, line 1: TODO: fix doc-comment.

PR #910, opened 2023-04-19
fix(git): simplify and generalise "symlinked `.git` directory" check

mlcui-google:
This handles `repo`-managed working copies, and should be generic enough to handle most other cases (unless the symlink points to a child of an ancestor).

### DIFF
--- a/git-branchless-lib/src/git/run.rs
+++ b/git-branchless-lib/src/git/run.rs
@@ -3,7 +3,7 @@ use std::convert::TryInto;
 use std::ffi::{OsStr, OsString};
 use std::fmt::Write;
 use std::io::{BufRead, BufReader, Read, Write as WriteIo};
-use std::path::{Component, Path, PathBuf};
+use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


mlcui-google:
I haven't seen any false positives with this, including the case if the root of the working tree is a symlink.

See #900 for the PR that implemented the original check.

mlcui-google:
DO NOT SUBMIT THIS PR until I have tested this with Git worktrees. I have a suspicion that this will result in incorrect behaviour with them...

mlcui-google:
I just tested it with a worktree and it seems to work fine. Apologies for the false alarm - I haven't used worktrees before, so I wasn't sure about how things are handled in them...

arxanas (approved):
Looks good to me, thanks! I've added you as a collaborator on the repository so you can land this at your leisure.

mlcui-google:
Thanks!

This PR feels a bit risky, but from my (limited) testing it should be okay. I'll keep an eye out for any issues relating to this.

PR #913, opened 2023-04-21
[WIP] hook: Ignore updates to remote branches in reference-transaction

pink:


pink:
Are there any additional ways I should verify this change? Also I'm getting some lint errors from `git-branchless-opts` but the files were untouched.

git-branchless-hook/src/lib.rs
```diff
@@ -456,7 +456,12 @@ fn hook_reference_transaction(effects: &Effects, transaction_state: &str) -> eyr
                  ref_name,
                  old_oid: _,
                  new_oid: _,
-             }| !should_ignore_ref_updates(ref_name),
+             }| {
+                !should_ignore_ref_updates(ref_name)
```
arxanas, line 460: Rather than render this to a string, you'll want to write `match CategorizedReferenceName::new(...) { ... }` and check for the remote branch case.

arxanas (changes requested):
Thanks! Ignore the lint failure for now. The linter runs using Rust nightly. Normally, I would just fix the issue in a separate commit, but the issue is occurring inside a macro expansion, so we can't actually fix it ourselves.

git-branchless-hook/src/lib.rs
pink: Ah I'm new to rust, is this the right way to match against enum variants?

git-branchless-hook/src/lib.rs
```diff
@@ -458,9 +458,10 @@ fn hook_reference_transaction(effects: &Effects, transaction_state: &str) -> eyr
                  new_oid: _,
              }| {
                 !should_ignore_ref_updates(ref_name)
-                    && !CategorizedReferenceName::new(ref_name)
-                        .friendly_describe()
-                        .starts_with("remote")
+                    && match CategorizedReferenceName::new(ref_name) {
+                        CategorizedReferenceName::RemoteBranch { name: _, prefix: _ } => false,
+                        _ => true,
```
arxanas, line 463: Yep, this is how we destructure enums. Let's just make a few minor changes:

- Let's explicitly list out the missing cases in `_`. The idea is that if a new case is added, the exhaustiveness checker will make us look at this `match` again and update it to ensure that the logic is right.
  - Now that I think about all of the cases... let's return `false` for both `RemoteBranch` *and* `Other`. Occasionally, other tooling generates references that we also don't care about (such as `git bisect` or perhaps [Jujutsu](https://github.com/martinvonz/jj)'s `keep` refs).
- In this case, we can write `Foo { .. }` instead of `Foo { field1: _, field2: _ }` to ignore all fields in the enum case.
  - Actually, I don't recommend doing this in general, but the `CategorizedReferenceName` enum case was somewhat poorly designed. Very often we ignore the enum fields and use one of the helper methods on the type. I don't think the fields ought to have been attached to the enum cases to begin with.

arxanas (changes requested):

arxanas:
I merged https://github.com/arxanas/git-branchless/pull/916, so if you want to get rid of the lint errors in CI, you can try updating on top of that.

git-branchless-hook/src/lib.rs
pink: Going through and updating the failing tests for this change. Returning false for `Other` appears to [skip processing updates](https://github.com/arxanas/git-branchless/actions/runs/4776616837/jobs/8491756315#step:6:308) to `HEAD` - just wanted to double check that this behavior is expected?

arxanas: @pink I think it's fine (although it didn't occur to me ahead of time that `HEAD` is an `Other` reference). It looks like none of the tests are failing as a result. You can update the snapshot tests manually or by using the `cargo-insta` tool. The `HEAD` reference is also tracked via the `post-checkout` hook, so we're not losing any data by dropping the update.

Also, I forgot to mention that you can collapse the match cases by using `|`:

```rust
match expr {
    Foo::Bar { .. } | Foo::Baz { .. } => false,
    Foo::Qux => true,
}
```

(You can do this as long as the set of variables bound in each sub-pattern is the same. Since we're using `..`, we're not binding any variables, so it should work.)

PR #922, opened 2023-04-29
feat(record:scm-diff-editor): add key "a" for toggling all selections

martinvonz:
This copies a feature from hg's record interface. It's most often useful before you do anything, to select all lines.


arxanas:
Thanks! The test failure looks spurious, so I restarted the jobs.

Is the behavior of Mercurial precisely to toggle all lines? That is, if there is a mix of toggled lines, does it invert each one, or does it try to match their states so that they're all either toggled or untoggled?

martinvonz:
Mercurial has "a" for this behavior of toggling everything, but it also has "A" (i.e. shift+A) for toggling all on/off.

PR #929, opened 2023-05-06
scm-record features

arxanas:
Sticky header demo:

https://user-images.githubusercontent.com/454057/236634332-7f8e3853-e7e6-4cb8-894d-7fca52bc66a1.mov

### DIFF
--- a/scm-record/README.md
+++ b/scm-record/README.md
@@ -31,8 +31,8 @@ Ideally, these would be documented in the UI itself.
 - `c`: confirm changes and quit.
 - `f`: expand/collapse the current item.
 - `F`: expand/collapse all items.
-- `up`: select the next item.
-- `down`: select the previous item.
+- `up`/`k`: select the next item.
+- `down`/`j`: select the previous item.
 - `space`: toggle the current item.
 - `enter`: toggle the current item and move to the next item of the same kind.
 - `a`: invert the toggled state of each item.
--- a/scm-record/README.md
+++ b/scm-record/README.md
@@ -1,0 +1,65 @@
+# License
+
+MIT OR Apache-2.0
+
+# scm-record
+
+`scm-record` is a UI component to interactively select changes to include in a commit. It's meant to be embedded in source control tooling.
+
+You can think of this as an interactive replacement for `git add -p`, or a reimplementation of `hg crecord`/`hg commit -i`. Given a set of changes made by the user, this component presents them to the user and lets them select which of those changes should be staged for commit.
+
+`scm-record` was originally developed as a supporting library for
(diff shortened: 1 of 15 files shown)


scm-record/src/types.rs
```diff
@@ -288,6 +288,19 @@ impl File<'_> {
         (acc_selected, acc_unselected)
     }
 
+    /// Set the selection of all sections in this file.
+    pub fn set_toggled(&mut self, toggled: bool) {
```
martinvonz, line 356: What do you think about calling it `set_selected()`? I find "toggled" (the adjective) a bit unclear. I think of "toggle" (the verb) as switching something between two states, but it's not obvious what to me what "set toggled" means. `set_selected()` would also better match the function's description.

scm-record/README.md
```diff
@@ -0,0 +1,64 @@
...
+# Feature wishlist
+
+Here are some features in the UI which are not yet implemented:
+
+- Select inner/outer element.
+- Jump to next/previous element of same kind.
+- Expand/collapse sections.
+- Menu bar to explain available actions and keybindings.
+- Toggle all items within some scope.
+- "Sticky" file and/or section headers.
```
martinvonz: Add "edit one side of the diff in an editor"? That's a feature I miss compared to Meld. The editing could be done directly in scm-record, but I suspect we don't want to write our own editor.

scm-record/src/ui.rs
```diff
@@ -211,28 +211,27 @@ impl From<crossterm::event::Event> for Event {
             ) => Self::PageDown,
 
             Event::Key(KeyEvent {
-                code: KeyCode::Up,
+                code: KeyCode::Up | KeyCode::Char('k'),
```
martinvonz, line 215: Update README with the new keys?

scm-record/README.md
```diff
@@ -57,7 +57,6 @@ Here's some projects that don't use `scm-record`, but could benefit from integra
 
 Here are some features in the UI which are not yet implemented:
 
-- Select inner/outer element.
```
martinvonz: Looks like the Keybindings section should be update with `left` and `right` now (and `j` and `l`)

scm-record/src/types.rs
arxanas: Updated to `checked`. I think a "selection" is an item or subset of items which can then be checked or toggled.

scm-record/README.md
arxanas: Added.

arxanas:
@martinvonz thanks for the review!

martinvonz (approved):
Thanks! I use scm-record ~daily

PR #939, opened 2023-05-22
feat(hooks): Suppress warning for peeled refs while parsing packed-refs

claytonrcarter:
As discussed on [discord](https://discord.com/channels/915309546984050709/915309546984050712/1108952258877599855), this suppresses the warnings generated when `branchless` encounters a peeled ref while parsing the `packed-refs` file. This does make any attempt to process these refs correctly, which -- as I understand it -- would be to associate there OIDs with the preceding "normal" ref.


arxanas:
Thanks!

PR #978, opened 2023-06-29
feat(nix): add 'scm-diff-editor' package

thoughtpolice:
This makes it trivial to install `scm-diff-editor` using `nix-env` or `nix profile install`, e.g.

    nix profile install github:arxanas/git-branchless#scm-diff-editor


thoughtpolice:
Mostly a copy-paste job (if it works it works!), but the tests pass, too.

arxanas:
Thanks!

PR #1071, opened 2023-09-23
cargo: add LICENSE file to each crate we publish

martinvonz:
According to clap-rs/clap#2810, the Apache license must be distributed with sources, including with the sources we publish to crates.io. I suspect the MIT has the same requirement.


arxanas:
Interesting, thanks for the fix!

PR #1095, opened 2023-10-10
fix(hooks): use same `branchless` dir for all worktrees

arxanas:
Closes #928.

### DIFF
--- a/git-branchless-lib/src/git/repo.rs
+++ b/git-branchless-lib/src/git/repo.rs
@@ -565,26 +565,36 @@ impl Repo {
         Ok(Config::from(config))
     }
 
+    /// Get the directory where all repo-specific git-branchless state is stored.
+    pub fn get_branchless_dir(&self) -> Result<PathBuf> {
+        let dir = self.get_path().join("branchless");
+        std::fs::create_dir_all(&dir).map_err(|err| Error::CreateBranchlessDir {
+            source: err,
+            path: dir.clone(),
+        })?;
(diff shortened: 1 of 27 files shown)


git-branchless-lib/src/git/repo.rs
```diff
@@ -565,26 +565,41 @@ impl Repo {
         Ok(Config::from(config))
     }
 
+    /// Get the directory where all repo-specific git-branchless state is stored.
+    pub fn get_branchless_dir(&self) -> Result<PathBuf> {
+        let maybe_worktree_parent_repo = self.open_worktree_parent_repo()?;
```
miquella, line 570: A couple of weeks ago I had a hunch that the issue I described in #928 was being caused by `git-branchless` using separate directories for each worktree (sorry for not being more responsive in that discussion, it's been a busy few months!).

I even pulled down the code and started digging through things to see if I could figure out where the changes might need to be made. One thing I did come across was `git_repository_commondir` from `libgit2`, which seems like it may be a built-in version of `open_worktree_parent_repo`, but I'm not 100% sure. Unfortunately, the `git2` crate doesn't seem to expose it and that's as far as I got before I ran out of time that day 😝

https://libgit2.org/libgit2/#HEAD/group/repository/git_repository_commondir

arxanas: Ah, commondir does sound like what I'm looking for (also available on the CLI as `git rev-parse --git-common-dir`).

PR #1097, opened 2023-10-10
fix(smartlog) smartlog does not panic when run in a non-git directory

### LINKED ISSUE #1069: Panic on `git branchless init` in a non-git directory
### Description of the bug

First time using git branchless, I accidently ran the command in the parent folder of my git repos, rather than in a git repo.

I did not expect an uncaught panic.

### Expected behavior

A polite error message informing me of the mistake I'd made.

### Actual behavior

A panic.

```
   0: could not open repository: could not find repository from '/Users/dragon/dxw/archives'; class=Repository (6); code=NotFound (-3)
   1: could not find repository from '/Users/dragon/dxw/archives'; class=Repository (6); code=NotFound (-3)

Location:
   git-branchless-init/src/lib.rs:606

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: git_branchless_init::command_init with effects=<Output fancy=true> git_run_info=<GitRunInfo path_to_git="git" working_directory="/Users/dragon/dxw/archives" env=not shown> main_branch_name=None
      at git-branchless-init/src/lib.rs:599
   1: git_branchless_init::command_main with ctx=CommandContext { effects: <Output fancy=true>, git_run_info: <GitRunInfo path_to_git="git" working_directory="/Users/dragon/dxw/archives" env=not shown> } args=InitArgs { uninstall: false, main_branch_name: None }
      at git-branchless-init/src/lib.rs:668

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
Location: git-branchless/src/commands/mod.rs:235

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

### Version of `rustc`

_No response_

### Automated bug report

_No response_

git-branchless-opts 0.8.0

### Version of `git`

git version 2.42.0

cshinaver:
Closes #1069 

First time contributing! tackling one of the smaller issues before I look into something more major.


arxanas (changes requested):
Hi @cshinaver, thanks for the PR! As written, this would only handle the panic for the case of `git branchless smartlog`, and only when failing to open the `Repo`. I would prefer to see an approach that can handle more panics. The best approach overall might be to replace `EyreExitOr` with an `enum` something like this:

```
enum CommandResult<T> {
    Ok(T),
    Exit { exit_code: isize, message: String },
    Unhandled(eyre::Error),
}
```

Then we could update the `try_exit_code!` macro (or create a new macro) that can wrap the result of `Repo::from_dir` and produce a `CommandResult::Exit` instead of letting the error propagate normally; and update `do_main_and_drop_locals` to handle the `CommandResult` and print the error message from the wrapped function.

PR #1098, opened 2023-10-12
(#1088) post-rewrite hook does not track intermediate commits from an interactive rebase

### LINKED ISSUE #1088: git rebase -i "fixup" strands a bunch of commits after squashing
### Description of the bug

I saw on the wiki that interactive rebase is fully supported, but when I "fixup"'d multiple commits with `git rebase -i`, all the interim commits were rewritten and stranded. The `smartlog` shows all of these commits as separate branches with useless commit messages. 

```
➜  repo git:(9b123ddbd9) git smartlog
⋮
◇ c6e6dba 2h (master) add cache (#37637)
┣━┓
┃ ◯ 8aee19e 6m # This is a combination of 2 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 9b7ce05 6m # This is a combination of 4 commits. # This is the 1st commit message:
┣━┓
┃ ◯ c941dd5 6m # This is a combination of 5 commits. # This is the 1st commit message:
┣━┓
┃ ◯ efeb7ad 6m # This is a combination of 3 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 9cbe061 6m # This is a combination of 6 commits. # This is the 1st commit message:
┣━┓
┃ ◯ bbf1d40 6m # This is a combination of 8 commits. # This is the 1st commit message:
┣━┓
┃ ◯ c45bab2 6m # This is a combination of 7 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 38c3f4b 6m # This is a combination of 9 commits. # This is the 1st commit message:
┣━┓
┃ ◯ e777063 6m # This is a combination of 10 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 0bf8ab5 6m # This is a combination of 13 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 5917b7b 6m # This is a combination of 11 commits. # This is the 1st commit message:
┣━┓
┃ ◯ d03e265 6m # This is a combination of 12 commits. # This is the 1st commit message:
┣━┓
┃ ◯ 0b8b5ff 6m # This is a combination of 14 commits. # This is the 1st commit message:
```

### Expected behavior

I expected the fixup commits to be hidden from the smartlog and only the remaining rewritten commit to be present. 

### Actual behavior

_No response_

### Version of `rustc`

_No response_

### Automated bug report

_No response_

git-branchless-opts 0.8.0

### Version of `git`

git version 2.37.1 (Apple Git-137.1)

cshinaver:
Fixes #1088 

I was able to repro the issue from #1088. It seems that we are tracking the intermediate commits created by a rebase and adding them to the events db during the post rewrite hook. I noticed that this code was added to prevent aborted rebases from adding intermediate commits, but the test written for that still passes. Looking for feedback on the approach from @arxanas regarding omitting deferred commits from the event log during the post-rewrite hook.

### DIFF
--- a/git-branchless-lib/src/core/rewrite/rewrite_hooks.rs
+++ b/git-branchless-lib/src/core/rewrite/rewrite_hooks.rs
@@ -3,10 +3,9 @@
 use std::collections::{HashMap, HashSet};
 
 use std::fmt::Write;
-use std::fs::{self, File};
-use std::io::{self, stdin, BufRead, BufReader, Read, Write as WriteIo};
+use std::fs::File;
+use std::io::{stdin, BufRead, BufReader, Read, Write as WriteIo};
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
 use std::time::SystemTime;
 
 use console::style;
@@ -42,20 +41,6 @@ use super::{find_abandoned_children, move_branches};
 /// available in the event log anywhere. This is probably acceptable.
 pub fn get_deferred_commits_path(repo: &Repo) -> PathBuf {
     repo.get_rebase_state_dir_path().join("deferred-commits")
-}
-
-fn read_deferred_commits(repo: &Repo) -> eyre::Result<Vec<NonZeroOid>> {
-    let deferred_commits_path = get_deferred_commits_path(repo);
-    let contents = match fs::read_to_string(&deferred_commits_path) {
-        Ok(contents) => contents,
-        Err(err) if err.kind() == io::ErrorKind::NotFound => Default::default(),
-        Err(err) => {
-            return Err(err)
-                .with_context(|| format!("Reading deferred commits at {deferred_commits_path:?}"))
-        }
-    };
-    let commit_oids = contents.lines().map(NonZeroOid::from_str).try_collect()?;
-    Ok(commit_oids)
 }
 
 #[instrument(skip(stream))]
@@ -147,19 +132,6 @@ pub fn hook_post_rewrite(
     let conn = repo.get_db_conn()?;
     let event_log_db = EventLogDb::new(&conn)?;
     let event_tx_id = event_log_db.make_transaction_id(now, "hook-post-rewrite")?;
-
-    {
-        let deferred_commit_oids = read_deferred_commits(&repo)?;
-        let commit_events = deferred_commit_oids
-            .into_iter()
-            .map(|commit_oid| Event::CommitEvent {
-                timestamp,
-                event_tx_id,
-                commit_oid,
-            })
-            .collect_vec();
-        event_log_db.add_events(commit_events)?;
-    }
 
     let (rewritten_oids, rewrite_events) = {
         let rewritten_oids = read_rewritten_list_entries(&mut stdin().lock())?;
--- a/git-branchless-lib/src/testing.rs
+++ b/git-branchless-lib/src/testing.rs
@@ -498,14 +498,15 @@ then you can only run tests in the main `git-branchless` and \
     /// used to set the commit timestamp, which is factored into the commit
     /// hash. The filename is always appended to the message prefix.
     #[instrument]
-    pub fn commit_file_with_contents_and_message(
+    pub fn commit_file_with_contents_and_message_and_file_name(
         &self,
         name: &str,
         time: isize,
         contents: &str,
         message_prefix: &str,
+        file_name: &str,
     ) -> eyre::Result<NonZeroOid> {
-        let message = format!("{message_prefix} {name}.txt");
+        let message = format!("{message_prefix} {file_name}");
         self.write_file_txt(name, contents)?;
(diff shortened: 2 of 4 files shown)


cshinaver:
The test that is failing is doing a commit in the middle of a rebase. The code I removed is meant to preserve commits done in the middle of the rebase. is this intentional? The final commits should be reserved, but if we want to track the intermediate commits, then all of git's automatically created intermediate commits are also tracked (#1088 )

arxanas (changes requested):
Thanks for looking into this! It's been a thorn in many people's sides for a while.

> I noticed that this code was added to prevent aborted rebases from adding intermediate commits, but the test written for that still passes

It looks like the test `test_rebase_no_process_new_commits_until_conclusion` is failing in CI. Unfortunately, due to bad reasons, you might have to run `cargo build --workspace` after your change and then run tests (related to https://github.com/arxanas/git-branchless/wiki/Runbook#testing-subcommands-in-isolation). You should also make sure that you're testing with a recent version of Git, as the test body is gated behind `Git::supports_reference_transactions`.

> The final commits should be reserved, but if we want to track the intermediate commits, then all of git's automatically created intermediate commits are also tracked

I think it's okay to not preserve intermediate commits directly as long as 1) the final commits are preserved and 2) aborting the rebase does not preserve the final commits. There may have been a workflow where one uses `edit` and creates new commits during a rebase, which, in principle, should be tracked, but I am willing to sacrifice that workflow if necessary to support the `fixup` workflow.

I suspect (but haven't investigated) that the `fixup` rebase command in Git does not register the fixed-up commits to be emitted as part of the `post-rewrite` hook, so, as far as git-branchless knows, they're legitimate commits. If so, then, in principle, it would be good to upstream a fix for that to Git, but it wouldn't solve most of our users' `fixup` problems for a long while.

I would be willing to accept some kind of hack/workaround for `fixup` specifically, like maybe ignoring commits whose commit messages start with `#` (which it looks like interactive `fixup` commits do, but not sure about `fixup!` with `git rebase --autosquash`). Another idea is that, during the `post-commit` hook, you could check the current status of the rebase and see if the most recent command was a `fixup`, although I'm not sure how feasible it is.

cshinaver:
In the failing test, a commit (`test 4`) is added in the middle of the rebase. In my change, this commit is ignored. Given that you mentioned you would be okay sacrificing the commit in middle of rebase workflow, I'll update the unit test to not commit in the middle of the rebase.

git-branchless-lib/src/testing.rs
```diff
@@ -498,14 +498,15 @@ then you can only run tests in the main `git-branchless` and \
     /// used to set the commit timestamp, which is factored into the commit
     /// hash. The filename is always appended to the message prefix.
     #[instrument]
-    pub fn commit_file_with_contents_and_message(
+    pub fn commit_file_with_contents_and_message_and_file_name(
```
arxanas, line 501: I would rather keep the API smaller. Can you revert this change and call `write_file` followed by `commit_file` instead? (Or feel free to use `git.run(&["commit", ...])`; the timestamp should still be stable for tests.)

git-branchless-lib/src/core/rewrite/rewrite_hooks.rs
```diff
@@ -148,19 +133,6 @@ pub fn hook_post_rewrite(
     let event_log_db = EventLogDb::new(&conn)?;
     let event_tx_id = event_log_db.make_transaction_id(now, "hook-post-rewrite")?;
 
-    {
```
arxanas, line 151: Is there code that also writes the deferred commits that is now also obsolete? If so, we should delete it too. Or, alternatively, keep all the code but comment out the `add_events` line with a note, as we might need similar code in the future to fix certain other bugs.

git-branchless/tests/test_hooks.rs
```diff
@@ -385,3 +388,54 @@ fn test_git_am_recorded() -> eyre::Result<()> {
...
+
+    run_in_pty_with_command(
+        &git,
+        &["rebase"],
+        &["-i", "--autosquash", "master"],
+        &[
+            PtyAction::WaitUntilContains(" "),
+            PtyAction::Write(CARRIAGE_RETURN),
+        ],
+    )?;
```
arxanas, line 429: Can we just write `git.run(&["rebase", "--autosquash", "master"])`? It seems to work in my testing. If so, do that and revert the changes to `run_in_pty`.

arxanas (changes requested):
Thanks, I think the approach/tests are good; we should just try to not expand the testing API surface unless necessary. (It's already bigger than it should be — for example, practically none of the tests actually need to set a timestamp explicitly, and it's just convention by now...)

claytonrcarter:
Superceded by (merged as part of) #1419

PR #1115, opened 2023-10-25
build: update `ratatui` to v0.24

arxanas:


arxanas:
Can't upgrade yet as new MSRV is 1.70.0.

PR #1126, opened 2023-11-08
Change 'suggest' to 'discuss'

snan:
Several of the suggestions are already pretty good,
so I wanted to rephrase it so that people gave them
a look too, without preventing newer, even awesomer
suggestions, of course.


claytonrcarter:
Makes sense to me. Thank you!

PR #1127, opened 2023-11-08
fix(status): rudimentary support for filenames with spaces

claytonrcarter:
The regex in `StatusEntry::try_from()` was mistaken parsing status lines for files with spaces in such a way to discard the first word in the path name. This attempts to make that less likely by making the regex more specific, but I don't think it's really unavoidable without completely reworking this part of the code to be less flexible. eg if we parsed each of the 1/2/u statuses separately, then we could be more specific about how many fields we need to match/capture for each option.

I suspect that this issue was already rare, and that this fix will make it even moreso. Given that, I feel like a more robust/complex fix probably isn't warranted. With this fix, files like "r101 project.txt" or "Coffee notes.txt" will work, but "R101 project.txt" or "coffee notes.txt" ("R101" looks like a "change score", and "coffee" is a ~~valid~~ hex value.) (**Edit:** 🤦 `coffee` is not a "hex word"; `c0ffee` is, though.)

The options that come to mind to tighten this up further:
- parse each 1/2/u status separately, making field counts and patterns more discrete
- require a minimum number of chars in an object ID (on my Mac, the longest "hex word" in /usr/share/dict/words is 7 chars, so requiring at least 8 chars in an OID would eliminate "coffee"-style false positives)

### DIFF
--- a/git-branchless-lib/src/git/status.rs
+++ b/git-branchless-lib/src/git/status.rs
@@ -200,7 +200,8 @@ impl TryFrom<&[u8]> for StatusEntry {
                 r#"[\w.]+ "#,                                                // Submodule state.
                 r#"(\d{6} ){2,3}(?P<working_copy_filemode>\d{6}) "#,         // HEAD, Index, and Working Copy file modes;
                                                                              // or stage1, stage2, stage3, and working copy file modes.
-                r#"([\w\d]+ ){2,3}"#,                                        // HEAD and Index object IDs, and optionally the rename/copy score.
+                r#"([a-f\d]+ ){2,3}([CR]\d{1,3} )?"#,                        // HEAD and Index object IDs, and optionally the rename/copy score;
+                                                                             // or stage1, stage2, and stage3 object IDs.
                 r#"(?P<path>[^\x00]+)(\x00(?P<orig_path>[^\x00]+))?$"#       // Path and original path (for renames/copies).
             ))
             .expect("porcelain v2 status line regex");
--- a/git-branchless-lib/tests/test_status.rs
+++ b/git-branchless-lib/tests/test_status.rs
@@ -22,12 +22,38 @@ fn test_parse_status_line() {
 
     assert_eq!(
             StatusEntry::try_from(
+                "1 .M N... 100644 100644 100644 51fcbe2362663a19d132767b69c2c7829023f3da 51fcbe2362663a19d132767b69c2c7829023f3da filename with spaces.rs".as_bytes(),
+            ).unwrap(),
+            StatusEntry {
+                index_status: FileStatus::Unmodified,
+                working_copy_status: FileStatus::Modified,
+                path: "filename with spaces.rs".into(),
+                orig_path: None,
+                working_copy_file_mode: FileMode::Blob,
+            }
+        );
+
+    assert_eq!(
+            StatusEntry::try_from(
                 "1 A. N... 100755 100755 100755 51fcbe2362663a19d132767b69c2c7829023f3da 51fcbe2362663a19d132767b69c2c7829023f3da repo.rs".as_bytes(),
             ).unwrap(),
             StatusEntry {
                 index_status: FileStatus::Added,
                 working_copy_status: FileStatus::Unmodified,
                 path: "repo.rs".into(),
+                orig_path: None,
+                working_copy_file_mode: FileMode::BlobExecutable,
+            }
+        );
+
+    assert_eq!(
+            StatusEntry::try_from(
+                "1 A. N... 100755 100755 100755 51fcbe2362663a19d132767b69c2c7829023f3da 51fcbe2362663a19d132767b69c2c7829023f3da filename with spaces.rs".as_bytes(),
+            ).unwrap(),
+            StatusEntry {
+                index_status: FileStatus::Added,
+                working_copy_status: FileStatus::Unmodified,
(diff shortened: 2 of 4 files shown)


git-branchless/tests/test_amend.rs
```diff
@@ -412,6 +412,41 @@ fn test_amend_head() -> eyre::Result<()> {
...
+    let git = make_git()?;
+
+    if !git.supports_committer_date_is_author_date()? {
+        return Ok(());
+    }
+
+    git.init_repo()?;
+    git.detach_head()?;
+    git.commit_file("test file with space", 1)?;
+    git.write_file_txt("test file with space", "updated contents")?;
```
claytonrcarter, line 426: I added a test for `git amend` because that's where I first encountered this, while working on some documentation.

git-branchless/tests/test_amend.rs
```diff
@@ -412,6 +412,41 @@ fn test_amend_head() -> eyre::Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_amend_head_with_file_with_space() -> eyre::Result<()> {
+    let git = make_git()?;
+
+    if !git.supports_committer_date_is_author_date()? {
+        return Ok(());
+    }
```
claytonrcarter, line 421: Oh, this may be leftover from the test I copied to create this one. 🤔

git-branchless-lib/src/git/status.rs
```diff
@@ -200,7 +200,8 @@ impl TryFrom<&[u8]> for StatusEntry {
                 r#"[\w.]+ "#,                                                // Submodule state.
                 r#"(\d{6} ){2,3}(?P<working_copy_filemode>\d{6}) "#,         // HEAD, Index, and Working Copy file modes;
                                                                              // or stage1, stage2, stage3, and working copy file modes.
-                r#"([\w\d]+ ){2,3}"#,                                        // HEAD and Index object IDs, and optionally the rename/copy score.
+                r#"([[a-f]\d]+ ){2,3}([CR]\d{1,3} )?"#,                      // HEAD and Index object IDs, and optionally the rename/copy score;
+                                                                             // or stage1, stage2, and stage3 object IDs.
```
claytonrcarter, line 204: There are numerous ways to update this pattern that could make this less likely to happen, but I don't think we can really eliminate it so long as we use 1 regex for all these porcelain status types. I go into some of them in the PR/commit description.

In particular, the git docs at https://git-scm.com/docs/git-status#_porcelain_format_version_2 state that the `-z` flag (which is used here) causes path names not to be quoted, which ends up leaving the path name a little ambiguous for us.

git-branchless-lib/tests/test_status.rs
```diff
@@ -20,6 +20,19 @@ fn test_parse_status_line() {
             }
         );
 
+    assert_eq!(
+            StatusEntry::try_from(
+                "1 .M N... 100644 100644 100644 51fcbe2362663a19d132767b69c2c7829023f3da 51fcbe2362663a19d132767b69c2c7829023f3da filename with spaces.rs".as_bytes(),
```
claytonrcarter, line 25: These are all identical to the preceding assertions, except for the file name.

git-branchless/tests/test_amend.rs
```diff
@@ -412,6 +412,41 @@ fn test_amend_head() -> eyre::Result<()> {
...
+
+    git.init_repo()?;
+    git.detach_head()?;
+    git.commit_file("test file with space", 1)?;
+    git.write_file_txt("test file with space", "updated contents")?;
+
+    {
+        let (stdout, _stderr) = git.branchless("amend", &[])?;
+        // insta::assert_snapshot!(stderr, @r###"
+        // "###);
```
arxanas, line 431:
```suggestion
```

git-branchless-lib/src/git/status.rs
arxanas: Oops, this is unfortunate. Yeah, refactoring it further sounds non-trivial, so we can hack around it for now.

git-branchless/tests/test_amend.rs
arxanas: I'm assuming this is referring to the commented-out lines below (and that this comment is on the wrong lines now due to a force-push?).

arxanas (approved):
Thanks!

git-branchless/tests/test_amend.rs
claytonrcarter: No, this is the lines I was intending, but I also think that I was wrong here. I was thinking that I wasn't doing anything w/ committer/author data, but then realized that this is probably intended to ensure that the commit IDs generated by `amend` are deterministic for the tests.

git-branchless/tests/test_amend.rs
claytonrcarter: Oops! I left this in from when I was debugging this. Thanks for catching it!

git-branchless-lib/src/git/status.rs
claytonrcarter: Thanks. NOTE that I updated this line before merge because I had mistakenly doubled the square brackets. This is the change I made just now (`[[a-f]\d]+` → `[a-f\d]+`:
```diff
-                r#"([[a-f]\d]+ ){2,3}([CR]\d{1,3} )?"#,
+                r#"([a-f\d]+ ){2,3}([CR]\d{1,3} )?"#,

PR #1129, opened 2023-11-08
feat(submit): add --dry-run flag

claytonrcarter:
`submit` is inherently side effecting in non-obvious-to-undo ways. This is a rudimentary attempt to add a `--dry-run` flag to see what `submit` would be trying to do.

### DIFF
--- a/git-branchless-opts/src/lib.rs
+++ b/git-branchless-opts/src/lib.rs
@@ -396,6 +396,11 @@ pub struct SubmitArgs {
     /// An optional message to include with the create or update operation.
     #[clap(short = 'm', long = "message")]
     pub message: Option<String>,
+
+    /// Don't push or create anything. Instead, report what would be pushed or
+    /// created. (Still triggers a fetch.)
+    #[clap(short = 'n', long = "dry-run")]
+    pub dry_run: bool,
 }
 
 /// Run a command on each commit in a given set and aggregate the results.
--- a/git-branchless-submit/src/lib.rs
+++ b/git-branchless-submit/src/lib.rs
@@ -158,6 +158,7 @@ pub fn command_main(ctx: CommandContext, args: SubmitArgs) -> EyreExitOr<()> {
         resolve_revset_options,
         forge,
         message,
+        dry_run,
     } = args;
     submit(
         &effects,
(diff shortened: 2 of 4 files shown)


git-branchless-opts/src/lib.rs
```diff
@@ -396,6 +396,10 @@ pub struct SubmitArgs {
     /// An optional message to include with the create or update operation.
     #[clap(short = 'm', long = "message")]
     pub message: Option<String>,
+
+    /// Fetch remote but don't push or create anything; only show what would be pushed or created.
+    #[clap(short = 'n', long = "dry-run")]
+    pub dry_run: bool,
```
claytonrcarter, line 402: I have exactly 0 invested in these flags and would welcome any input on what to use. I seem to recall other tools dry run flags being `-n/--dry-run` but I'm open to anything.

git-branchless-submit/src/lib.rs
```diff
@@ -422,7 +431,8 @@ fn submit(
     if !skipped_branch_names.is_empty() {
         writeln!(
             effects.get_output_stream(),
-            "Skipped {} (already up-to-date): {}",
+            "{} {} (already up-to-date): {}",
+            if dry_run { "Would skip" } else { "Skipped" },
```
claytonrcarter, line 435: The tense of these messages felt relevant to change in this context, but another option would be to leave there messages as-is and instead display a separate message w/ something like `This is a DRY RUN; no branches have been pushed or created.`

git-branchless-opts/src/lib.rs
arxanas: We use `-n`/`--dry-run` in one other place, so this is fine.

https://github.com/arxanas/git-branchless/blob/5871499a63f8332e3780554baa9b256d6225dc6e/git-branchless-opts/src/lib.rs#L892

(But... in `scm-record`, we use `-N`/`--dry-run` instead 😬).

git-branchless-submit/src/lib.rs
arxanas: The tense change is nice in my opinion. If you've already done it, then there's no need to change it to the other way.

arxanas (approved):
Thanks!

git-branchless-opts/src/lib.rs
claytonrcarter: Oh, of course, I didn't even think to look around in this project. I've updated the rustdoc comment here to be a little more similar to the other `--dry-run` in this file.

PR #1130, opened 2023-11-08
feat(revsets): add `merges()` and allow `branches()` to filter branch names

claytonrcarter:
- I have found `merges()` to be useful, in particular on public commits.
- for `branches(<pattern>)` see #1117
- ~~`branches(<pattern>)` needs more work; it's currently mush slower that I would expect, and it seems to not match some branchs (at least it didn't match all the branches I expected when I used it to push these recent PRs)~~ **Edit:** I believe that this is fixed.

### DIFF
--- a/git-branchless-revset/src/builtins.rs
+++ b/git-branchless-revset/src/builtins.rs
@@ -3,11 +3,12 @@ use eden_dag::nameset::hints::Hints;
 
 use lib::core::dag::CommitSet;
 use lib::core::eventlog::{EventLogDb, EventReplayer};
+use lib::core::repo_ext::RepoExt;
 use lib::core::rewrite::find_rewrite_target;
 use lib::git::{
-    get_latest_test_command_path, get_test_tree_dir, Commit, MaybeZeroOid, Repo,
-    SerializedNonZeroOid, SerializedTestResult, TEST_ABORT_EXIT_CODE, TEST_INDETERMINATE_EXIT_CODE,
-    TEST_SUCCESS_EXIT_CODE,
+    get_latest_test_command_path, get_test_tree_dir, CategorizedReferenceName, Commit,
+    MaybeZeroOid, Repo, SerializedNonZeroOid, SerializedTestResult, TEST_ABORT_EXIT_CODE,
+    TEST_INDETERMINATE_EXIT_CODE, TEST_SUCCESS_EXIT_CODE,
 };
 use std::borrow::Cow;
 use std::collections::HashMap;
@@ -179,8 +180,48 @@ fn fn_heads(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
 
 #[instrument]
 fn fn_branches(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
-    eval0(ctx, name, args)?;
-    Ok(ctx.dag.branch_commits.clone())
+    let pattern = match eval0_or_1_pattern(ctx, name, args)? {
+        Some(pattern) => pattern,
+        None => return Ok(ctx.dag.branch_commits.clone()),
+    };
+
+    let branch_oid_to_names = ctx
+        .repo
+        .get_references_snapshot()
+        .wrap_err("Could not get references snapshot for repo")
+        .map_err(EvalError::OtherError)?
+        .branch_oid_to_names;
+
+    let branch_commits = make_pattern_matcher_for_set(
+        ctx,
+        name,
+        args,
+        Box::new(move |_repo, commit| {
+            let branches_at_commit = match branch_oid_to_names.get(&commit.get_oid()) {
+                Some(branches) => branches,
+                None => return Ok(false),
+            };
+
+            let result = branches_at_commit
+                .iter()
+                .filter_map(
+                    |branch_name| match CategorizedReferenceName::new(branch_name) {
+                        name @ CategorizedReferenceName::LocalBranch { .. } => {
+                            Some(name.render_suffix())
+                        }
+                        // we only care about local branches
+                        CategorizedReferenceName::RemoteBranch { .. }
+                        | CategorizedReferenceName::OtherRef { .. } => None,
+                    },
+                )
+                .any(|branch_name| pattern.matches_text(branch_name.as_str()));
+
+            Ok(result)
+        }),
+        Some(ctx.dag.branch_commits.clone()),
+    )?;
+
+    Ok(branch_commits)
 }
 
 #[instrument]
@@ -275,12 +316,24 @@ fn fn_stack(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
 
 type MatcherFn = dyn Fn(&Repo, &Commit) -> Result<bool, PatternError> + Sync + Send;
 
-#[instrument(skip(f))]
+/// Make a pattern matcher that operates on all visible commits.
 fn make_pattern_matcher(
     ctx: &mut Context,
     name: &str,
     args: &[Expr],
     f: Box<MatcherFn>,
+) -> Result<CommitSet, EvalError> {
+    make_pattern_matcher_for_set(ctx, name, args, f, None)
+}
+
+/// Make a pattern matcher that operates only on the given set of commits.
+#[instrument(skip(f))]
+fn make_pattern_matcher_for_set(
+    ctx: &mut Context,
+    name: &str,
+    args: &[Expr],
+    f: Box<MatcherFn>,
+    commits_to_match: Option<CommitSet>,
 ) -> Result<CommitSet, EvalError> {
     struct Matcher {
         expr: String,
@@ -301,7 +354,7 @@ fn make_pattern_matcher(
         expr: Expr::FunctionCall(Cow::Borrowed(name), args.to_vec()).to_string(),
         f,
     };
-    let matcher = make_pattern_matcher_set(ctx, ctx.repo, Box::new(matcher))?;
+    let matcher = make_pattern_matcher_set(ctx, ctx.repo, Box::new(matcher), commits_to_match)?;
     Ok(matcher)
 }
 
--- a/git-branchless-revset/src/builtins.rs
+++ b/git-branchless-revset/src/builtins.rs
@@ -61,6 +61,7 @@ lazy_static! {
             ("committer.date", &fn_committer_date),
             ("exactly", &fn_exactly),
             ("current", &fn_current),
+            ("merges", &fn_merges),
             ("tests.passed", &fn_tests_passed),
             ("tests.failed", &fn_tests_failed),
             ("tests.fixable", &fn_tests_fixable),
@@ -511,6 +512,19 @@ fn fn_current(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
         }
     }
     Ok(result.into_iter().collect::<CommitSet>())
+}
+
+#[instrument]
+fn fn_merges(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
+    eval0(ctx, name, args)?;
+    // Use a "pattern matcher" that – instead of testing for a pattern –
+    // examines the parent count of each commit to find merges.
+    make_pattern_matcher(
+        ctx,
+        name,
+        args,
+        Box::new(move |_repo, commit| Ok(commit.get_parent_count() > 1)),
+    )
 }
 
 fn read_all_test_results(repo: &Repo, commit: &Commit) -> Option<Vec<SerializedTestResult>> {
--- a/git-branchless-revset/src/pattern.rs
+++ b/git-branchless-revset/src/pattern.rs
@@ -135,12 +135,13 @@ pub(super) fn make_pattern_matcher_set(
     ctx: &mut Context,
     repo: &Repo,
     matcher: Box<dyn PatternMatcher>,
+    commits_to_match: Option<CommitSet>,
 ) -> Result<CommitSet, PatternError> {
     struct MatcherNameSetQuery {
         effects: Effects,
         matcher: Box<dyn PatternMatcher>,
         repo: Arc<Mutex<Repo>>,
-        visible_commits: CommitSet,
+        commits_to_match: CommitSet,
     }
 
     impl MatcherNameSetQuery {
@@ -152,10 +153,10 @@ pub(super) fn make_pattern_matcher_set(
                     )));
             let _effects = effects;
 
-            let len = self.visible_commits.count().await?;
+            let len = self.commits_to_match.count().await?;
             progress.notify_progress(0, len);
 
-            let stream = self.visible_commits.iter().await?;
+            let stream = self.commits_to_match.iter().await?;
             let commit_oids = stream.collect::<Vec<_>>().await;
             let repo = self.repo.lock().unwrap();
             let repo_pool = RepoResource::new_pool(&repo).map_err(make_dag_backend_error)?;
@@ -191,7 +192,7 @@ pub(super) fn make_pattern_matcher_set(
         }
 
         async fn contains(&self, name: &CommitVertex) -> eden_dag::Result<bool> {
-            if !self.visible_commits.contains(name).await? {
+            if !self.commits_to_match.contains(name).await? {
                 return Ok(false);
             }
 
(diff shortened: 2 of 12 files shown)


claytonrcarter:
> branches(<pattern>) needs more work; it's currently mush slower that I would expect, and it seems to not match some branchs (at least it didn't match all the branches I expected when I used it to push these recent PRs)

For example, on my copy of the `git-branchless` repo, I have approx 1600 commits and 7 branches, as follows:
```❯ git slb
⋮
◆ 5871499 2h (ᐅ master) Change 'suggest' to 'discuss'
┣━┓
┃ ◯ 2bd4827 5s (crc/submit-dry-run) feat(submit): add --dry-run flag
┣━┓
┃ ◯ 306e9b4 5s (crc/fix-status-spaces, local) fix(status): rudimentary support for filenames with spaces
┣━┓
┃ ◯ 9661224 5s (crc/sync-resolve-stack) feat(sync): resolve revset arguments to their entire stacks
┣━┓
┃ ⊘ 2 omitted commits
┃ ⋮
┃ ◯ 620a061 5s (move-fixup-on-disk) temp(on disk fixup): Try to fix abandoned branches
⋮
⊘ 1 omitted commit
⋮
◯ fa5342c 5s (crc/revsets) feat(revset): allow branches() to accept a text pattern
```

Running `git sl 'branches(crc)'` takes around 3s to run, and seems to process every commit in the repo. Also, `git submit 'branches(crc)' --dry-run` doesn't match/report the `crc/fix-status-spaces` branch, which I'm currently assuming has something to do w/ the other branch that points to that commit. (Hmm, on second thought, perhaps that's an issue w/ the implementation of #1129?)

claytonrcarter:
> Running `git sl 'branches(crc)'` takes around 3s to run, and seems to process every commit in the repo. 

This was my misunderstanding of how revset `Matchers` work. I hadn't connected that they work on all visible commits, and there is currently no way to restrict that. (This is the first revset feature – that I can tell – that would make any sense to operate on anything other than all visible commits, because we *know* that branches are (probably) a subset of visible commits.)

git-branchless-revset/src/builtins.rs
```diff
@@ -178,8 +180,57 @@ fn fn_heads(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+        Some(pattern) => pattern,
+        None => return Ok(ctx.dag.branch_commits.clone()),
+    };
+
+    let all_branch_commits: HashSet<_> = ctx
+        .dag
+        .commit_set_to_vec(&ctx.dag.branch_commits)
+        .map_err(EvalError::OtherError)?
+        .into_iter()
+        .collect();
```
claytonrcarter, line 193: I converted `branch_commits` from `NameSet` to a `HashSet` to get around some waaay over-my-head compiler errors about calling `.contains()` on a `NameSet` and how that's actually a `Pin<Box<dyn Future<...>>>`. In this case, I don't think we need any Eden DAG magic, but I'm open to suggestions if there's a better way to do this.

git-branchless-revset/src/builtins.rs
```diff
@@ -178,8 +180,57 @@ fn fn_heads(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+    let branch_commits = make_pattern_matcher(
+        ctx,
+        name,
+        args,
+        Box::new(move |repo: &Repo, commit: &Commit| {
+            if !all_branch_commits.contains(&commit.get_oid()) {
+                return Ok(false);
+            }
```
claytonrcarter, line 201: This is my workaround for not being able to restrict the commits that the `Matcher` will run on from the default of all visible commits to what we really need, which is just the branch commits. Refactoring `make_pattern_matcher()` etc felt like more that was needed for this limited use case.

Happy to reconsider if you think that would be useful, though.

git-branchless-revset/src/builtins.rs
```diff
@@ -178,8 +180,57 @@ fn fn_heads(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+                None => return Ok(false),
+            };
+
+            let result = branches_at_commit
+                .iter()
+                .filter_map(
+                    |branch_name| match CategorizedReferenceName::new(branch_name) {
+                        name @ CategorizedReferenceName::LocalBranch { .. } => {
+                            Some(name.render_suffix())
+                        }
```
claytonrcarter, line 221: I recall using `render_suffix()` on a different feature, perhaps in `hide`, but I also note that in `submit` we use `name.remove_prefix()` in a similar situation. I couldn't tell the difference off hand, so I went w/ what I used previously here.

git-branchless-revset/src/builtins.rs
```diff
@@ -513,6 +564,37 @@ fn fn_current(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+            let visible_heads = ctx
+                .dag
+                .query_visible_heads()
+                .map_err(EvalError::OtherError)?;
+            let visible_commits = ctx.dag.query_ancestors(visible_heads.clone())?;
+            ctx.dag
+                .filter_visible_commits(visible_commits)
+                .map_err(EvalError::OtherError)?
+        }
+    };
```
claytonrcarter, line 582: I chose to allow `merges()` to accept an arg because it seemed like a possible optimization. (Please correct me if I'm misunderstanding how revsets are evaluated.) My thinking is that `merges() & draft()` would have to calculate all visible merges and then intersect that w/ draft commits, while `merges(draft())` would be able to use the imput (ie draft commits) instead of all visible.

This differs from several other revset fns, though, that operate only on all visible commits w/o this sort of arg-optimization. Open to input, and happy to revert to a simpler only-visible commit version, if you prefer.

claytonrcarter: FWIW this may not be a worthwhile optimization. I just played around in my work repo (11k commits including 2.8k merges) and got the following results:
```sh
❯ time git query 'merges() & date("after:1 year ago")' | wc -l
     571

________________________________________________________
Executed in    4.79 secs    fish           external
   usr time    3.87 secs    0.48 millis    3.87 secs
   sys time    4.06 secs    2.81 millis    4.06 secs


❯ time git query 'merges(date("after:1 year ago"))' | wc -l
     571

________________________________________________________
Executed in    4.07 secs    fish           external
   usr time    3.43 secs    0.45 millis    3.43 secs
   sys time    3.62 secs    2.77 millis    3.62 secs
```

So it's 10-15% faster in this case, but then a much smaller set of commits (`merges(draft())` vs `merges() & draft()`) yielded no speed up. So I guess I'm leaning towards simplifying this implementation and omitting this "optimization". What do you think?

git-branchless-revset/src/builtins.rs
arxanas: > This was my misunderstanding of how revset Matchers work. I hadn't connected that they work on all visible commits, and there is currently no way to restrict that. (This is the first revset feature – that I can tell – that would make any sense to operate on anything other than all visible commits, because we know that branches are (probably) a subset of visible commits.)

Take a look at `CommitSet::from_evaluate_contains`/`CommitSet::from_async_evaluate_contains`. For the branches case, we want to provide separate paths for `evaluate` vs `contains`. The `Matcher` stuff is just for the case where we want to provide only `contains` and define `evaluate` to mean "run `contains` on all commits in the repo".

In principle, if we provide `Hints::FILTER`, then the runtime library should know how to optimize e.g. `merges() & expr()` so that we can try to evaluate `expr()` first if it seems to be more efficient to do so, but I don't know if/how this actually happens in practice. If the library doesn't do that, then perhaps we should do the check ourselves in our intersection implementation and swap the operands. Possibly something like this —

```rust
#[instrument]
fn fn_intersection(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
    let (lhs, rhs) = eval2(ctx, name, args)?;
    let (lhs, rhs) = if rhs.hints().contains(Flags::FILTER) && !lhs.hints().contains(Flags::FILTER)
    {
        (rhs, lhs)
    } else {
        (lhs, rhs)
    };
    Ok(lhs.intersection(&rhs))
}
```

As written, I think that this pessimizes the zero-argument case of `branches()`, because now we have to traverse all visible commits to see if each one points to a branch. We should be able to get both contains and evaluate paths to be efficient.

git-branchless-revset/src/builtins.rs
arxanas: Hmm... it looks like the only difference is that `remove_prefix` returns `Ok(...)` (unconditionally). They were probably different in the path, but we could remove `remove_prefix` now. (There was a time when we encoded references as `OsString`s instead of `String`s, but I dropped that feature, so some fallible code paths are no longer necessary).

git-branchless-revset/src/builtins.rs
arxanas: I would prefer to keep the zero-argument version only (and operate on all merges, not just draft commits) and try to optimize as per the above discussion on `contains`/`evaluate`.

arxanas (changes requested):

git-branchless-revset/src/builtins.rs
claytonrcarter: > keep the zero-argument version only

Done!

git-branchless-revset/src/builtins.rs
claytonrcarter: I went ahead and refactored these in this PR

git-branchless-revset/src/builtins.rs
claytonrcarter: > pessimizes the zero-argument case of branches(),

I'm probably misunderstanding you, but I think that the zero-argument case is nearly identical to the same case on `master`: it just immediately returns `ctx.dag.brance_commits`. When an arg is passed, though, then – yes – it does have to traverse all visible commits to find those at matching branches.

> Take a look at CommitSet::from_evaluate_contains/CommitSet::from_async_evaluate_contains

OK, cool, but I'm still having a hard time pulling apart how to best make use of them. For now, I've made the following updates:
- `merges()` now uses the pattern matcher code (w/o a pattern matcher, though) to leverage the `evalulate`/`contains` paths w/ a very simple filtering fn. This revset is now basically just 1 line of code.
- `branches()` I removed the hacky `HashSet` thing I was doing and instead altered the API of `make_pattern_matcher_set()` so that it can (optionally) take a set of commits to traverse instead of just "all visible commits"
  - this made sense because we already know the commits we need to examine, via `dag.branch_commits`
  - another option I considered was to use `repo.get_all_local_branches()`, and then using the `pattern` to filter by name, finally resolving the branches to a set via `branch.get_oid()` ... this seemed like it may be less work on our end (more direct), but I didn't explore it because the current implementation is at least working 😄 

But I may also just be misunderstanding what you're suggesting! Sorry, if so.

> For the branches case, we want to provide separate paths for evaluate vs contains.

I wonder if my change (to pass the branch commits into `make_pattern_matcher_set()`) makes this less of an issue, because now the `evaluate()` path will not operate on all visible commits? If I'm misunderstanding what you're asking for, please let me know and I'll dig in again.

> If the library doesn't do that, then perhaps we should do the check ourselves in our intersection implementation and swap the operands

I checked briefly and it was unclear if the upstream is doing this level of optimization (though I'd be a bit surprised if that's the case). They seem mostly concerned if either side is `FULL` or `EMPTY`, but I did not dig too deeply. (I found [this](https://github.com/facebook/sapling/blob/fa6824a3add28a36d7860dc91d8534f08b3941d8/eden/scm/lib/dag/src/nameset/mod.rs#L228) and [this](https://github.com/facebook/sapling/blob/main/eden/scm/lib/dag/src/nameset/intersection.rs#L92); the latter suggests that they prefer to keep `FILTER` sets on the rhs?)

Regardless, I can look at this, but I think I would prefer to do so in a separate PR, unless you think the perf of this PR is really not good. I may need a suggestion on how to benchmark that, though, as most of the repos I have access to are pretty fast already! 😆

claytonrcarter: > another option I considered was to use repo.get_all_local_branches(), and then using the pattern to filter by name, finally resolving the branches to a set via branch.get_oid()

FWIW I did end up playing with this, and this seems to work, and with a lot less code!

```rust
fn fn_branches(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
    let pattern = match eval0_or_1_pattern(ctx, name, args)? {
        Some(pattern) => pattern,
        None => return Ok(ctx.dag.branch_commits.clone()),
    };

    let branch_commits = ctx
        .repo
        .get_all_local_branches()?
        .iter()
        .filter(|branch| pattern.matches_text(branch.get_name().expect("TODO")))
        .filter_map(|branch| branch.get_oid().expect("TODO never happens"))
        .collect();

    Ok(branch_commits)
}
```

claytonrcarter:
Sorry, I lost track of this. As I recall, I have addressed the review as best I could figure out; see my comments from 11/13. I think I just forgot to un-draft this after doing so. Thank you!

git-branchless-revset/src/builtins.rs
```diff
@@ -275,12 +316,24 @@ fn fn_stack(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
     ctx: &mut Context,
     name: &str,
     args: &[Expr],
     f: Box<MatcherFn>,
+) -> Result<CommitSet, EvalError> {
+    make_pattern_matcher_for_set(ctx, name, args, f, None)
+}
+
+#[instrument(skip(f))]
+/// Make a pattern matcher that operates only on the given set of commits.
```
arxanas, line 330: nit: reorder these two lines so that the doc-comment comes before any annotations.

git-branchless-revset/src/builtins.rs
```diff
@@ -513,6 +567,19 @@ fn fn_current(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+#[instrument]
+fn fn_merges(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
+    eval0(ctx, name, args)?;
+    // "pattern matchers" normally operate with a pattern, but we can make use
+    // of this system to look for merge commits w/o the use of one
+    make_pattern_matcher(
+        ctx,
+        name,
+        args,
+        Box::new(move |_repo, commit| Ok(commit.get_parents().len() > 1)),
```
arxanas, line 579: issue: Use `commit.get_parent_count()` here. If you use `get_parents()` directly, then we have to actually load each parent object from disk, even though we know the parent count from just looking at the current commit.

git-branchless-revset/src/builtins.rs
```diff
@@ -178,8 +180,48 @@ fn fn_heads(ctx: &mut Context, name: &str, args: &[Expr]) -> EvalResult {
...
+        Some(pattern) => pattern,
+        None => return Ok(ctx.dag.branch_commits.clone()),
+    };
+
+    let branch_commits = make_pattern_matcher_for_set(
+        ctx,
+        name,
+        args,
+        Box::new(move |repo, commit| {
+            let branch_oid_to_names = match repo.get_references_snapshot() {
```
arxanas, line 193: issue: I'm concerned that this means that we'll try to read all references on the repo once for every commit. Instead, can you get the references snapshot just once outside of this closure (and implicitly `move` it in)?

git-branchless-revset/src/builtins.rs
arxanas: I don't see this in the latest code, but there exists a `Dag::set_contains` method that wraps the `Future` for you that you could use in this case.

git-branchless-revset/src/builtins.rs
arxanas: I think specifying the universe of commits in the pattern matcher should work performantly.

git-branchless-revset/src/pattern.rs
```diff
@@ -135,6 +135,7 @@ pub(super) fn make_pattern_matcher_set(
     ctx: &mut Context,
     repo: &Repo,
     matcher: Box<dyn PatternMatcher>,
+    commits_to_match: Option<CommitSet>,
```
arxanas, line 138: suggestion: We should rename the fields/variables `visible_commits` to `commits_to_match` in `MatcherNameSetQuery` and the rest of this function.

arxanas (approved):
Looks good pending a few comments!

git-branchless-revset/src/builtins.rs
claytonrcarter: Oops, done!

git-branchless-revset/src/builtins.rs
claytonrcarter: Hmm, yes. I think I was concerned that `repo` was being passed to the closure to handle some advanced usage where we may be working with commits from different repos, but that doesn't make sense in hindsight, and extracting it makes the code a lot cleaner too. 👍

git-branchless-revset/src/pattern.rs
claytonrcarter: Sure thing. Done!

claytonrcarter:
> Looks good pending a few comments!

Great, thank you!

PR #1150, opened 2023-12-14
feat(record): add --stash flag

YakoYakoYokuYoku:
Because I need to stash things and `git stash` falls short I've created this PR. Fixes #979.

### DIFF
--- a/git-branchless-opts/src/lib.rs
+++ b/git-branchless-opts/src/lib.rs
@@ -320,6 +320,10 @@ pub struct RecordArgs {
     /// if any.
     #[clap(action, short = 'I', long = "insert")]
     pub insert: bool,
+
+    /// After making the new commit, switch back to the previous commit.
+    #[clap(action, short = 's', long = "stash", conflicts_with_all(&["create", "detach"]))]
+    pub stash: bool,
 }
 
 /// Display a nice graph of the commits you've recently worked on.
--- a/git-branchless-record/src/lib.rs
+++ b/git-branchless-record/src/lib.rs
@@ -18,7 +18,7 @@ use git_branchless_invoke::CommandContext;
 use git_branchless_opts::RecordArgs;
 use git_branchless_reword::edit_message;
 use itertools::Itertools;
-use lib::core::check_out::{check_out_commit, CheckOutCommitOptions};
+use lib::core::check_out::{check_out_commit, CheckOutCommitOptions, CheckoutTarget};
 use lib::core::config::{get_commit_template, get_restack_preserve_timestamps};
 use lib::core::dag::{CommitSet, Dag};
 use lib::core::effects::{Effects, OperationType};
@@ -57,6 +57,7 @@ pub fn command_main(ctx: CommandContext, args: RecordArgs) -> EyreExitOr<()> {
         create,
         detach,
         insert,
+        stash,
     } = args;
     record(
         &effects,
@@ -66,6 +67,7 @@ pub fn command_main(ctx: CommandContext, args: RecordArgs) -> EyreExitOr<()> {
         create,
         detach,
         insert,
+        stash,
     )
 }
 
@@ -78,6 +80,7 @@ fn record(
     branch_name: Option<String>,
     detach: bool,
     insert: bool,
+    stash: bool,
 ) -> EyreExitOr<()> {
     let now = SystemTime::now();
     let repo = Repo::from_dir(&git_run_info.working_directory)?;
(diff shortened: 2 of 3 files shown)


arxanas (changes requested):
Thanks! Could you add a test to `git-branchless-record/tests/test_record.rs`? It looks like we would want to verify the behavior both with a branch checked out and without one.

git-branchless-record/tests/test_record.rs
```diff
@@ -279,6 +279,66 @@ fn test_record_detach() -> eyre::Result<()> {
     Ok(())
 }
 
+#[test]
+fn test_record_stash() -> eyre::Result<()> {
```
arxanas, line 283: Test looks good, thanks!

git-branchless-record/src/lib.rs
```diff
@@ -199,12 +202,33 @@ fn record(
...
-                    )
+                    )?);
                 }
                 parent_commits => {
                     eyre::bail!("git-branchless record --detach called on a merge commit, but it should only be capable of creating zero- or one-parent commits. Parents: {parent_commits:?}");
                 }
-            };
+            }
+        }
+        if stash && (head_info.oid.is_some() || head_info.reference_name.is_some()) {
```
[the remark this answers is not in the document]
YakoYakoYokuYoku: Yeah, I couldn't decide on a better option at the moment, I'll do something among what you've suggested.

git-branchless-record/src/lib.rs
```diff
@@ -199,7 +202,23 @@ fn record(
...
-                    )
+                    );
+                    if stash {
+                        try_exit_code!(check_out_commit(
+                            effects,
+                            git_run_info,
+                            &repo,
+                            &event_log_db,
+                            event_tx_id,
+                            Some(CheckoutTarget::Reference(reference_name.clone())),
```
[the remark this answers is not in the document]
YakoYakoYokuYoku: Now I've modified it to not make the early return, but I cannot picture how a `--detach --insert` or a `--stash --insert` would work. If you happen to know, could you please explain it?

git-branchless-opts/src/lib.rs
```diff
@@ -320,6 +320,10 @@ pub struct RecordArgs {
     /// if any.
     #[clap(action, short = 'I', long = "insert")]
     pub insert: bool,
+
+    /// Detach the current branch before committing and then reattach.
```
arxanas, line 324: Since there may not be a branch explicitly involved, let's rewrite this, maybe something like

```suggestion
    /// After making the new commit, switch back to the previous commit.
```

arxanas (approved):
Looks good, thanks for implementing! I just had one comment on the doc-string.

It looks like linting/static analysis is failing (probably due to an upgrade in Clippy nightly). I'll fix that on `master`, then you can rebase on top of it and land this commit.

git-branchless-record/src/lib.rs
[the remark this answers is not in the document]
arxanas: I think it's fine to not support these for now, but I think there are reasonable semantics:

- If you are on e.g. `main` and want to create a new commit without committing to `main`, then you could use `--detach` (or `--create` with a new branch name). If you want to rebase all of your other draft work based on main onto this branch, then you could pass `--insert` as well to rebase all descendants of `main` onto the new commit. I sometimes do this if there's a build or test fix that I need to use in all of my local stacks.
- I don't imagine `--stash --insert` would be that common, but I guess if you make a change and you want to apply it to already-stashed work (which are already descendants of the current commit), then you could do `--stash --insert`. It would oftentimes be equivalent to `--stash` and then `move` it on top of the already-stashed work.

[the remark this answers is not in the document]
YakoYakoYokuYoku: Got it! Thanks for the explanation.

git-branchless-opts/src/lib.rs
YakoYakoYokuYoku: You are right. Done it.

git-branchless-record/tests/test_record.rs
```diff
@@ -279,6 +279,66 @@ fn test_record_detach() -> eyre::Result<()> {
...
+    let git = make_git()?;
+
+    if !git.supports_reference_transactions()? {
+        return Ok(());
+    }
+    git.init_repo_with_options(&GitInitOptions {
+        make_initial_commit: false,
+        run_branchless_init: false,
+    })?;
+    git.branchless("init", &["--main-branch", "master"])?;
```
arxanas, line 293: It looks like the test is failing now. You probably want to just initialize the repository normally? Currently, if no initial commit is made, then the `master` branch can't be found.

YakoYakoYokuYoku: It worked on my machine, I don't exactly know why, but now I made it to include the root commit.

PR #1169, opened 2023-12-26
feat(record): allow multiple --message args

claytonrcarter:
This is a minor QOL improvement, and should bring `record` in line with how `commit` and `reword` accept multiple `-m/--message` args, turning them into multiple paragraphs in the commit message.

### DIFF
--- a/git-branchless-record/src/lib.rs
+++ b/git-branchless-record/src/lib.rs
@@ -52,7 +52,7 @@ pub fn command_main(ctx: CommandContext, args: RecordArgs) -> EyreExitOr<()> {
         git_run_info,
     } = ctx;
     let RecordArgs {
-        message,
+        messages,
         interactive,
         create,
         detach,
@@ -62,7 +62,7 @@ pub fn command_main(ctx: CommandContext, args: RecordArgs) -> EyreExitOr<()> {
     record(
         &effects,
         &git_run_info,
-        message,
+        messages,
         interactive,
         create,
         detach,
@@ -75,7 +75,7 @@ pub fn command_main(ctx: CommandContext, args: RecordArgs) -> EyreExitOr<()> {
 fn record(
     effects: &Effects,
     git_run_info: &GitRunInfo,
-    message: Option<String>,
+    messages: Vec<String>,
     interactive: bool,
     branch_name: Option<String>,
     detach: bool,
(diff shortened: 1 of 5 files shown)


git-branchless-record/tests/test_record.rs
```diff
@@ -13,22 +14,31 @@ fn test_record_unstaged_changes() -> eyre::Result<()> {
...
         let (stdout, _stderr) = git.run(&["show"])?;
-        insta::assert_snapshot!(stdout, @r###"
-        commit 914812ae3220add483f11d851dc59f0b5dbdeaa0
+        // For the empty lines between paragraphs, git show prints 4 spaces
+        // instead of just an empty line. (ie "    \n" instead of "\n") It's
+        // hard to work with this in an inline snapshot, as editors often trim
+        // trailing whitespace, so here we just modify the output (by replacing
+        // any such lines) to make it easier to assert against with insta
+        let whitespace_only_line_regex = Regex::new(r"(?m)^ {4}$").unwrap();
+        let trimmed_stdout = whitespace_only_line_regex.replace_all(&stdout, "");
```
claytonrcarter, line 32: This feels wrong, but I couldn't figure out how else to do it. I looked around a bit and didn't see anywhere else that we check snapshots for multiparagraph commit messages from `git show`. In `test_reword`, we look at `git log` output, which doesn't seem to suffer the same behavior.

git-branchless-record/src/lib.rs
```diff
@@ -150,15 +150,15 @@ fn record(
...
         }
     } else {
         let args = {
             let mut args = vec!["commit"];
-            if let Some(message) = &message {
-                args.extend(["--message", message]);
-            }
+            messages
+                .iter()
+                .for_each(|message| args.extend(["--message", message]));
```
arxanas, line 161: I think we can write it something like this:

```suggestion
            messages.extend(messages.iter().flat_map(|message| ["--message", message]));
```

git-branchless-record/tests/test_record.rs
arxanas: This is fine, and ends up being necessary for snapshot testing in various cases. Can you actually move it into this function?

https://github.com/arxanas/git-branchless/blob/81e298dad0bd0335c4d330816c97fab38a24524a/git-branchless-lib/src/testing.rs#L96-L133

arxanas (approved):
Thanks!

git-branchless-record/src/lib.rs
claytonrcarter: Oh, of course. Thanks for spotting that! Fixed.

git-branchless-record/tests/test_record.rs
claytonrcarter: Yes, done. I adjusted the comment now that it's more general and less specific to that one test, and adjusted the code slightly to match the style of the other regex in that fn.

PR #1203, opened 2024-01-29
cleanup(reword): Remove redundant reassignment in `edit_message`

mlcui-google:
Both arms of the match results in a `&mut Editor` pointing to the same editor, which is redundant.

### DIFF
--- a/git-branchless-reword/src/lib.rs
+++ b/git-branchless-reword/src/lib.rs
@@ -61,10 +61,13 @@ pub enum InitialCommitMessages {
 /// Open the user's configured commit editor seeded with the provided message.
 #[instrument]
 pub fn edit_message(git_run_info: &GitRunInfo, repo: &Repo, message: &str) -> eyre::Result<String> {
-    let mut editor = Editor::new();
-    let (editor, editor_program) = match get_editor(git_run_info, repo)? {
-        Some(editor_program) => (editor.executable(&editor_program), editor_program),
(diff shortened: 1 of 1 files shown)


git-branchless-reword/src/lib.rs
```diff
@@ -62,9 +62,12 @@ pub enum InitialCommitMessages {
 #[instrument]
 pub fn edit_message(git_run_info: &GitRunInfo, repo: &Repo, message: &str) -> eyre::Result<String> {
     let mut editor = Editor::new();
-    let (editor, editor_program) = match get_editor(git_run_info, repo)? {
-        Some(editor_program) => (editor.executable(&editor_program), editor_program),
-        None => (&mut editor, "<default>".into()),
+    let editor_program = match get_editor(git_run_info, repo)? {
```
arxanas, line 65: As written after the change, this `match` expression both returns a value and has mutating side-effects, which can be surprising for further readers. I would prefer to keep this as is, or maybe move the `Editor` instantiation inside both cases of the `match` expression.

arxanas (changes requested):

git-branchless-reword/src/lib.rs
mlcui-google: Done - PTAL. I think b549e51 is around as verbose as the original code, but reads better (at least to me).

arxanas: I like it better too, thanks!

PR #1204, opened 2024-01-29
fix(reword): Wrap rewording instructions to 72 chars

mlcui-google:
The first line will be 71 chars if there is only one commit, and >=74 chars if there are multiple commits - which is close enough to 72, and avoids splitting up "these 2 commits" into its own line.


arxanas (approved):
Thanks!

PR #1244, opened 2024-02-14
feat(submit): support multiple revset arguments

claytonrcarter:
I have a bunch of stale branches that I don't want to think about while I work on newer stuff, so I often find myself running `git sync branch-1 branch-2 ...` and then `git submit branch-1 branch-2 ...` ... only for `submit` to tell me that it doesn't support multiple args. 😢  I usually resort to concating them to `branch-1+branch-2+...`, but I would prefer `submit` to do that for me, so here we are. 😄

### DIFF
--- a/git-branchless-submit/src/lib.rs
+++ b/git-branchless-submit/src/lib.rs
@@ -24,7 +24,7 @@ use git_branchless_test::{RawTestOptions, ResolvedTestOptions, Verbosity};
 use github::GithubForge;
 use itertools::Itertools;
 use lazy_static::lazy_static;
-use lib::core::dag::{CommitSet, Dag};
+use lib::core::dag::{union_all, CommitSet, Dag};
 use lib::core::effects::Effects;
 use lib::core::eventlog::{EventLogDb, EventReplayer};
 use lib::core::formatting::{Pluralize, StyledStringBuilder};
@@ -188,7 +188,7 @@ pub fn command_main(ctx: CommandContext, args: SubmitArgs) -> EyreExitOr<()> {
         create,
         draft,
         strategy,
-        revset,
+        revsets,
         resolve_revset_options,
         forge,
         message,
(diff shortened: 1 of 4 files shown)


git-branchless-submit/src/lib.rs
```diff
@@ -294,14 +289,15 @@ fn submit(
         message,
     };
 
+    let unioned_revset = Revset(revsets.iter().cloned().map(|r| r.0).join("+"));
```
claytonrcarter, line 292: This is the only part that feels iffy to me. It's just joining the vec of revset stings (eg `["a", "b", "c"]`) into a "unioned" revset like `Revset("a+b+c")`. I *did not* dig into how the forges use this, except to see that it's only in use by the Phabricator forge, but semantically I think it makes sense since we want "all of these revsets".

It does feel a little specific or procedural, and I wondered if adding an `impl From<Vec<Revset>> for Revset` may be a cleaner approach. Open to feedback, as always.

arxanas: It looks like the Phabricator forge passes it to `run_tests`, which only uses it for display purposes, so I guess this is fine.

nit: We don't need to `clone` the internal strings (and it doesn't help us because we need to allocate an entirely new `String` as part of `join`). So you just need to use a form that gets a reference to the inner value. You can do `&r.0` in your current code, or as below, as it happens to transfer the ref-ness to the destructured members (I like to destructure tuples/structs where possible to check exhaustiveness):

```suggestion
    let unioned_revset = Revset(revsets.iter().map(|Revset(inner)| inner).join(" + "));
```

suggestion: My preference is for a space around `+`, but I don't feel strongly.

arxanas (approved):
Small nits but looks good, thanks!

git-branchless-submit/src/lib.rs
claytonrcarter: Updated all as suggested, thank you!

> don't need to clone the internal strings

Gotcha, thanks. I guess I hacked something into place and then totally forgot to double check it once the compiler accepted it. Updated as suggested; pattern matching FTW!

PR #1246, opened 2024-02-21
feat(nix): improve flake.nix, follows nixpkgs

bryango:
- Flatten the flake outputs for clarity, and use updated flake outputs instead of the [deprecated ones](https://nixos.org/manual/nix/unstable/release-notes/rl-2.7.html):
	- `packages.${system}.default` instead of `defaultPackage`
	- `overlays.default` instead of `overlay`.
- Reuse nixpkgs definition for git-branchless, and supply local changes through overrides. This ensures that nixpkgs updates such as https://github.com/NixOS/nixpkgs/pull/289324 are automatically propagated to the package definition here.
- Reuse git-branchless packaging for scm-diff-editor, since their package definitions are almost identical.
- Bump nixpkgs flake.lock. With the above changes in place this automatically pulls in the nixpkgs update https://github.com/NixOS/nixpkgs/pull/289324.


flake.nix
```diff
@@ -11,126 +11,65 @@
...
-                "--skip=test_switch_pty"
-                "--skip=test_next_ambiguous_interactive"
-                "--skip=test_switch_auto_switch_interactive"
-              ];
-            }
-          )
-          {
-            inherit (final.darwin.apple_sdk.frameworks) Security SystemConfiguration;
+        # reuse the definition from nixpkgs git-branchless
+        git-branchless = prev.git-branchless.overrideAttrs ({ meta, ... }: {
```
hmenke, line 20: Back when I contributed the Flake I consciously decided against this approach because it will be cumbersome to add extra dependencies or disable test that fail in the Nix sandbox. However, if you believe that the package definition has converged sufficiently, please don't feel any obligation.

hmenke (approved):

flake.nix
bryango: I also have this concern! Though I do think that currently things have stabilized enough to allow us to do this. I would like to contribute to both nixpkgs and git-branchless in the long term (this is really an awesome project), and if things got complicated again, it would then be better to bring back the fully local package definition. But for now, I think overrides are good enough, and this eliminates the need to port things back and forth (between nixpkgs and flake.nix here).

bryango:
New commit: added some comments to direct people to contribute to nixpkgs for the package definition.

flake.nix
bryango: I will mark this as resolved for now, but if you have any other comments, please feel free to unresolve any time!

arxanas:
I can't read Nix so I'll just trust @hmenke and CI 😇

PR #1258, opened 2024-03-02
cleanup: rm dead code for (feature = "man-pages") & recycle for a new (feature = "separate-binaries")

bryango:
Hi! I was exploring the clap related code to see if we can generate completions, and accidentally came across some dead code on the removed "man-pages" feature. This patch:
- Removes dead code for (feature = "man-pages"), which has been deleted since #205. This avoids confusion with the current working implementation, namely the "install-man-pages" command.
- Renames some probably useful (feature = "man-pages") to an unimplemented (feature = "separate-binaries") which is more descriptive. This is unimplemented for now, but it could be revived in the future so it would probably be useful to keep the code?
- Adds nix devShells for the development environment and bumps nixpkgs.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1708373374,
-        "narHash": "sha256-yEyDvDj/YQc4GOZa/cXx6YUzexD8uv1rUvD6SJVr6UI=",
+        "lastModified": 1728249353,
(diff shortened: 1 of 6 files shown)


bryango:
This PR is no longer needed as the man-pages feature is effectively resurrected in 7e55f2afc58b3d1fe78ecff3adb59cff95bf108b :laughing: Although a full resurrection may require fully reverting d235fc7abaca116437cd198217e4d53bfd638760. Closing...

PR #1276, opened 2024-03-13
fix: install manpages during Nix installation

hauleth:


bryango:
This is no longer necessary as the man page installation has been [upstreamed](https://github.com/NixOS/nixpkgs/commit/d5a849cb0fc23af09f1a668e46d20510cf446884) (by me haha), so one can simply update `flake.lock` and the man pages should get installed automatically.

hauleth:
Yeah, but upstream fix will not fix cases, when I want to use current development tree, as for example when I want to use `git branchless submit` with GitHub PRs.

bryango:
The `flake.nix` here uses nixpkgs's packaging functions (including `postInstall`), but builds on top of the current git master:

https://github.com/arxanas/git-branchless/blob/3faa0744cf0b0c7c2fe2bf20e5f441776f21c6dd/flake.nix#L18-L26

So we should get the best of both worlds. Please try `nix flake lock --update-input nixpkgs` and it should just work. Otherwise there might be a bug...

arxanas:
Is this PR still necessary given the above? I don't use Nix, so I can't tell myself.

hauleth:
No it is not needed anymore

PR #1315, opened 2024-05-20
smartlog: replace `--reverse` flag with config opt

e-q:
My perspective, based on my own experience, is that folks generally stick to one log orientation/direction. In principle, the reverse flag can be incorporated by default via an alias, but this doesn’t catch the auto-smartlog that is emitted after some operations. 

This PR adds a Boolean configuration option that replaces the —reverse flag: branchless.smartlog.reverse

This is a breaking CLI change, so I don’t imagine this would incorporated as-is. However, I didn’t want to go too far down this path without considering your thoughts and opinions on how a configuration option may be incorporated. (Not to mention my very limited Rust experience 😅)

What do you think? Thanks for all your work on such a helpful tool!

### DIFF
--- a/git-branchless-lib/src/core/config.rs
+++ b/git-branchless-lib/src/core/config.rs
@@ -108,6 +108,13 @@ pub fn get_smartlog_default_revset(repo: &Repo) -> eyre::Result<String> {
         })
 }
 
+/// Whether to reverse the smartlog direction by default
+#[instrument]
+pub fn get_smartlog_reverse(repo: &Repo) -> eyre::Result<bool> {
+    repo.get_readonly_config()?
+        .get_or("branchless.smartlog.reverse", false)
+}
+
 /// Get the default comment character.
 #[instrument]
 pub fn get_comment_char(repo: &Repo) -> eyre::Result<char> {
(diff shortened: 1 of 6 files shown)


arxanas:
A configuration option makes sense to me, since I've never once wanted to enable it for git-branchless and never once wanted to disable the equivalent option for jj 😅.

The git-branchless-smartlog output isn't really meant to be machine parseable (you can use git-branchless-query for that), so I'm optimistic that nobody's workflow would be broken by this. I posted in [git-branchless Discord](https://discord.com/channels/915309546984050709/915309546984050712/1244388541698408498) and [jj Discord](https://discord.com/channels/968932220549103686/968932220549103689/1244389100065390633) as well to see if anyone had any thoughts.

If we're worried about it, we could make the command-line flag deprecated (print a warning when used) but continue to support it.

kevinmost:
Hi! Any blockers to merging this? I was just writing up a discussion topic requesting this exact config flag. :)

duanyutong:
Any chance a new release can be made to pick up these recent useful updates? The last release was 18 months ago.

claytonrcarter:
> Any chance a new release can be made to pick up these recent useful updates? The last release was 18 months ago.

@duanyutong FYI this is now released in v0.11! https://github.com/arxanas/git-branchless/releases/tag/v0.11.0

Issue #1321, opened 2024-05-22
Error message:  "Could not parse OID from string ref:refs/heads/main"


jblebrun:
### Description of the bug

I've started to see an error appearing when I navigate between branches in git branchless:

```
2024-05-22T18:25:56.228109Z ERROR git_branchless_hook: Could not parse reference-transaction-line err=
   0: Could not parse OID from string ref:refs/heads/main
   1: Git error GenericError: unable to parse OID - contains invalid characters

Location:
   git-branchless-lib/src/git/repo.rs:289

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: git_branchless_hook::reference_transaction::parse_reference_transaction_line with line="0000000000000000000000000000000000000000 ref:refs/heads/main HEAD"
      at git-branchless-hook/src/lib.rs:289
   1: git_branchless_hook::hook_reference_transaction with effects=<Output fancy=false> transaction_state="committed"
      at git-branchless-hook/src/lib.rs:418
   2: git_branchless_hook::command_main with ctx=CommandContext { effects: <Output fancy=false>, git_run_info: <GitRunInfo path_to_git="git" working_directory="/usr/local/google/home/jibbl/Projects/oak" env=not shown> } args=HookArgs { subcommand: ReferenceTransaction { transaction_state: "committed" } }
      at git-branchless-hook/src/lib.rs:527
```

It seems that it somehow gets the value `ref:refs/heads/main` in a position where a commit hash is expected. 

I'm still digging into this more, but figured I'd file a bug in case anyone else has any hunches about what might be happening. 
So far, all I can say is that I can't find any entry in the SQLite db that contains `ref:refs/heads/main` in that position; so I'm not sure where that's being introduced.

It seems that it's related to the state of the repository; even when I check out a fresh copy of the repository, this is happening.  It just started happening yesterday or today, so perhaps it's somehow related to some commit metadata? 

Functionally, everything still seems to work OK (it still switches tot he branch, for example).

### Expected behavior

I expect the message not to appear.

### Actual behavior

The message appears.

### Version of `git-branchless`

0.8.0

### Version of `git`

 2.45.1.288.patched (internal company version)

jblebrun:
It looks like this may be related to this recent change in git: 

[...]

PR #1322, opened 2024-05-22
Fix for Bug #1321 - Support symbolic refs in tx lines for git 2.45

jblebrun:
In git 2.45, reference transactions now support symbolic refs (see:
https://github.com/git/git/commit/a8ae923f85da6434c3faf9c39719d6d5e5c77e65)

To support this, we add a transaction reference resolver that's capable of
looking up the Oid for a named reference. It requires a repository object to
do this successfully.

To support the functionality, a new `refname_to_id` function is added to the
Repo object that calls the similarly named method on the inner repository to
get the `git2::Oid` for a given reference name.

### DIFF
--- a/git-branchless-hook/src/lib.rs
+++ b/git-branchless-hook/src/lib.rs
@@ -197,6 +197,56 @@ mod reference_transaction {
 
     use lib::git::{MaybeZeroOid, ReferenceName, Repo};
 
+    /// A reference target parsed from a reference line.
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub enum ReferenceTarget {
+        /// A reference target that was in the transaction line as a normal commit hash.
+        Direct { oid: MaybeZeroOid },
+
+        /// A reference target that was in the transaction line as a symbolic
+        /// reference: `ref:<refname>`.
+        Symbolic { name: ReferenceName },
+    }
+
+    impl ReferenceTarget {
+        /// Attempt to convert the provided reference target into an OID hash value.
+        ///
+        /// For [ReferenceTarget::Direct] types, this always succeeds, and just
+        /// returns the wrapped OID.
+        ///
+        /// For [ReferenceTarget::Symbolic] types, this attempts to convert the
+        /// provided symbolic ref name into an OID hash using the provided
+        /// [Repo].
+        #[instrument]
+        pub fn as_oid(&self, repo: &Repo) -> eyre::Result<MaybeZeroOid> {
+            match self {
+                ReferenceTarget::Direct { oid } => Ok(*oid),
+                ReferenceTarget::Symbolic { name } => Ok(repo.reference_name_to_oid(name)?),
+            }
+        }
+    }
+
+    impl FromStr for ReferenceTarget {
+        type Err = eyre::ErrReport;
+
+        /// Attempts to parse a string as a [ReferenceTarget].
+        /// The string is expected to be one field from a reference transaction
+        /// line, which can be one of the following values:
+        ///
+        /// ref:<symbolic ref name>
+        /// <commit hash>
+        fn from_str(value: &str) -> Result<Self, Self::Err> {
+            match value.strip_prefix("ref:") {
+                Some(refname) => Ok(ReferenceTarget::Symbolic {
+                    name: ReferenceName::from(refname),
+                }),
+                None => Ok(ReferenceTarget::Direct {
+                    oid: value.parse()?,
+                }),
+            }
+        }
+    }
+
     #[instrument]
     fn parse_packed_refs_line(line: &str) -> Option<(ReferenceName, MaybeZeroOid)> {
         if line.is_empty() {
@@ -282,8 +332,8 @@ mod reference_transaction {
     #[derive(Debug, PartialEq, Eq)]
     pub struct ParsedReferenceTransactionLine {
         pub ref_name: ReferenceName,
-        pub old_oid: MaybeZeroOid,
-        pub new_oid: MaybeZeroOid,
+        pub old_value: ReferenceTarget,
+        pub new_value: ReferenceTarget,
     }
 
     #[instrument]
@@ -294,8 +344,8 @@ mod reference_transaction {
         match fields.as_slice() {
             [old_value, new_value, ref_name] => Ok(ParsedReferenceTransactionLine {
                 ref_name: ReferenceName::from(*ref_name),
-                old_oid: MaybeZeroOid::from_str(old_value)?,
-                new_oid: MaybeZeroOid::from_str(new_value)?,
+                old_value: ReferenceTarget::from_str(old_value)?,
+                new_value: ReferenceTarget::from_str(new_value)?,
             }),
             _ => {
                 eyre::bail!(
@@ -309,30 +359,51 @@ mod reference_transaction {
     #[cfg(test)]
     #[test]
     fn test_parse_reference_transaction_line() -> eyre::Result<()> {
-        use lib::core::eventlog::should_ignore_ref_updates;
+        use lib::{core::eventlog::should_ignore_ref_updates, testing::make_git};
 
-        let line = "123abc 456def refs/heads/mybranch";
+        let git = make_git()?;
+        git.init_repo()?;
+        let oid1 = git.commit_file("README", 1)?;
+        let oid2 = git.commit_file("README2", 2)?;
+
+        let zero = "0000000000000000000000000000000000000000";
+        let branch_ref = "refs/heads/mybranch";
+        let orig_head_ref = "ORIG_HEAD";
+        let master_tx_ref = "ref:refs/heads/master";
+        let master_ref = "refs/heads/master";
+        let head_ref = "HEAD";
+
+        let line = format!("{oid1} {oid2} {branch_ref}");
         assert_eq!(
-            parse_reference_transaction_line(line)?,
+            parse_reference_transaction_line(&line)?,
             ParsedReferenceTransactionLine {
-                old_oid: "123abc".parse()?,
-                new_oid: {
-                    let oid: MaybeZeroOid = "456def".parse()?;
-                    oid
+                old_value: ReferenceTarget::Direct { oid: oid1.into() },
+                new_value: ReferenceTarget::Direct { oid: oid2.into() },
+                ref_name: ReferenceName::from(branch_ref),
+            }
+        );
+
+        let line = format!("{zero} {master_tx_ref} HEAD");
+        assert_eq!(
+            parse_reference_transaction_line(&line)?,
+            ParsedReferenceTransactionLine {
+                old_value: ReferenceTarget::Direct { oid: zero.parse()? },
+                new_value: ReferenceTarget::Symbolic {
+                    name: ReferenceName::from(master_ref)
                 },
-                ref_name: ReferenceName::from("refs/heads/mybranch"),
+                ref_name: ReferenceName::from(head_ref)
             }
         );
 
         {
-            let line = "123abc 456def ORIG_HEAD";
+            let line = &format!("{oid1} {oid2} ORIG_HEAD");
             let parsed_line = parse_reference_transaction_line(line)?;
             assert_eq!(
                 parsed_line,
                 ParsedReferenceTransactionLine {
-                    old_oid: "123abc".parse()?,
-                    new_oid: "456def".parse()?,
-                    ref_name: ReferenceName::from("ORIG_HEAD"),
+                    old_value: ReferenceTarget::Direct { oid: oid1.into() },
+                    new_value: ReferenceTarget::Direct { oid: oid2.into() },
+                    ref_name: ReferenceName::from(orig_head_ref)
                 }
             );
             assert!(should_ignore_ref_updates(&parsed_line.ref_name));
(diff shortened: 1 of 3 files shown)


git-branchless-lib/src/git/oid.rs
```diff
@@ -154,10 +154,18 @@ impl FromStr for MaybeZeroOid {
     type Err = eyre::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.parse() {
+        let oid = s
+            // First try parsing as a hex commit hash.
+            .parse()
+            // If that doesn't work, attempt parsing as a commit ref.
+            .or_else(|_| git2::Oid::hash_object(git2::ObjectType::Commit, s.as_bytes()));
```
jblebrun, line 161: hash_object is wrong; we need to resolve the reference name. I think this requires a repository object. Working on a draft for this, but will probably require some architectural guidance, since I'm pretty new to the codebase.

jblebrun: Ok, this has been improved.

git-branchless-hook/src/lib.rs
```diff
@@ -42,6 +42,41 @@ pub use lib::core::rewrite::rewrite_hooks::{
     hook_skip_upstream_applied_commit,
 };
 
+/// A component that will find the hash for a given reference transaction OID.
+/// Reference transaction OIDs may be one of:
+///
+/// * An up-to-40-hex-digit hash string
+/// * A symbolic ref, in which case, it will start with "ref:"
+///
+/// See: https://github.com/git/git/commit/a8ae923f85da6434c3faf9c39719d6d5e5c77e65
```
arxanas, line 51:
```suggestion
/// See: "refs: support symrefs in 'reference-transaction' hook" https://github.com/git/git/commit/a8ae923f85da6434c3faf9c39719d6d5e5c77e65
```

suggestion: Add the commit title here somewhere so that the reader can more easily decide if they want to click the link (I don't feel strongly about the specific formatting).

comment (non-blocking):

> We do not add any tests in this commit since there is no git command which activates this flow

great thank you very much Git developers

git-branchless-hook/src/lib.rs
```diff
@@ -42,6 +42,41 @@ pub use lib::core::rewrite::rewrite_hooks::{
...
+    }
+
+    #[instrument]
+    /// Attempt to resolve the provided string as a transaction reference OID.
+    ///
+    /// If name starts with `ref:`, and the implementation has a reference lookup
+    /// mechanism, then the OID for the provided reference name will be
+    /// returned, or a failure, if no target was found.
+    ///
+    /// Otherwise, an attempt is made to parse `name` as an up-to-40-hex-digit hash,
```
arxanas, line 69: comment: I believe that this is imprecise, and the object ID will always be exactly 40 characters long (nothing about abbreviating it is mentioned in https://git-scm.com/docs/githooks#_reference_transaction), and, uh... the tests that use 6 characters instead of 40 characters were probably me being lazy because I knew it would work under the current implementation and I thought that it would be easier to read the tests that way 😬. But I can see now that it probably a bad idea insofar as one uses tests to determine the typical use of an interface.

You don't need to fix the tests, but, if we specify the exact format of the object ID in the comments/spec, it would be best to claim that we expect the object ID to be exactly 40 characters, or just don't specify that explicitly in the spec and maybe link to the `reference-transaction` hook documentation directly.

git-branchless-hook/src/lib.rs
```diff
@@ -289,13 +326,14 @@ mod reference_transaction {
...
         line: &str,
+        resolver: &TransactionReferenceFinder,
     ) -> eyre::Result<ParsedReferenceTransactionLine> {
         let fields = line.split(' ').collect_vec();
         match fields.as_slice() {
             [old_value, new_value, ref_name] => Ok(ParsedReferenceTransactionLine {
                 ref_name: ReferenceName::from(*ref_name),
-                old_oid: MaybeZeroOid::from_str(old_value)?,
-                new_oid: MaybeZeroOid::from_str(new_value)?,
+                old_oid: resolver.oid_for_transaction_reference(old_value)?,
```
arxanas, line 335: issue: This is only a function that claims to "parse" the line, and isn't (currently) supposed to do any processing/resolution. The question is whether we would prefer to preserve that, or just concede that mixing the parsing and resolution steps together is somehow better/more maintainable.

Based on my previous experience, I think it's actually best to keep this as a pure parsing function and handle this in a later phase. Specifically, it turned out to be valuable to separate and test the parsing of `git status` output in isolation.

To fix this, let's return some kind of `enum` and preserve the symbolically-represented input, rather than resolving it immediately. It looks like there already exists the `git::ReferenceTarget` type which we can use here.

- If it's not suitable for some reason, you can make your own, scoped to this file or even a function.
- I don't know your level of Rust experience. Here's an example of constructing the `Cow` if you need it (I recall it wasn't obvious to me): https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=afa8cdac4b58b09b217449b1d4e0897a. Feel free to clone any borrowed `str`s to owned `String`s if it helps.

After that, I think we can resolve the reference OIDs when constructing `events: Vec<Events>`. (Also depending on your level of Rust experience, you may find it easier to use an explicit loop rather than `into_iter`; up to you.)

suggestion (non-blocking): It may make sense to move the parsing code to belong to `ReferenceTarget`, either as an implementation of `FromStr`, or by a custom function. I don't know if any other caller will ever need it, though, so fine to keep local if you prefer.

comment: I assume that the field names in `ParsedReferenceTransactionLine` were meant to match the `reference-transaction` hook documentation directly, so the appropriate new names would be `old_value` and `new_value`. (I think we should actually reorder the fields to match the input order as well, so `ref_name` would be last, but not a big deal.)

git-branchless-hook/src/lib.rs
```diff
@@ -42,6 +42,41 @@ pub use lib::core::rewrite::rewrite_hooks::{
...
 
+/// A component that will find the hash for a given reference transaction OID.
+/// Reference transaction OIDs may be one of:
+///
+/// * An up-to-40-hex-digit hash string
+/// * A symbolic ref, in which case, it will start with "ref:"
+///
+/// See: https://github.com/git/git/commit/a8ae923f85da6434c3faf9c39719d6d5e5c77e65
+#[derive(Debug)]
+struct TransactionReferenceFinder<'a> {
```
arxanas, line 53: suggestion: I don't think we'll need a `struct`; it should be fine to just have a free function that accepts a `&Repo` and reference name (or possibly elide it altogether if it becomes unnecessary as part of other restructuring).

suggestion: Put it in the `reference_transaction` module somewhere near `parse_reference_transaction_line` (unless it's already there and I just can't read the Github code review rendering).

git-branchless-hook/src/lib.rs
```diff
@@ -42,6 +42,41 @@ pub use lib::core::rewrite::rewrite_hooks::{
...
+    /// mechanism, then the OID for the provided reference name will be
+    /// returned, or a failure, if no target was found.
+    ///
+    /// Otherwise, an attempt is made to parse `name` as an up-to-40-hex-digit hash,
+    /// returning the results of that parse as a [MaybeZeroOid].
+    fn oid_for_transaction_reference(&self, name: &str) -> eyre::Result<MaybeZeroOid> {
+        match name.strip_prefix("ref:") {
+            Some(refname) => Ok(self.repo.refname_to_oid(&refname.into())?),
+
+            None => name.parse(),
```
arxanas, line 75: issue: I'm not 100% sure, but I believe this drops/doesn't include the containing function in the span-trace. You need to write `Ok(name.parse()?)` to be sure to include it. This is unfortunate language design on Rust's part.

arxanas (changes requested):
Thanks for the fix!

I'll note that the long-term move is probably to stop using the `reference-transaction` hook, as it's problematic both fundamentally and incidentally:

- You have to walk the full event log to get the current reference locations.
- It's not actually possible to ensure atomic reference-transaction updates using the hook.
  - If we subscribe to the `prepared` hook, then there's no guarantee that the transaction will actually be committed.
  - If we subscribe to the `committed` hook, then it's possible that the hook call gets interrupted and we lose the update.
  - Occasionally my undo log shows stale branches that don't actually exist on disk anymore, which I think arises because of bad `reference-transaction` hook handling.
- Git currently doesn't group updates into reference transactions very well, so big fetches of $n$ branches induce $O(n)$ individual `reference-transaction` hook calls, which is incredibly slow and annoying.
  - For example: https://github.com/arxanas/git-branchless/discussions/1269#discussioncomment-8758122 😬 
  - I think Git itself doesn't currently have the ability to do batch transactional updates, at least not without migrating to the reftable format fully?

I think the [jj](https://github.com/martinvonz/jj) approach of snapshotting all the references every time is the only tractable long-term approach. At present, we actually walk all the references anyways to update our internal commit graph, so it's a cost we basically already incur.

jblebrun:
Also, sorry for force-pushing and destroying all of the conversation threads. It's muscle memory :-(

.github/workflows/linux.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       matrix:
         # Use a tag from https://github.com/git/git/tags
         # Make sure to update `git-version` in the `run-tests` step as well.
-        git-version: ["v2.24.3", "v2.29.2", "v2.33.1", "v2.37.3"]
+        git-version: ["v2.24.3", "v2.29.2", "v2.33.1", "v2.37.3", "v2.45.1"]
```
arxanas, line 21: issue: It looks like https://github.com/git/git/commit/a8ae923f85da6434c3faf9c39719d6d5e5c77e65 is not actually included in v2.45.1, so this doesn't actually reproduce the issue 😅. So we can remove this for now.

That commit is on `next`, and it's causing tests to fail on the git-devel CI job in an incidental way: https://github.com/arxanas/git-branchless/actions/runs/9047534056/job/24859622426, though (not this bug as far as I can tell, but expected if the reference-transaction interface has changed).

If you can get a test case that reproduces this against git `next`, then we can include it in this PR to prevent regressions, a.s it should be checked by the daily "devel" job instead of regular PR checks.

Also, just to be sure, you might want to test with `cargo build --workspace && cargo test ...` due to [this attempted build-time optimization](https://github.com/arxanas/git-branchless/wiki/Runbook#testing-subcommands-in-isolation). In principle, the optimization shouldn't apply unless you opt-in, but it's possible there's a bug. Hooks, by design, involve calling a bunch of differences subprocesses, so I just wanted to raise it in case you see unexpected stale tests.

arxanas:
I looked into the failing `test_amend` case on `master`, and it might be that the best long-term solution is to delete the `reference-transaction` hook entirely and rely on working-copy snapshotting instead 😬.

If we wanted to capture the new information provided by the reference-transaction hook, we'd have to update the event log schema (as it currently stores an OID _or_ a reference name), but not both; and even if we did all that, we'd still have all the reference-transaction hook issues as described above.

The main implementation problem with removing the reference-transaction hook is that we'd have to unpack and diff against the most recent working-copy snapshot in order to restore the current undo workflow and show the list of changed references. (That code path is also effectively used in the `bug-report` subcommand.)

jblebrun:
> I looked into the failing `test_amend` case on `master`, and it might be that the best long-term solution is to delete the `reference-transaction` hook entirely and rely on working-copy snapshotting instead 😬.

Sounds like it might be better to focus on the longer-term approach, then? Since the git commit that causes the issue isn't released yet, I can just use my personal branch with the patch for now, and we can abandon this PR, if you'd like. It was a fun exercise anyway.
 
> If we wanted to capture the new information provided by the reference-transaction hook, we'd have to update the event log schema (as it currently stores an OID _or_ a reference name), but not both; and even if we did all that, we'd still have all the reference-transaction hook issues as described above.

Is there a benefit to this, vs just sticking with the OIDs like now?

> The main implementation problem with removing the reference-transaction hook is that we'd have to unpack and diff against the most recent working-copy snapshot in order to restore the current undo workflow and show the list of changed references. (That code path is also effectively used in the `bug-report` subcommand.)

I could potentially help with this, although it's a larger-scope changed, so I'd have to get up to speed on a few more things first.

arxanas:
> I could potentially help with this, although it's a larger-scope changed, so I'd have to get up to speed on a few more things first.

I posted https://github.com/arxanas/git-branchless/issues/1334 to detail the necessary work.

> Is there a benefit to this, vs just sticking with the OIDs like now?

The main problem is that the reference-transaction hook would potentially misbehave when trying to undo/revert a symbolic reference. We treat `HEAD` specially in most cases (since there are very few other symbolic references that work well in Git) to get around this, so it's probably not a big deal in practice. It would be fine to just stick with the OIDs for now.

> Sounds like it might be better to focus on the longer-term approach, then? Since the git commit that causes the issue isn't released yet, I can just use my personal branch with the patch for now, and we can abandon this PR, if you'd like. It was a fun exercise anyway.

It will presumably be released at some point 😉. If you can add a test to expose the current failing behavior, then I'm happy to merge the PR.

jblebrun:
Ok, I've added a test and done a round of cleanup on this.

arxanas (approved):
I was able to reproduce the test failure on my end. Thanks for the fix!

mochja:
When may we expect a new release containing this fix?

arxanas:
@mochja unfortunately I haven't been able to release yet, but you can follow #1388 for updates.

PR #1420, opened 2024-10-14
fix(tests): ignore no-op checkouts in some tests

bryango:
This makes use of insta's `filters` feature.

### DIFF
--- a/git-branchless/tests/test_amend.rs
+++ b/git-branchless/tests/test_amend.rs
@@ -592,6 +592,14 @@ fn test_amend_undo() -> eyre::Result<()> {
     {
         let (stdout, _stderr) = git.branchless("undo", &["-y"])?;
         let stdout = trim_lines(stdout);
+        insta::with_settings!({filters => vec![
+            // ignore no-op checkouts, which is present in e.g. git 2.43
+            // but disappears in e.g. git 2.46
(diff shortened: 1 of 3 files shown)


git-branchless/tests/test_amend.rs
```diff
@@ -592,20 +592,26 @@ fn test_amend_undo() -> eyre::Result<()> {
...
         let (stdout, _stderr) = git.branchless("undo", &["-y"])?;
         let stdout = trim_lines(stdout);
+        insta::with_settings!({filters => vec![
+            // ignore no-op checkouts, which is present in e.g. git 2.43
+            // but disappears in e.g. git 2.46
+            (r"\n2\. Check out from 94b1077 create file1.txt\s*to 94b1077 create file1.txt\n", "\n"),
+            // ignore different numbering of events
+            (r"(?m)^[0-9]+\. ", "#. "),
+            ("Applied ([0-9]+) inverse events.", "Applied # inverse events."),
+        ]}, {
```
claytonrcarter: This is interesting! I didn't know about `insta` filters. So the idea for these no-op checkouts is that they were introduced into git and then quickly-but-not-quickly-enough removed again?

FYI I also see some of these no-op checkout messages in `test_undo_garbage_collected_commit`.

Issue #1404, opened 2024-09-20
Panic on executing `git prev` with custom `status` configuration: unknown status line prefix ASCII character: 35


TrueDoctor:
### Description of the bug

When executing `git prev` I get this error:

```
Graphite on  master [$?] ➜ git prev
branchless: creating working copy snapshot
The application panicked (crashed).
Message:  A fatal error occurred:
   0: unknown status line prefix ASCII character: 35

Location:
   git-branchless-lib/src/core/check_out.rs:221

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: branchless::core::check_out::check_out_commit with effects=<Output fancy=true> git_run_info=<GitRunInfo path_to_git="git" working_directory="/home/dennis/Projects/rust/Graphite" env=not shown> repo=<Git repository at: "/home/dennis/Projects/rust/Graphite/.git/"> event_log_db=<EventLogDb> event_tx_id=EventTransactionId(1264) target=Some(Oid(NonZeroOid(ef007736f5400602348436c978963baa7c51e232))) options=CheckOutCommitOptions { additional_args: [], reset: false, render_smartlog: true }
      at git-branchless-lib/src/core/check_out.rs:107
   1: git_branchless_navigation::traverse_commits with effects=<Output fancy=true> git_run_info=<GitRunInfo path_to_git="git" working_directory="/home/dennis/Projects/rust/Graphite" env=not shown> command=Prev options=TraverseCommitsOptions { num_commits: None, all_the_way: false, move_by_branches: false, oldest: false, newest: false, interactive: false, merge: false, force: false }
      at git-branchless-navigation/src/lib.rs:309

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 3 frames hidden ⋮
   4: branchless::core::check_out::create_snapshot::ha2ce3d73aeffbb12
      at <unknown source file>:<unknown line>
   5: branchless::core::check_out::check_out_commit::h333d15830b408054
      at <unknown source file>:<unknown line>
   6: git_branchless_navigation::traverse_commits::hc20dc93f7e901dd9
      at <unknown source file>:<unknown line>
[...]

PR #1439, opened 2024-11-08
Skip headers in status output

TrueDoctor:
Fixes: https://github.com/arxanas/git-branchless/issues/1404


arxanas:
Thanks! Presumably you found that it fixes the issue that arose with your local configuration?

TrueDoctor:
Yes, I have confirmed that this fixes my issue

arxanas:
Thanks! Ideally we would add a test, but I think we have no explicit snapshot tests at the moment 😅.

PR #1446, opened 2024-11-15
fix(init): resolve hooks path relative to repo root

ethanwu10:
Ensure that `core.hooksPath` is resolved relative to the repository root of the main worktree, or the git dir if it is bare (git itself uses the same logic).


ethanwu10:
I'm not clear on which worktree `core.hooksPath` is being read from and where it's supposed to be resolved relative to. If I'm reading the [git docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath) correctly, they should be read from the current worktree regardless of which worktree the config value itself was read from. I'll need to figure out how to test this properly.

arxanas:
> I'm not clear on which worktree core.hooksPath is being read from and where it's supposed to be resolved relative to. If I'm reading the [git docs](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corehooksPath) correctly, they should be read from the current worktree regardless of which worktree the config value itself was read from. I'll need to figure out how to test this properly.

I agree with that interpretation of the docs: a relative hook path always needs to be interpreted from the worktree root that the hook is running in (which means that there's not a way to specify a hook path that's relative to the *main* worktree).

git-branchless-init/src/lib.rs
```diff
@@ -606,8 +606,7 @@ fn command_init(
     main_branch_name: Option<&str>,
 ) -> EyreExitOr<()> {
     let mut in_ = BufReader::new(stdin());
-    let repo = Repo::from_current_dir()?;
-    let mut repo = repo.open_worktree_parent_repo()?.unwrap_or(repo);
+    let mut repo = Repo::from_current_dir()?;
 
     let default_config = Config::open_default()?;
     let readonly_config = repo.get_readonly_config()?;
```
ethanwu10, line 612: This PR changes `readonly_config` to be read from the current worktree instead of the root. (It's used for modifying the root config to include `branchless/config`) This doesn't seem to break any tests, and also `uninstall` uses the current worktree instead of the root. I'm not sure if this is a problem or not, since I'm not clear on how much of the config file init/uninstall process is covered in the worktree tests.

PR #1461, opened 2024-12-08
feat(revset): add + as shortcut for "only child"

claytonrcarter:
This adds `+` as a postfix revset operator that evaluates to `sole(children(...))`. I chose `+` because it matches the similar operator in [`jj` revsets](https://martinvonz.github.io/jj/latest/revsets/#operators), but I didn't add `-` for symmetry because I didn't feel the need since we already have `~` and `^`.

Note that I chose `sole(...)` because I have found that I (almost?) always wanted this shorthand in cases where there is only 1 child, and if I wanted all children, I could reach for the regular `children(...)` function. I don't think I care about this *that* strongly, though, and I'd be OK to remove `sole(...)` if preferred.

Handy use cases:
- when it's easier to paste the commit already on the clipboard and just add `+` or `++` instead of copying the commit you really want :)
- unstacking a stack of branches: `git move -s prev-branch+ -d main`
- another way to do `git next`: `git sw @+`


arxanas:
Sorry, haven't had time to go through my pending PRs yet. Generally speaking, I would want to be consistent with jj's revset language where possible, since git-branchless is probably a gateway to jj for many users.

claytonrcarter:
> consistent with jj's revset

OK, for now I have reimplemented this to use `!` instead of `+`. Not only does it avoid the conflict with `jj`, but I found that my original implementation wasn't sophisticated ~~at all~~ enough to correctly handle a case like `@+ + @++`: it was parsed as `(@+ + @+)+` instead of my intended `@+ + (@++)`. I didn't understand the lalrpop grammar enough to fix that. Using `!` sidesteps both of these issues, but at the cost of introducing a totally new operator.

If I can get my head around the union conflict and the lalrpop grammar, I would prefer to implement it to mimic `jj`, so I'll just draft this `!` version for now.

PR #1463, opened 2024-12-08
feat(switch): accept revset arguments, and explicitly support `-` shorthand

claytonrcarter:
This updated `git switch` to accept a revset as it's argument, and also to support `-` (to checkout previously checked out branch/commit) by passing it through to git.

The revset parameter must evaluate to a set with a single head, and I've found this handy in cases like `current(...)` to move to the latest version of a commit if I had an old smartlog or SHA still on screen or such, and also `switch foo:` for when I have a branch ref handy but I really want to checkout the latest detached commit in that stack.

But with revset support added, `switch -` broke because it looked a poorly formed revset:
```
   0: parse error in "-": parse error: Unrecognized token `-` found at 0:1
      Expected one of "(", "..", ":", "::", a commit/branch/tag or a string literal
   1: parse error: Unrecognized token `-` found at 0:1
      Expected one of "(", "..", ":", "::", a commit/branch/tag or a string literal
```
The last commit in this stack adds an explicit check for `-` so that this support is restored. 

Apparently, `revparse_single()` in `libgit2` doesn't support this. ~~I didn't patch `-` support into `repo.revparse_single_commit()` but I'm thinking as I write this that perhaps that may be better? There is already a special case in that function to work around another `libgit2` issue, and I suppose we could detect if the passed spec is `-` and then pass `@{-1}` to `revparse_single()` instead. 🤔 ~~ **edit** I made this change and it seems to be working.

### DIFF
--- a/git-branchless-navigation/src/lib.rs
+++ b/git-branchless-navigation/src/lib.rs
@@ -474,7 +474,7 @@ pub fn switch(
     switch_options: &SwitchOptions,
 ) -> EyreExitOr<()> {
     let SwitchOptions {
-        interactive: _,
+        interactive,
         branch_name,
         force,
         merge,
@@ -510,27 +510,23 @@ pub fn switch(
         false,
     )?;
 
-    let initial_query = match switch_options {
-        SwitchOptions {
-            interactive: true,
-            branch_name: _,
-            force: _,
-            merge: _,
-            detach: _,
-            target,
-        } => Some(target.clone().unwrap_or_default()),
-        SwitchOptions {
-            interactive: false,
-            branch_name: _,
-            force: _,
-            merge: _,
-            detach: _,
-            target: _,
-        } => None,
+    enum Target {
+        /// The (possibly empty) target expression should be used as the initial
+        /// query in the commit selector.
+        Interactive(String),
+
+        /// No target expression was specified.
+        None,
+    }
+    let initial_query = match (interactive, target) {
+        (true, Some(target)) => Target::Interactive(target.clone()),
+        (true, None) => Target::Interactive(String::new()),
+        (false, Some(_)) => Target::None,
+        (false, None) => Target::None,
     };
     let target: Option<CheckoutTarget> = match initial_query {
-        None => target.clone().map(CheckoutTarget::Unknown),
-        Some(initial_query) => {
+        Target::None => None,
+        Target::Interactive(initial_query) => {
             match prompt_select_commit(
                 None,
                 &initial_query,
--- a/git-branchless-navigation/src/lib.rs
+++ b/git-branchless-navigation/src/lib.rs
@@ -25,11 +25,11 @@ use lib::core::repo_ext::RepoExt;
 use lib::util::{ExitCode, EyreExitOr};
 use tracing::{instrument, warn};
 
-use git_branchless_opts::{SwitchOptions, TraverseCommitsOptions};
-use git_branchless_revset::resolve_default_smartlog_commits;
+use git_branchless_opts::{ResolveRevsetOptions, Revset, SwitchOptions, TraverseCommitsOptions};
+use git_branchless_revset::{resolve_commits, resolve_default_smartlog_commits};
 use git_branchless_smartlog::make_smartlog_graph;
 use lib::core::config::get_next_interactive;
-use lib::core::dag::{sorted_commit_set, CommitSet, Dag};
+use lib::core::dag::{sorted_commit_set, union_all, CommitSet, Dag};
 use lib::core::effects::Effects;
 use lib::core::eventlog::{EventLogDb, EventReplayer};
 use lib::core::formatting::Pluralize;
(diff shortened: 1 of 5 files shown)


git-branchless-navigation/src/lib.rs
```diff
@@ -512,16 +512,22 @@ pub fn switch(
 
     enum Target {
         Interactive(String),
+        Revision(String),
```
arxanas, line 515: issue: I think it's misleading to call this `Revision`, because it sounds like it's meant to refer to a revision/commit specifically, and not be a general revision specification, which can indicate a branch.

Some ideas for renaming it:
- `Revspec` (but I'm not sure if this is legitimate Git terminology)
- `RevisionOrRef`
- `RevparseSingleCommit`
- `Uninterpreted` or similar
  - I guess it would be better if `CheckoutTarget::Unknown` were actually named `CheckoutTarget::Uninterpreted` or similar, since `Unknown` is kind of weird to say when the value is clearly known.

Let's also add a comment to this `enum` case explaining that it can also represent branches, etc.

git-branchless-navigation/src/lib.rs
```diff
@@ -544,6 +550,36 @@ pub fn switch(
...
+                &ResolveRevsetOptions::default(),
+            )?;
+
+            let commit_set = union_all(&commit_sets);
+            let commit_set = dag.query_heads(commit_set)?;
+            let commits = sorted_commit_set(&repo, &dag, &commit_set)?;
+
+            match commits.as_slice() {
+                [commit] => Some(CheckoutTarget::Unknown(commit.get_oid().to_string())),
+                [] => None,
```
arxanas, line 568: question: Is it intentional for you that having a revset that resolves to the empty set should do nothing? I could imagine it being useful for some workflow, although I don't have one in mind.

issue: Technically speaking, isn't the stated specification that the set must have exactly 1 head not correct?

git-branchless-navigation/src/lib.rs
```diff
@@ -544,6 +550,36 @@ pub fn switch(
...
+                [..] => {
+                    writeln!(
+                        effects.get_error_stream(),
+                        "Cannot switch to target: expected '{target}' to contain 1 head, but found {}.",
```
arxanas, line 572: comment (non-blocking): In principle we could list all of the candidate targets, which we also do when `git next` or `git prev` is ambiguous, but it's not necessary to implement that here.

git-branchless/tests/test_navigation.rs
```diff
@@ -909,6 +909,32 @@ fn test_navigation_switch_revset() -> eyre::Result<()> {
...
+        |\
+        | o 62fc20d create test1.txt
+        |
+        @ fe65c1f create test2.txt
+        "###);
+    }
+
+    {
+        // switching back to "last checkout"
+        let (stdout, _stderr) = git.branchless("switch", &["-"])?;
```
arxanas, line 927: suggestion: Can you add a test to ensure that `git switch -` is also capable of checking out the previous branch, not just the previous commit? (I believe that's how it works in Git.)

arxanas (approved):
Thanks! Both of these issues have bothered me personally as well.

git-branchless-navigation/src/lib.rs
claytonrcarter: Good call. I ended up with `Passthrough` here and documented as follows:
```
        /// The target expression is probably a git revision or reference and
        /// should be passed directly to git for resolution.
        Passthrough(String),
```

I was tempted by `RevisionOrRef` but wasn't sure if that might also fall into the same trap as just `Revision`. I too am not sure on my git terminology on this one, though, so maybe  revision and ref really are the only two options?

Happy to change this again if you'd like.

git-branchless-navigation/src/lib.rs
claytonrcarter: > Is it intentional for you that having a revset that resolves to the empty set should do nothing? 

No, I don't recall a specific intention to this. I suspect I added this to mimic the behavior of `Target::None`, but I'm also blanking on a use case. I think I'll remove this and *actually* enforce the "exactly 1 head" statement. We can relax that later if a desired workflow involving empty sets emerges. 😄

git-branchless/tests/test_navigation.rs
claytonrcarter: Oh, good catch. Done.

git-branchless-navigation/src/lib.rs
claytonrcarter: Yes, that makes sense, but I'm going to defer on that for now and leave that as a great future enhancement.

claytonrcarter:
> Thanks! Both of these issues have bothered me personally as well.

Thanks for the review! I've updated my change, and I'll park this here for a few days while I continue using it locally in case I missed anything.

claytonrcarter:
This has been working fine for me. Let's see what everyone else thinks of it. 😄

Issue #180, opened 2021-10-29
Implement an interactive `git split` command


bcongdon:
As discussed in #48.

Note: Depending on how this is implemented, we may need a `git stage` command as a prerequisite for this

arxanas:
Note that all of the individual parts to implement this are now present, if anyone wants to give it a try. The interactive change selector already works for `git record -i` and replacing a commit with multiple commits can be done with the rebase engine.

claytonrcarter:
Something about my `git-split` shell script stopped working recently and instead of trying to fix, I decided to take on a new project 😆 so I've started working on a non-interactive version of this. I'll post a PR after I've used it for a bit. Right now the UI is `git split <revset> <file> [<file> ...]` and it's only extracting entire file diffs. We could extend it with an interactive mode later. I'd also like to eventually be able to specify hunks on the command line, eg `split @ file.txt:123` to extract only the part of the diff that intersects with line 123.

PR #1464, opened 2024-12-08
feat: add `split` command

claytonrcarter:
This adds a `git split` command to extract changes from commits. This mostly addresses #180, except for the `-i/--interactive` part. The split and rebase happens in-memory, so it should be fast. The command signature is

`git [branchless] split <revset> <file> [<file>...]`

- `<revset>` must resolve to a single commit (or just be a single commit, or `@`, etc)
- if only one file is given, then the file name is used in the extracted commit message
- if 2+ files are split out, then `<n> files` is used in the extracted commit message
- in either case, a brief insert/delete summary is also used, eg `(+2/-5)`
  - eg `temp(split): lib.rs (+5)` or `temp(split): 3 files (+3/-3)`

#### Implemented flags

These alter how the split/rebase happens:

- _default_ (no flags) is "insert after": extracted changes are inserted into commit graph as a child of the target commit
  - any children of the target commit are rebased onto the extracted commit
  - eg `A - B - C` → `A - B' - b - C`
- `--detach`: similar to above, but
  - children of the target commit are rebased onto the _newly split target_ commit, not onto the extracted commit
  - eg `A - B - C` → `A - B' - C` and `B' - b`
  - that is: `b` and `C` are now both children of `B'`
- `--discard`: similar to `--detach`, but after extracting the changes, they are simply discarded
  - eg `A - B - C` → `A - B' - C`
- `--before`: the extracted changes are inserted as a parent of the newly split target commit
  - children of the target commit are rebased onto the _newly split target_ commit
  - eg `A - B - C` → `A - b - B' - C`
- with `--detach` and `--discard`, rebase conflicts are possible, and are  intended to be resolved in the same way are for `git move` conflicts

#### Status

This is "works for me" and has been stable and useful over months of personal testing. That said, the code is mostly OK, but there are still a few FIXME/TODOs to resolve, and I'm certain that the code could be more idiomatic or branchless
in general. I have written a bunch of tests, but it's possible that I have overlooked some edges.

Any and all feedback is welcome. I'm happy with this as is, but I'm looking forward to some review helping me be really proud of it. =)

#### Specific areas needing feedback and review

- The temporary commit messages. Style, format, etc. The code that generates  them feels sort of complicated; maybe it's not pulling its weight, and we  should use a simpler or more familiar option?
- The structure and flow of the business logic still confuses me a bit, and I  suspect I'm overlooking some simplifications. For example, the variable names  used for the new commits (`extracted_commit` and `remainder_commit`) feel good, but they  actually mean the opposite if the `--before` flag is passed. :shrug: I've  commented this to explain, but I'm hoping someone else will have some  suggestions.
- I introduced a `force_detach` field to `CheckOutCommitOptions` in order to alter the behavior of the "automatically checkout branch" feature. This is because I was having issues w/ a detached branch becoming attached/checked out after being split. This seems like something that may not be needed, but it's the best that I could come up with at this time. If it *is* unnecessary, it's a self-contained change (commit 972d430 `refactor: add force_detach checkout option`) with good test coverage in other commands as well as in this new command, so it should be straightforward to back out later.,

### DIFF
--- a/git-branchless-init/src/lib.rs
+++ b/git-branchless-init/src/lib.rs
@@ -95,6 +95,7 @@ const ALL_ALIASES: &[(&str, &str)] = &[
     ("restack", "restack"),
     ("reword", "reword"),
     ("sl", "smartlog"),
+    ("split", "split"),
     ("smartlog", "smartlog"),
     ("submit", "submit"),
     ("sw", "switch"),
--- a/git-branchless-lib/src/core/check_out.rs
+++ b/git-branchless-lib/src/core/check_out.rs
@@ -44,6 +44,9 @@ pub struct CheckOutCommitOptions {
     /// Additional arguments to pass to `git checkout`.
     pub additional_args: Vec<OsString>,
 
+    /// Ignore the `autoSwitchBranches` setting?
+    pub force_detach: bool,
+
     /// Use `git reset` rather than `git checkout`; that is, leave the index and
     /// working copy unchanged, and just adjust the `HEAD` pointer.
     pub reset: bool,
@@ -56,6 +59,7 @@ impl Default for CheckOutCommitOptions {
     fn default() -> Self {
         Self {
             additional_args: Default::default(),
+            force_detach: false,
             reset: false,
             render_smartlog: true,
         }
@@ -116,6 +120,7 @@ pub fn check_out_commit(
 ) -> EyreExitOr<()> {
     let CheckOutCommitOptions {
         additional_args,
+        force_detach,
         reset,
         render_smartlog,
     } = options;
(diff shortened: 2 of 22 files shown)


claytonrcarter:
I pushed a fix for an issue that I had noticed for a while but couldn't reproduce: using `split --discard` on the current commit when the current commit was a checked out branch would leave the discarded changes in the index, making them look like they had been staged. My fix is to reset the index to the state of the newly split commit. That passes tests, but it's not clear to me how it will interact w/ something like the working copy snapshots. I think (hope?) those are applied later, but I'll play with this some more to see.

PR #1531, opened 2025-04-03
wip: supports gpg signing

Abhijay:
This works, but I don't plan on writing tests. Feel free to pick it up or close it if you don't want to.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,7 +69,7 @@ git-branchless-smartlog = { version = "0.10.0", path = "git-branchless-smartlog"
 git-branchless-submit = { version = "0.10.0", path = "git-branchless-submit" }
 git-branchless-test = { version = "0.10.0", path = "git-branchless-test" }
 git-branchless-undo = { version = "0.10.0", path = "git-branchless-undo" }
-git2 = { version = "0.20.0", default-features = false }
+git2 = { version = "0.18.0", default-features = false }
 glob = "0.3.2"
 indexmap = "2.7.1"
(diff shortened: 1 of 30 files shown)


Cargo.toml
```diff
@@ -69,7 +69,7 @@ git-branchless-smartlog = { version = "0.10.0", path = "git-branchless-smartlog"
 git-branchless-submit = { version = "0.10.0", path = "git-branchless-submit" }
 git-branchless-test = { version = "0.10.0", path = "git-branchless-test" }
 git-branchless-undo = { version = "0.10.0", path = "git-branchless-undo" }
-git2 = { version = "0.20.0", default-features = false }
+git2 = { version = "0.18.0", default-features = false }
```
forivall, line 72: What was the rationale for the downgrade?

mhammerly: the signing capabilities come from https://github.com/gitext-rs/git2-ext which locks its `git2` dep at 0.18.0 (as does its dependency https://github.com/gitext-rs/git-fixture). but `git-fixture` already updated the dep and just needs a new release, and from what i can tell `git2-ext` can upgrade without issue as well

[it sounds like 0.20.0 unlocked some functionality in `git-branchless`](https://github.com/arxanas/git-branchless/commit/47431f3db7479fd0bdcd92b81d2ffdb7b729ef7e) so downgrading would be a breaking change

mhammerly:
i included your `--show-signature` commits in a new PR which also adds integration tests: https://github.com/arxanas/git-branchless/pull/1538

i like your addition, by the way! i have been using it at work

PR #966, opened 2023-06-14
[WIP] Support GPG-signing commits

tommyip:
This is a work in progress, just wanted to publish it early to gather some feedback.

TODO:
- [x] `git record` (calls `git commit` directly so signing already works in master, added new flags to override keyid or disable signing)
- [x] `git amend`
- [ ] `git move`
  - [ ]  in-memory
  - [ ] on-disk
- [ ] `git sync`
- [ ] `git restack`
- [ ] `git branchless snapshot create`

Fixes #465

Issue #465, opened 2022-07-18
Support GPG-signing commits


arxanas:
Some resources:

- Using `git2`'s `Repository::commit_signed`: https://github.com/rust-lang/git2-rs/issues/507. It looks like you still need to sign the commit via GPG yourself.
  - Another example of signing the commit: https://blog.hackeriet.no/signing-git-commits-in-rust/
- An implementation for git-revise: https://github.com/mystor/git-revise/pull/73

There's a few APIs in git-branchless which are capable of creating commits:
- [`Repo::create_commit`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.Repo.html#method.create_commit).
- [`Repo::cherry_pick_commit`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.Repo.html#method.cherry_pick_commit) and [`Repo::cherry_pick_fast`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.Repo.html#method.cherry_pick_fast).
- [`Repo::amend_fast`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.Repo.html#method.amend_fast).

These commands are capable of creating commits:

- `git move`.
  - Note that it can create commits in-memory or on-disk via `git rebase`, so both cases will have to be handled.
  - The calling code is in `core::rewrite::execute`. We'll probably want to add some configuration to  to indicate that any commits created in this way should be signed. Also [`CherryPickFastOptions`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.CherryPickFastOptions.html).
- `git amend`.
- `git reword`.
- `git sync` and `git restack`: these should be handled fairly straightforwardly by whatever changes are made to support `git move`.
- `git record`: experimental command, so not a priority at the moment.
- `git branchless snapshot create`: these commits don't need to be signed since they're not supposed to be really committed/pushed.

These data structures will probably have to be updated to carry GPG information. You should be able to add a field and then follow the compiler errors to see where should be updated to use the new field:

- [`AmendFastOptions`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/enum.AmendFastOptions.html).
- [`CherryPickFastOptions`](https://docs.rs/git-branchless-lib/0.3.12/branchless/git/struct.CherryPickFastOptions.html)
[...]

PR #1538, opened 2025-04-19
support for openpgp, ssh, x509 signatures with tests

mhammerly:
yet another PR enters the mix for https://github.com/arxanas/git-branchless/issues/465 

mostly based on:
- https://github.com/arxanas/git-branchless/pull/966
  - mostly i rebased, made a couple changes to reading config
- https://github.com/arxanas/git-branchless/pull/1531
  - did not change this at all, just stole the smartlog stuff

my additions:
- updating the MSRV to 1.80, in line with the `git2-ext` MSRV
- fixed existing snapshot tests
- adds integration tests for creating commit signatures using `gpg`, `ssh-keygen`, and `gpgsm`.

there are three integration tests:
- one for `openpgp` signatures, which is gated by a `test_gpg` feature flag and requires a `TEST_GPG` env var to point at a `gpg` binary
- one for `ssh` signatures, which is gated by a `test_ssh` feature flag and requires a `TEST_SSH_KEYGEN` env var to point at an `ssh-keygen` binary
- one for `x509` signatures, which is gated by a `test_x509` feature flag and requires a `TEST_GPGSM` env var to point at a `gpgsm` binary

the `openpgp` and `x509` tests actually expect GnuPG tools because their keyring format or whatever is checked in (meaning, no `smimesign` for mac)

### `git2` dependency version disagreement

**this PR is not ready to merge.** `git-branchless` depends on `git2` version 0.20.0 whereas the new dependency `git2-ext` caps its dependency at 0.18.0. we need to resolve that before merging

to get this PR building until then, check out `git2-ext` at `../git2-ext` and edit its Cargo.toml to:
- update its `git2` dep to 0.20.0
- update its `git-fixture` dep to point to that project's `main` branch on GitHub (which was already updated but not released)

we either need a maintainer over there to do a little chore to update or we need to downgrade to 0.18.0 ([but i think that would be breaking](https://github.com/arxanas/git-branchless/commit/47431f3db7479fd0bdcd92b81d2ffdb7b729ef7e)).


claytonrcarter:
> this PR is not ready to merge. git-branchless depends on git2 version 0.20.0 whereas the new dependency git2-ext caps its dependency at 0.18.0. we need to resolve that before merging

If I understand the deps situation correctly, we are waiting on:
- `git-fixture` is updated but needs a release
- `git2-ext` (which depends on `git-fixture`) needs an update and release

The good news is that I believe @epage maintains both of those and has contributed to this project in the past. So, if you can open a PR to update `git2-ext`, we may be able to beg a favor to get the deps straightened out.

I don't have any knowledge or experience w/ gpg or commit signing, but I may be able to do some code review to shepherd this along. It's a much requested feature!

mhammerly:
i pinged ed on https://github.com/arxanas/git-branchless/issues/465#issuecomment-2799044139 to ask whether there was a reason the dependency version was capped at 0.18 and request these updates. i suppose opening a PR to do the `git2` update is easy enough, though we'll need another PR to update `git-fixture` when that gets released

epage:
Sorry, no one had told me they were moving forward with this and needed a release with support for new git2 versions.

I've released v0.6.3.

mhammerly:
no apology needed! thank you for implementing the signing functionality in the first place, and for doing this little chore to unblock integrating it! i'm using it at work every day and it works great!

mhammerly:
alright, updated `git2-ext`, quoted the rustc version strings because yaml was parsing `1.80` to `1.8`, set the paths for `gpg`, `gpgsm`, `ssh-keygen` in test invocations

remaining issue is apparently `gpg.format ssh` support was only added in git >= 2.34 so we can't use `--all-features` on older versions. i'll deal with that later. but the other two tests pass, and if the 2.37.3 test job goes first then all three should pass

git-branchless-lib/src/git/repo.rs
```diff
@@ -23,6 +23,7 @@ use chrono::{DateTime, Utc};
 use cursive::theme::BaseColor;
 use cursive::utils::markup::StyledString;
 use git2::DiffOptions;
+use git2_ext::ops::Sign;
```
claytonrcarter, line 26: [**suggestion (maybe/probably change)**] I think that it would be best to wrap our own type around this (for example `crate::git::sign::Signer`) to reduce the number of places we may have to rely on `git2_ext` and make future refactoring easier. (eg if we ever switch to signing support directly in `git2`, we can just update our implementation of `Signer` (hopefully) without having to update all of the args, params, etc that depend on it.)

Here's an example of how this is done for `Commit`:
https://github.com/arxanas/git-branchless/blob/b1f7ae5951342825eea01f6fe1ce3860f47841da/git-branchless-lib/src/git/object.rs#L19-L23

git-branchless-lib/src/git/object.rs
```diff
@@ -291,26 +293,59 @@ impl<'repo> Commit<'repo> {
...
+                }
+            },
+        };
+        let new_tree = match tree {
+            Some(tree) => &tree.inner,
+            None => &self.inner.tree().map_err(|err| Error::FindTree {
+                source: err,
+                oid: self.inner.tree_id().into(),
+            })?,
+        };
```
claytonrcarter, line 337: [**question**] Why these changes? Or were these tagging along from the rebased implementation? At a glance, it seems like we may be able to use the previous versions (eg `author.map(|author| &author.inner)`, etc) but I haven't tested that.

This seems to introduce a break in behaviour: previously, passing `None` into this fn for any of these values would have passed that directly to `git2::amend()`, but here we seem to be discarding `None` and instead converting it to the existing author/message/tree/etc. Is that needed for signing? If not, can you see if this could be simplified and/or if any of these defaults are needed?

git-branchless-record/src/lib.rs
```diff
@@ -154,17 +157,23 @@ fn record(
...
+
+        if working_copy_changes_type == WorkingCopyChangesType::Unstaged {
+            args.push("--all");
+        }
+
+        let sign_flag = sign_option.as_git_flag();
+        if let Some(flag) = &sign_flag {
+            args.push(flag);
+        }
+
```
claytonrcarter, line 176: [**nitpick**] Can you restore the block expression syntax (in other words `let args = { ... }`) while including the new signing lines? That way `args` can remain immutable in the `else`.

git-branchless-record/src/lib.rs
```diff
@@ -405,13 +416,17 @@ fn record_interactive(
...
+    if !message.is_empty() {
+        args.extend(["--message", &message]);
+    }
+
+    let sign_flag = sign_option.as_git_flag();
+    if let Some(flag) = &sign_flag {
+        args.push(flag);
+    }
+
```
claytonrcarter, line 429: [**please change**] Same here, please: `let mut args = ...;` => `let args = { ... }`

claytonrcarter, line 429: [**please change**] Same here, please: `let mut args = ...;` => `let args = { ... }`

claytonrcarter, line 429: [**please change**] Same here, please: `let mut args = ...;` => `let args = { ... }`

claytonrcarter, line 429: [**please change**] Same here, please: `let mut args = ...;` => `let args = { ... }`

git-branchless-lib/src/core/node_descriptors.rs
```diff
@@ -516,6 +516,79 @@ impl NodeDescriptor for RelativeTimeDescriptor {
...
+            .args(["-C", &self.repo_path, "verify-commit", &oid, "--raw"])
+            .output();
+
+        match output {
+            Ok(output) => {
+                if output.status.success() {
+                    // Good signature
+                    let mut result = StyledString::new();
+                    result.append_styled("[G]", BaseColor::Green.light());
+                    Ok(Some(result))
```
claytonrcarter, line 562: [**nitpick (maybe change)**] It looks like other uses of `StyledString` in this file seem to use `let result = StyledString::styled("[G]", BaseColor::Green.light());`. To be consistent, can you update these uses of `::new()` to match. Or is there a reason that won't work in this context?

git-branchless-smartlog/tests/test_smartlog.rs
```diff
@@ -812,3 +812,95 @@ fn test_exact() -> eyre::Result<()> {
...
+
+#[test]
+fn test_show_signature_flag() -> eyre::Result<()> {
+    let git = make_git()?;
+
+    git.init_repo()?;
+
+    // Just create a regular commit without trying to sign it
+    git.commit_file("signed_file", 1)?;
+    git.commit_file("unsigned_file", 2)?;
```
claytonrcarter, line 824: [**question**] Are these filenames/comments correct? I don't know much about signing commits, but it seems like something is missing to actually do the signing.

git-branchless-smartlog/tests/test_smartlog.rs
```diff
@@ -812,3 +812,95 @@ fn test_exact() -> eyre::Result<()> {
...
+    git.commit_file("another_file", 2)?;
+
+    // Run smartlog with --show-signature flag
+    let (stdout, _) = git.branchless("smartlog", &["--show-signature"])?;
+
+    // Simply verify the command runs without error
+    assert!(!stdout.is_empty(), "Smartlog output should not be empty");
+
+    Ok(())
+}
```
claytonrcarter, line 867: [**please change**] This seems to be very similar to the previous test. Can you double to check and confirm that they're both really needed?

git-branchless-smartlog/tests/test_smartlog.rs
```diff
@@ -812,3 +812,95 @@ fn test_exact() -> eyre::Result<()> {
...
+    );
+
+    let result = git.branchless("smartlog", &["--show-signature", "--reverse"]);
+    assert!(
+        result.is_ok(),
+        "The --show-signature flag should work with --reverse"
+    );
+
+    Ok(())
+}
```
claytonrcarter, line 906: [**please change**] I also think that this test can be removed; mostly it seems to be testing that `clap` is working as expected. If I'm misunderstanding, let's update it to be more clear about which flags may or may not work with the new `--show-signature` flag.

claytonrcarter (changes requested):
Thank you for shepherding this work through, and to the previous authors for getting it started. Most of this looked pretty straightforward and made sense to me, but there were a few places I requested some changes, most of which were to refactor this code into a similar style as the rest of the code base.

I'll take another look when that's done. Thanks again!

git-branchless-lib/src/git/object.rs
claytonrcarter: Please disregard this comment. After sleeping on it, I realized why we need these here.

PR #1590, opened 2025-08-26
build: update MSRV to 1.85

claytonrcarter:
We are currently pinned to 1.74 which is from late 2023. This updates us to 1.85 (early 2025) which is currently about 6 months old and matches the current MSRV of `jj`.

TBD I need to look in the context of when we last updated our MSRV (in e87ab50) to see if we should use a different version. (ie maybe we should continue following `clap` instead of `jj`)

Ref: e87ab50

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -30,7 +30,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.74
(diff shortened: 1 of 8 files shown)


claytonrcarter:
Well, this seems to be working with 1.85 except for the nix build, which is apparently still using 1.75. I don't know much about nix, but I assume that I just need to update the nix store/hash that we pin to.

Aside from that, I checked and `clap` is still using 1.74 as MSRV.

PR #1612, opened 2025-09-26
feat(record): add `--new` flag

claytonrcarter:
This adds `--new` support to `git record`, to create a new, empty commit. `--insert` works as expected, too.

This required a couple of changes to our tests, too, because this is the first feature in which we create entirely new commits (vs rewriting or modifying existing commits) and I was running into test failures in CI that all focused on time:
- use a correctly formatted RFC2822 date for our deterministic "dummy date" (git parsed and used the existing date format, but the git2 crate rejected it)
- explicitly set timezone to UTC in the test env to avoid git picking up whatever timezone is set locally in a dev environment (eg git was using EST/EDT for me, but CI was using UTC)

<details>
<summary>original description</summary>
This adds `--allow-empty` support to `git record`. This is not very interesting on it's own, but the fun starts when you pass `--insert`, at which point this should behave as a short hand for `git commit --allow-empty -m ... ; git move -s 'siblings(@)'`.

As is, this change will only create an empty commit if there are no changes at all (staged or in the working copy). Eventually, it would be nice to have "real" support for `record --new` to create an empty commit regardless of what changes are currently staged or in the working copy, but I don't intend to tackle that at this time.
</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1673,6 +1673,7 @@ dependencies = [
 name = "git-branchless-record"
 version = "0.11.1"
 dependencies = [
+ "chrono",
  "cursive",
  "cursive_buffered_backend",
  "eyre",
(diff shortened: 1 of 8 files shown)


claytonrcarter:
I've been using this for a while and it's been working well.

PR #1631, opened 2025-11-30
feat(move): implement move --reparent like amend --reparent

### LINKED ISSUE #1537: Support `git branchless move --reparent`
### Description of the bug

The use case is the same as `git amend --reparent`, but in particular, sometimes I do a regular `git amend` and get a note that I'll need to run `git restack --merge`, and now it's too late to run `git amend --reparent`.

So in particular, the steps to reproduce are:
- Run `git move -s foo -d bar --reparent`
Or perhaps even:
- Run `git restack --reparent`

(Sorry if a Bug isn't the right venue for this, but I figured it was a sufficiently direct feature request that it made sense as an issue, rather than a discussion.)

### Expected behavior

The move happens, but without any changes being applied to the trees of the moved patches.

### Actual behavior

Not supported.

### Version of `rustc`

_No response_

### Automated bug report

_No response_

_No response_

### Version of `git`

_No response_

bryango:
Closes #1537. This is done in the following commits:
1. refactor out the `amend --reparent` logic into a `.reparent_subtree` method for the `RebasePlanBuilder`
2. move the --reparent flag into MoveOptions which may be generalized for `move` and possibly `restack`, `split` etc ~~in the future (currently not implemented and would be a no-op)~~ (now implemented, see below)
3. actually implement --reparent for the `move` command with a few `builder.reparent_subtree` calls. Currently `--reparent` conflicts with `--fixup` since ~~I am not yet sure how it should be implemented (help wanted)~~ the behavior for `--fixup --parent` may be ambiguous or confusing for users
4. add tests for `move --reparent` and showcase its intended behavior.
5. refactor: use eyre::bail! instead of assert!
6. implement `restack --reparent`
7. implement `split --reparent` (for --stack or --detach)
8. implement `sync --reparent`

I would like to test this more in real life and see if it's actually working, but I think it would be good to post it here early to gather some feedback: is this feature desired? Would this be the correct way to implement it? I have very limited knowledge on the codebase so all feedbacks are much welcomed. Thank you!

### DIFF
--- a/git-branchless-lib/src/core/rewrite/plan.rs
+++ b/git-branchless-lib/src/core/rewrite/plan.rs
@@ -1025,7 +1025,9 @@ impl<'a> RebasePlanBuilder<'a> {
         source_oid: NonZeroOid,
         dest_oids: Vec<NonZeroOid>,
     ) -> eyre::Result<()> {
-        assert!(!dest_oids.is_empty());
+        if dest_oids.is_empty() {
+            eyre::bail!("Destination OIDs must not be empty for move_subtree");
+        }
         self.initial_constraints.push(Constraint::MoveSubtree {
             parent_oids: dest_oids,
             child_oid: source_oid,
@@ -1093,7 +1095,9 @@ impl<'a> RebasePlanBuilder<'a> {
         parent_oids: Vec<NonZeroOid>,
         repo: &Repo,
     ) -> eyre::Result<()> {
-        assert!(!parent_oids.is_empty());
+        if parent_oids.is_empty() {
+            eyre::bail!("Parent OIDs must not be empty for reparent_commit");
+        }
 
         // To keep the contents of all descendant commits the same, forcibly
         // replace the child commit, and then rely on normal patch
--- a/git-branchless-lib/src/core/rewrite/plan.rs
+++ b/git-branchless-lib/src/core/rewrite/plan.rs
@@ -4,6 +4,7 @@ use std::ops::Sub;
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use bstr::ByteSlice;
 use chashmap::CHashMap;
 use eyre::Context;
 use itertools::Itertools;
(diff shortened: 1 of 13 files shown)


git-branchless/tests/test_move.rs
```diff
@@ -6329,6 +6329,258 @@ fn test_move_fixup_added_files() -> eyre::Result<()> {
...
+        |
+        o 4b5cd3e create test3_will_also_contain_test2_when_reparented.txt
+        In-memory rebase succeeded.
+        ");
+    }
+
+    // the reparented test3 contains everything, including test1 and test2
+    git.branchless("prev", &[])?;
+    {
+        let (stdout, _stderr) = git.run(&["show"])?;
```
claytonrcarter: [**suggestion (maybe change)**] Maybe consider using `git.run(&["show", "--pretty=format:", "--stat", "HEAD"])?;` ([as is done in `test_split`](https://github.com/arxanas/git-branchless/blob/master/git-branchless/tests/test_split.rs#L42)). That makes the output much less verbose and may be easier to maintain.

claytonrcarter (changes requested):
Thank you for this. The code looks simple enough, and solid. And the test seems thorough enough for the questions that I came in with.

> Currently --reparent conflicts with --fixup since I am not yet sure how it should be implemented

I suspect that it would be fine to make `--reparent` and `--fixup` conflict, since it's not at all clear to me what the intended result *should* be. (ie if you are reparenting the source commit, then you're replacing the destination commit with the source ... I think?; or would you be doing the fixup and then reparenting the original child commits of the target onto the newly squashed commit. (effectively reverting the squashed changes))

> would like to test this more in real life and see if it's actually working

The tests seem to confirm the behavior, so I think this would be OK to merge (pending the review comments), but it would also be interesting to see how it works for you in real life.

git-branchless-lib/src/core/rewrite/plan.rs
```diff
@@ -1077,6 +1078,45 @@ impl<'a> RebasePlanBuilder<'a> {
...
 
+    /// Generate a sequence of rebase steps that cause the subtree at `source_oid`
+    /// to be "reparented" by `dest_oids`, namely, keeping all contents
+    /// of descendant commits exactly the same.
+    pub fn reparent_subtree(
+        &mut self,
+        source_oid: NonZeroOid,
+        parent_oids: Vec<NonZeroOid>,
+        repo: &Repo,
+    ) -> eyre::Result<()> {
```
claytonrcarter: [**please clarify**] Please correct me if I'm wrong, but I don't think that this comment nor the fn name match the actual behavior:

- typo: I think that "by `dest_oids`"` probably should be "onto `parent_oids`".
- please confirm: does calling `self.replace_commit()` also trigger a rebase of the original commit's descendants onto the replacement commit? Or does it just do a replacement w/o rebase, leaving the descendants remaining on the original commit? ? The comment for `replace_commit()` doesn't mention anything about subtrees, which leads me to believe that this fn is also not actually triggering a rebase, but I don't know.
- if `replace_commit()` does trigger a rebase, then we just need to address that first typo; the it doesn't rebase, then we should rewrite the comment and fn name to match

claytonrcarter (changes requested):
Thank you again @bryango for working on this. 

I would like us to back out the changes to `sync`, `split` and `restack` for now. Let's move them into a different PR for now. I think that I would prefer to add `--reparent` only to `move` for now, and then see what feedback (if any) comes in the next release before extending it into more commands. 

My concern is basically the same as I mentioned before: the behavior may be surprising, especially in the commands like `sync` and `restack` that are supposed to be more automatic and/or "hands off". And since both of those commands are just special cases of `move`, there is still a way for users to have this behavior if they wish.

As for `split`, I'm left wondering if a different name for the flag may be appropriate. For example, if `--discard --reparent` will remove the file from one commit but leave it in the child, it may be that `--commute-down` may be a more descriptive name. ?? 🤷 

Then again, I tend to think of commits in terms of diffs, which isn't accurate. I'm open to compromise on this though. I don't want to dampen your enthusiasm!

bryango:
> > I would like us to back out the changes to sync, split and restack for now. Let's move them into a different PR for now. I think that I would prefer to add --reparent only to move for now, and then see what feedback (if any) comes in the next release before extending it into more commands.
> 
> @bryango Please disregard this ☝️ comment of mine. The flags are opt-in only, so there should be no risk of unwanted misuse, and I'm OK to merge this PR for all affected commands. Once it's released, we'll find out if my concerns about confusion and awkward flag names are founded or not! 😄
> 
> I think that the other comments still need to be addressed before merge, though.
> 
> I may try to make a release soon, do you think you'll have a chance to work on this again soon?

Thank you for the reviews! I will try to work on this again but a) I am slow, and b) I would like to fix all the correctness issues before it is out, so if this couldn't make into the release, I would try to target the following one. Thanks!

claytonrcarter:
Sorry for the delay on this. I will try to get to it soon. In the meantime, I've rebased it onto current master.

claytonrcarter:
Thank you, @bryango! Thank you for your patience on this and very sorry for the drawn out review. 😞 This has merged! 🚀 Can't wait to play with it.

This final version is identical to the version you pushed to 0b0a5fc in February except that I did some squashing and absorbing of some of the later commits into the relevant earlier commits. 💄

bryango:
Awesome! Thank you so much for maintaining git-branchless!

PR #1632, opened 2025-12-07
feat(record): add --fixup option similar to reword --fixup

bryango:
Implement `record --fixup=<rev>` similar to `reword --fixup=<rev>`, i.e. by generating a commit message with the `fixup! ...` prefix, suitable for use with `git rebase --autosquash`.

There is an alternative behavior that we can implement: since now we have `move --fixup -d <rev>` we can directly squash the changes into `<rev>` without even committing it at all. This behavior would be similar to `jj squash --into=<rev>`. However, I opted not to do this as the squash operation may fail (e.g. due to merge conflicts) and it seems more natural for the semantics of `record` to be the same as `reword`. What do you think?

By the way, although `record` and `reword` have similar semantics, their implementations are quite different: `record` simply calls `git commit` directly, and we simply pass along the `--fixup` flag after resolving the revset expressions.

### DIFF
--- a/git-branchless-record/src/lib.rs
+++ b/git-branchless-record/src/lib.rs
@@ -15,8 +15,8 @@ use std::fmt::Write;
 use std::time::SystemTime;
 
 use git_branchless_invoke::CommandContext;
-use git_branchless_opts::{MessageArgs, RecordArgs};
-use git_branchless_reword::edit_message;
+use git_branchless_opts::{MessageArgs, RecordArgs, ResolveRevsetOptions, Revset};
+use git_branchless_reword::{edit_message, resolve_commit_to_fixup, ResolveFixupCommitError};
 use itertools::Itertools;
 use lib::core::check_out::{check_out_commit, CheckOutCommitOptions, CheckoutTarget};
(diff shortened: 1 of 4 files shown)


git-branchless-record/src/lib.rs
```diff
@@ -169,11 +170,38 @@ fn record(
...
+                    &repo,
+                    &mut dag,
+                    effects,
+                    &revset,
+                    &ResolveRevsetOptions::default(),
+                    None
+                )?);
+                let commit = commit.get_oid().to_string();
+                args.extend(["--fixup".into(), commit]);
+            };
```
claytonrcarter: [**nitpick**] I think I would prefer to use `.to_string()` instead of `.into()` for all of these `&str` conversions. 

To be honest, I don't know if one or the other is considered "more" idiomatic, but – in this file at least – there are few/no other uses of `.into()` while there are a few of `.to_string()`

claytonrcarter (changes requested):
Thank you, @bryango! The code looks good to me, but I have a couple of requests.

Also, please add a test for `record --fixup`.

PR #1651, opened 2026-02-23
fix(worktree): find parent repo for bare clones

feosuna1:
Running `git-branchless` from a linked worktree of a bare-clone repo caused a fatal crash rather than displaying output.

**Reproduction Steps**

```bash
# cwd = $USER/repos
git clone --bare https://github.com/arxanas/git-branchless git-branchless
git -C git-branchless/ branchless init
git -C git-branchless/ worktree add ../master master
git -C master sl
```

**Result**

```text
   0: could not open repository: could not find repository at '$USER/repos/; class=Repository (6); code=NotFound (-3)
   1: could not find repository at '$USER/repos/'; class=Repository (6); code=NotFound (-3)

Location:
   git-branchless-hook/src/lib.rs:516
…
```

This was happening because `open_worktree_parent_repo` assumed the worktree's git directory looked like `$USER/repos/git-branchless/.git/worktrees/master`. For bare clones there is no `.git` subdirectory, so the path is `$USER/repos/git-branchless/worktrees/master` — three levels up lands at the wrong directory.

The fix uses git2's `commondir()`, which directly returns the parent repository's git directory regardless of whether it's a bare clone or not.

Fixes: #1652

### DIFF
--- a/git-branchless-lib/tests/test_repo.rs
+++ b/git-branchless-lib/tests/test_repo.rs
@@ -316,16 +316,19 @@ fn test_open_worktree_parent_repo_bare_clone() -> eyre::Result<()> {
     } = make_git_with_remote_repo()?;
 
     // Create a bare clone of the original repo.
-    original_repo.init_repo()?;
+    {
+        original_repo.init_repo()?;
 
-    // `clone_repo_into` reinitializes the destination after cloning, adding a `.git` directory
-    // and losing the bare state. Run `git clone --bare` directly instead.
-    original_repo.run(&[
-        "clone",
-        "--bare",
-        original_repo.repo_path.to_str().unwrap(),
-        cloned_repo.repo_path.to_str().unwrap(),
-    ])?;
+        // `git.clone_repo_into` reinitializes the destination after cloning,
+        // adding a `.git` directory and losing the bare state. Run `git clone
+        // --bare` directly instead.
+        original_repo.run(&[
+            "clone",
+            "--bare",
+            original_repo.repo_path.to_str().unwrap(),
+            cloned_repo.repo_path.to_str().unwrap(),
+        ])?;
+    }
 
     // Create a linked worktree from the bare clone.
     let GitWorktreeWrapper {
@@ -337,6 +340,7 @@ fn test_open_worktree_parent_repo_bare_clone() -> eyre::Result<()> {
         .get_repo()?
         .open_worktree_parent_repo()?
         .expect("expected to find parent repo for worktree of bare clone");
+
     assert_eq!(
         std::fs::canonicalize(parent_repo.get_path())?,
         std::fs::canonicalize(cloned_repo.repo_path)?,


feosuna1 (commented):
This is my first rust contribution. In other words, I have no idea what I'm doing, I'm happy to make any adjustments to this PR.

git-branchless-lib/tests/test_repo.rs
```diff
@@ -303,3 +303,42 @@ fn test_worktree_working_copy_path() -> eyre::Result<()> {
...
+        "worktree",
+        "add",
+        "--detach",
+        worktree_path.to_str().unwrap(),
+    ])?;
+
+    let worktree_repo = Repo::from_dir(&worktree_path)?;
+    let parent_repo = worktree_repo
+        .open_worktree_parent_repo()?
+        .expect("expected to find parent repo for worktree of bare clone");
```
claytonrcarter: I'm don't work with bare repos or worktrees often, so please correct me if I'm wrong, but I think that this can be replaced with (lifted this mostly from the preceding test.)

```suggestion
    let GitWorktreeWrapper {
        temp_dir: _,
        worktree,
    } = make_git_worktree(&bare_git, "worktree")?;

    let parent_repo = worktree
        .get_repo()?
        .open_worktree_parent_repo()?
        .expect("expected to find parent repo for worktree of bare clone");
```

Then you can also remove `worktree_path`, too. It looks like `make_git_worktree()` does what we need, but – again – please confirm.

git-branchless-lib/tests/test_repo.rs
```diff
@@ -303,3 +303,42 @@ fn test_worktree_working_copy_path() -> eyre::Result<()> {
...
+    assert_eq!(
+        std::fs::canonicalize(parent_repo.get_path())?,
+        std::fs::canonicalize(&bare_path)?,
+    );
+
+    drop(temp_dir);
```
claytonrcarter: This `drop()` can be removed, I think. `temp_dir` will be dropped when the fn returns already, and I believe that `tempfile::tempdir()` will be cleaned up automatically when that happens.

git-branchless-lib/tests/test_repo.rs
feosuna1: You're absolutely right. I ended up refactoring this test entirely to try to leverage as many of the `test.rs` helpers that I could find. Updated in 4e5c11d7d14215d39011a7e908b0321b269c1754.

I tried to use [`Git.clone_repo_into`](https://github.com/arxanas/git-branchless/blob/0504b5553a90443aba2772b997dbb1fecf5a3bd2/git-branchless-lib/src/testing.rs#L471) to create a bare clone, but it didn't work.  That method [reinitializes](https://github.com/arxanas/git-branchless/blob/0504b5553a90443aba2772b997dbb1fecf5a3bd2/git-branchless-lib/src/testing.rs#L494) the cloned repo and it loses its "bare" state.

Part of me thinks that [`init_repo_with_options`](https://github.com/arxanas/git-branchless/blob/0504b5553a90443aba2772b997dbb1fecf5a3bd2/git-branchless-lib/src/testing.rs#L430)'s `git config` and `git branchless init` calls should be factored out into a separate helper method, so that both `init_repo_with_options` and `clone_repo_into` calls the helper method directly. This would let `clone_repo_into` by-pass the git reinitialization.

Do you have any thoughts on how to make it better?

PR #1661, opened 2026-04-24
feat(nix): add devshell, fix flake build, and update nixpkgs input

bitbloxhub:
Not cherry-picked as it is based directly off of upstream master already.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,10 @@
 {
   description = "git-branchless";
 
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+  };
+
   outputs = { self, nixpkgs, ... }:
     let
       lib = nixpkgs.lib;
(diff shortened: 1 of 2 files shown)


claytonrcarter:
Thank you, @bitbloxhub. I'm not a nix user, though, so I'm not sure how to review this. Can you provide some more context on what I should be looking at and trying?

flake.nix
```diff
@@ -14,67 +19,102 @@
       foreachSystem = lib.genAttrs systems;
     in
     {
-      overlays.default = (final: prev: {
+      overlays.default = (
+        final: prev: {
```
claytonrcarter, line 23: If I understand correctly, it looks like this file has a combination of new stuff (eg `postPatch = "";` on line 35) and reformatting/whitespace changes. Can you extract the formatting changes into their own commit so that I can more easily see the new additions?

claytonrcarter (changes requested):

bitbloxhub:
Assuming you have nix installed with flakes enabled, here are a few things to try:
```bash
nix develop # Tests out the devshell, you can check that `which <cargo|rustc|etc.>` outputs to a path starting in `/nix/store`
nix build ".#git-branchless" # Compile git-branchless with Nix, outputs to `result/bin`
nix flake check # Run tests in a nix build
```

flake.nix
bitbloxhub: Redid changes, and nixfmt is in seperate commit now.

claytonrcarter: Thank you! I'll try to take a look at these tonight.

claytonrcarter:
Thank you!

Issue #1416, opened 2024-10-09
Many tests broken on Git v2.46


arxanas:
### Description of the bug

Many tests are spuriously broken under Git v2.46 due to changes in reference transaction handling. Example: https://github.com/arxanas/git-branchless/actions/runs/11254346187/job/31291546697

To reproduce:

1. Set `TEST_GIT` and `TEST_GIT_EXEC_PATH` to point to an instance of Git v2.46.
2. Run `cargo test --workspace --all-features --no-fail-fast`

The ideal fix would probably be to implement https://github.com/arxanas/git-branchless/issues/1334 and update all tests accordingly.

### Expected behavior

No test failures.

### Actual behavior

Test failures in these targets:

```
error: 8 targets failed:
    `-p git-branchless --test test_amend`
    `-p git-branchless --test test_gc`
    `-p git-branchless --test test_hooks`
    `-p git-branchless --test test_move`
    `-p git-branchless --test test_snapshot`
    `-p git-branchless --test test_sync`
    `-p git-branchless --test test_undo`
    `-p git-branchless-submit --test test_github_forge`
```

### Version of `git-branchless`

d011db96cf3f034523d05ab60fe233dd120d2a9a

### Version of `git`

git version 2.46.2

cdmistman:
I'm also experiencing this with git 2.46.0

arxanas:
@cdmistman are you running the test suite for a particular purpose?

cdmistman:
I use nix to manage my dotfiles, including installation of packages on my Darwin machines. The tests run by default when installing a package

alerque:
This has also fauled up Arch Linux packaging, I tried to bump the package to 0.10.0 and it won't pass the packaging check since the internal check function is failing. Is the whole software considered broken on new Git (in which case I should gate the upgrade behind a requirement on having old Git available, which would essentially block system upgrades for anybody with this app) or is just the test suite broken? In that case I can skip it, but it would also be nice to tag anything that shouldn't be run behind a feature flag or whatever so we can run tests that are expected to pass.

arxanas:
@alerque To the best of my knowledge, it's only the test suite that's broken.

PR #1671, opened 2026-05-21
ci: update linux versions

claytonrcarter:
We currently test against 4 git versions from 2020-2022 in CI. This updates our build matrix to test against 5 (approximately annual) versions from 2022-2026.


claytonrcarter:
Probably related: #1416

PR #1675, opened 2026-05-26
fix(help): show correct name with --version

claytonrcarter:
Without this, clap defaults the name to the crate name (in this case, git-branchless-opts), not the application or workspace name.

```
❯ git branchless --version
git-branchless-opts 0.11.1

❯ cargo run -- --version
git-branchless 0.11.1
```


arxanas:
Lol thank you, this always bugged me

PR #1676, opened 2026-05-31
feat(smartlog): show linked worktrees

casey-chow:
## Summary

Adds worktree awareness to `git smartlog`.

Smartlog now discovers linked worktrees via `git worktree list --porcelain [-z]`, includes active linked-worktree HEADs in the rendered graph, and annotates commits with the worktrees currently checked out there. This makes it easier to see work spread across multiple worktrees without switching context.

Worktrees are named in the view using shortest unique path suffix.

## Screenshots

Here's how it looks on my personal machine: 
CleanShot 2026-05-31 at 11 49 45@2x

## Details

- Adds a shared linked-worktree snapshot model.
- Shows current and linked worktree markers in smartlog output.
- Includes linked-worktree HEAD commits even when they are outside the default smartlog revset.
- Skips prunable worktrees.
- Syncs active worktree HEAD OIDs into the DAG before rendering, preventing crashes for detached worktree commits not already known to branchless.

## AI Assistance Disclosure

This PR was generated with heavy assistance from OpenAI Codex. I reviewed the design, behavior, and generated changes before submitting.

### DIFF
--- a/git-branchless-lib/src/core/node_descriptors.rs
+++ b/git-branchless-lib/src/core/node_descriptors.rs
@@ -395,7 +395,7 @@ impl<'a> WorktreeDescriptor<'a> {
 impl NodeDescriptor for WorktreeDescriptor<'_> {
     fn describe_node(
         &mut self,
-        _glyphs: &Glyphs,
+        glyphs: &Glyphs,
         object: &NodeObject,
     ) -> eyre::Result<Option<StyledString>> {
         if self.worktree_snapshot.entries.len() <= 1 {
@@ -407,7 +407,11 @@ impl NodeDescriptor for WorktreeDescriptor<'_> {
             .find_by_head_oid(object.get_oid())
             .into_iter()
             .map(|entry| {
-                let icon = if entry.is_current { "ᐅ" } else { "⎇" }.to_string();
+                let icon = if entry.is_current {
+                    glyphs.worktree_current
+                } else {
+                    glyphs.worktree_linked
+                };
                 format!("{icon} {}", entry.display_name())
             })
             .collect();
--- a/git-branchless-lib/src/core/node_descriptors.rs
+++ b/git-branchless-lib/src/core/node_descriptors.rs
@@ -377,6 +377,52 @@ impl NodeDescriptor for BranchesDescriptor<'_> {
     }
 }
 
+/// Display linked worktrees which currently have a commit checked out.
+#[derive(Debug)]
+pub struct WorktreeDescriptor<'a> {
+    worktree_snapshot: &'a crate::core::worktree::WorktreeSnapshot,
+}
+
+impl<'a> WorktreeDescriptor<'a> {
+    /// Constructor.
+    pub fn new(
+        worktree_snapshot: &'a crate::core::worktree::WorktreeSnapshot,
+    ) -> eyre::Result<Self> {
+        Ok(Self { worktree_snapshot })
+    }
+}
+
+impl NodeDescriptor for WorktreeDescriptor<'_> {
+    fn describe_node(
+        &mut self,
+        _glyphs: &Glyphs,
+        object: &NodeObject,
+    ) -> eyre::Result<Option<StyledString>> {
+        if self.worktree_snapshot.entries.len() <= 1 {
+            return Ok(None);
+        }
+
+        let worktrees: Vec<String> = self
+            .worktree_snapshot
+            .find_by_head_oid(object.get_oid())
+            .into_iter()
+            .map(|entry| {
+                let icon = if entry.is_current { "ᐅ" } else { "⎇" }.to_string();
+                format!("{icon} {}", entry.display_name())
+            })
+            .collect();
+
+        if worktrees.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(StyledString::styled(
+                format!("({})", worktrees.join(", ")),
+                BaseColor::Blue.light(),
+            )))
+        }
+    }
+}
+
 /// Display the associated Phabricator revision for a given commit.
 #[derive(Debug)]
 pub struct DifferentialRevisionDescriptor<'a> {
--- a/git-branchless-lib/src/core/worktree.rs
+++ b/git-branchless-lib/src/core/worktree.rs
@@ -1,0 +1,362 @@
+//! Utilities for discovering and describing linked worktrees.
+
+use std::collections::HashSet;
+#[cfg(unix)]
+use std::ffi::OsString;
+#[cfg(unix)]
+use std::os::unix::ffi::OsStringExt;
+use std::path::{Path, PathBuf};
+
+use eyre::Context;
+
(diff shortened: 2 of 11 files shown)


git-branchless-lib/src/core/worktree.rs
```diff
@@ -0,0 +1,362 @@
...
+        .map(|path| canonicalize_best_effort(&path));
+    let result = git_run_info.run_silent(
+        repo,
+        None,
+        &["worktree", "list", "--porcelain", "-z"],
+        GitRunOpts {
+            treat_git_failure_as_error: false,
+            ..Default::default()
+        },
+    )?;
```
claytonrcarter, line 335: [**suggestion (non-blocking)**] Did you consider using `git2::Repository::worktrees()` to get the list of existing worktrees instead of shelling out to git? 

This isn't a blocker, because I see that we shell out to git for some worktree operations already, specifically when adding worktrees (eg for `git test`), and adding a `repo.worktrees()` fn would probably some boilerplate to "wrap" or reexport the worktrees as our own type. 

I'm curious if you tried and ruled it out, or if you think it may be preferable?

git-branchless/tests/test_smartlog_worktree.rs
```diff
@@ -0,0 +1,183 @@
...
+        "worktree",
+        "add",
+        "--detach",
+        temp_dir.to_string_lossy().as_ref(),
+        commit_oid,
+    ])?;
+
+    let stdout = git.smartlog()?;
+    assert!(stdout.contains(&commit_oid[..7]), "stdout was: {stdout}");
+    assert!(stdout.contains("outside"), "stdout was: {stdout}");
```
claytonrcarter: [**please change**] Where possible, we prefer to use `insta::assert_snapshot!()` because it tends to be more revealing for debugging then output changes. I see that you've used that in some of these tests; can you apply to the rest, too? Or give more context on why you feel it may not be warranted here?

git-branchless-lib/src/core/node_descriptors.rs
```diff
@@ -377,6 +377,52 @@ impl NodeDescriptor for BranchesDescriptor<'_> {
...
+        if self.worktree_snapshot.entries.len() <= 1 {
+            return Ok(None);
+        }
+
+        let worktrees: Vec<String> = self
+            .worktree_snapshot
+            .find_by_head_oid(object.get_oid())
+            .into_iter()
+            .map(|entry| {
+                let icon = if entry.is_current { "ᐅ" } else { "⎇" }.to_string();
```
claytonrcarter: [**please change**] Let's move these into `Glyphs` and then reference them here like `glyphs.worktree_current` and `glyphs.worktree_linked` (or whatever names you prefer). I think that's the normal way that we record these sorts of things, and it makes it easy to maintain "simple" text versions vs "pretty" unicode versions.

claytonrcarter (changes requested):
Thank you @casey-chow! This looks great, with a couple of questions and some small change requests.

[**question**] I didn't look specifically at how this would affect performance. Do you have any reason to think that it would adversely affect rendering of the smartlog? (Aside from cases where someone may have a pathological number of worktrees 😆 )

[**please change**] The only behavior change that I'd like to see is to put this behind a command option and/or config option, with the default being "off" or "do not include linked worktrees in the smartlog", ie the current behaviour. For example, 
- `git smartlog` → no worktrees (current behavior)
- `git smartlog --worktrees` → include worktrees (this PR)
- `git config branchless.smartlog.showWorktrees true ; git smartlog` → include worktrees by default

As is, I don't directly use worktrees except via `git test`, and this branch includes those worktrees in the output, even though they are only an "implementation detail" of `git test` and not something I care about. I assume the same will be true for anyone else that's used `git test --strategy worktrees`, too, so I'd prefer to make this feature "opt in".

Here's an example of where we have a setting for the smartlog: https://github.com/arxanas/git-branchless/blob/master/git-branchless-lib/src/core/config.rs#L104-L109

Here's an example of another place where "flag + config" is supported: https://github.com/arxanas/git-branchless/blob/master/git-branchless-smartlog/src/lib.rs#L828-L836

If you have questions about how to implement that, please let me know!

Thank you again for working on this! Even though I may not use worktrees, I can see how helpful this will be to folks that do.

casey-chow:
Thanks for taking a look! For your top level comments:

> [question] I didn't look specifically at how this would affect performance. Do you have any reason to think that it would adversely affect rendering of the smartlog? (Aside from cases where someone may have a pathological number of worktrees 😆 )

I'm a pretty intense user of worktrees (and I am using this on a massive monorepo) and I can't say I've run into a situation (yet) where I've actually found this to be an issue! My assumption of course is that we're talking 15-20 worktrees on a human scale, and on a model scale there's relatively little reason to worry about the performance of git-branchless.

> [please change] The only behavior change that I'd like to see is to put this behind a command option and/or config option, with the default being "off" or "do not include linked worktrees in the smartlog", ie the current behaviour. [...]

Agreed. One thing that's a bit annoying about this option is that everything that returns a smartlog will now require the config option, so there's no way to do `git undo --worktrees` in a way that, I'd say, makes sense. I think it's a worthwhile addition for flexibility, but do want to make sure that tradeoff is called out.

---

I'll go ahead and address your fixes as they come through! 

I have a pile of other PRs I've been doing for worktree compatibility in my fork, but at this stage they're mostly (functional, but still) AI slop that I don't feel is ready for review. If there is anyone in orbit that has (more) thoughts on how git-branchless should work with worktrees I'd be happy to chat with them too to make sure everything stays coherent.

git-branchless-lib/src/core/worktree.rs
casey-chow: `git2::Repository::worktrees()` is weirdly specific about what it [returns](https://docs.rs/git2/latest/git2/struct.Worktree.html) (names only, with a numeric deduplicator), so retrieving it would require quite a fanout to get the necessary information, versus `git worktree list --porcelain` presents it all in one go.

git-branchless/tests/test_smartlog_worktree.rs
casey-chow: Done, I believe I got them all.

git-branchless-lib/src/core/node_descriptors.rs
casey-chow: huh, TIL that there's a simple version; I was wondering why sometimes I saw the non-Unicode variants.

Fixed!

git-branchless-lib/src/core/worktree.rs
claytonrcarter: Sounds good. Thanks for considering.

claytonrcarter:
Thank you for implementing this, @casey-chow!

> If there is anyone in orbit that has (more) thoughts on how git-branchless should work with worktrees I'd be happy to chat with them too to make sure everything stays coherent.

Things are pretty quiet around these days, but feel free to start a discussion about worktree ideas. You can also check out the Discord to see if anyone is around. Failing that, you can create issues w/ your ideas, or just push some code to discussion.

Thanks again!

PR #1677, opened 2026-05-31
fix(test): sanitize path-invalid characters in test command slug

### LINKED ISSUE #1629: (Windows) Panic during `git test` if `-x` script contains characters which are invalid in a file path
### Description of the bug

Running a `git test run ...` where the script part contains e.g. `:` panics on Windows.
It seems the `:` is copied as-is into the filepath, and because it's Windows, that's not ok. On WSL it works fine.

## Minimal reproduction
`git test run main..HEAD -x 'echo :'` 

### Expected behavior

I expected the script to run as usual, meaning I expect that the command will be converted to a valid path
before trying to make a new directory.

### Actual behavior

While running `git test main..HEAD -x 'pnpm check:prettier'` I got the following output:
```
   0: Failed waiting on workers
   1: Worker 1 failed when processing commit ec9d22ab59b5d0aca2679c8e5db19ddd07c5b34f: Creating command directory "D:/Projects/awpw-node/.git/branchless\\test\\13f2bb5e83b30b6b69add54247527753ed5c8ef1\\pnpm__run__check:prettier"
```

### Version of `rustc`

rustc 1.88.0-nightly (2e6882ac5 2025-05-05)

### Automated bug report

[git-branchless-bug-report.txt](https://github.com/user-attachments/files/23655842/git-branchless-bug-report.txt)

> edit: typos

StressTestor:
## Problem

`git test -x <command>` slugifies the command into a directory name (under `.git/branchless/test/<tree_oid>/<slug>`) via `make_test_command_slug`, which only replaced `/`, space, and newline:

```rust
command.replace(['/', ' ', '\n'], "__")
```

On Windows the other characters that are invalid in a path component (`< > : " \ | ? *` and ASCII control characters) passed through unchanged, so `create_dir_all` failed and `git test` errored out. Pipes, redirects, and globs are everyday `-x` syntax (`cargo build 2>&1 | tee log`, `eslint 'src/**/*.js'`), so this bites real commands. Fixes #1629.

## Fix

Replace any control character, or any of `/ \ < > : " | ? *`, or a space, with `__`. Sanitization is unconditional (not `#[cfg(windows)]`) so slugs are portable across platforms and the behavior is testable everywhere.

## On the design choice (replacing the invalid set, not all non-alphanumerics)

The issue suggested two robust directions: a platform-specific list of chars to exclude, **or** replacing all non-alphanumeric characters. I went with the targeted invalid-char list on purpose:

- It preserves the slug for ordinary commands. `bash test.sh 10` stays `bash__test.sh__10`; replacing all non-alphanumerics would make it `bash__test__sh__10`.
- That keeps users' existing cached test results valid across the upgrade (the cache is keyed by the slug), and keeps the existing `test_test.rs` snapshots unchanged. I checked the test corpus — no existing test command contains a character this newly replaces, so no snapshot changes.

Happy to switch to replace-all-non-alphanumeric if you'd prefer the simpler rule; it'd just churn the snapshots and existing caches.

**Tradeoff worth noting:** collapsing more characters to `__` does widen slug collisions (e.g. `echo a:b` and `echo a|b` now map to the same slug, where on Unix they previously differed). The function already collided by design (`a/b`, `a b`, `a\nb` all mapped to `a__b`), so this doesn't introduce a new class of behavior, and on Windows the affected commands didn't work at all before. A collision-free slug would need hashing, which would orphan every existing cache — out of scope for this fix, but I'm glad to follow up if you want it.

## Testing

Added unit tests on `make_test_command_slug` asserting (a) no path-invalid char survives, (b) control characters are replaced, and (c) an ordinary command keeps its readable slug. These assert the function's output contract, so they fail on macOS/Linux too — which sidesteps the "couldn't reproduce on a non-Windows machine" problem noted on the issue.

`cargo fmt --check`, `cargo clippy --workspace --all-features --all-targets -- --deny warnings`, and the new tests all pass locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- (#1629): `git test` no longer fails to create its results directory when the test command contains characters that are invalid in a path on Windows (such as `|`, `<`, `>`, `:`, `*`, `?`, `"`, or `\`).
+
 ## [v0.11.1] - 2026-05-21
 
 ### Changed
--- a/git-branchless-lib/src/git/test.rs
+++ b/git-branchless-lib/src/git/test.rs
@@ -43,7 +43,23 @@ pub const TEST_ABORT_EXIT_CODE: i32 = 127;
 
 /// Convert a command string into a string that's safe to use as a filename.
 pub fn make_test_command_slug(command: String) -> String {
-    command.replace(['/', ' ', '\n'], "__")
+    // The slug is used as a directory name, so replace every character that is
+    // invalid in a path component on any supported platform. Windows forbids
+    // `<>:"/\|?*` and ASCII control characters (`/` is also a path separator on
+    // Unix); spaces are replaced for readability. Without this, `git test -x`
+    // commands containing e.g. pipes or redirects fail to create their cache
+    // directory on Windows. See
+    // https://github.com/arxanas/git-branchless/issues/1629.
+    command.replace(
+        |c: char| {
+            c.is_control()
+                || matches!(
+                    c,
+                    '/' | '\\' | '<' | '>' | ':' | '"' | '|' | '?' | '*' | ' '
+                )
+        },
+        "__",
+    )
 }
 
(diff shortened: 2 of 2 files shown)


claytonrcarter:
Looks great! Thank you!

PR #1682, opened 2026-06-13
submit: omit commit title from GitHub PR body

### LINKED ISSUE #1680: Title duplicated in pull request body
### Description of the bug

Submit a GitHub pull request for a one-patch stack, whose commit message has the conventional Git format:

```
Brief summary of the change

Followed by 27 paragraphs of rambling explanation!
```

### Expected behavior

I would expect the title of the PR to be the first line of the commit message, and for the PR description to be the rest of the commit message, not repeating the first line.

### Actual behavior

In fact the title is repeated in the body of the commit message. Here is an example:

- https://github.com/endlessm/threadbare/pull/2174

This PR's body duplicates the title.

This matters to me because our project uses the squash-merge strategy, and we use the PR title + description as the squashed commit message. It is ugly to have the title in the commit log twice. I try to manually edit every PR I file to work around this!

### Version of `rustc`

_No response_

### Automated bug report

_No response_

git-branchless-opts 0.10.0

### Version of `git`

git version 2.54.0

VrtxOmega:
Fixes #1680.

## Problem
`git branchless submit --forge github` uses the commit summary as the GitHub pull request title, but the create and update paths also include the full commit message in the pull request body. For a conventional subject/body commit message, that repeats the subject in the PR description. Projects that squash-merge with title plus body then get the title twice in the final commit message.

## Changes
- Derive the GitHub pull request body from the commit-message body after the summary.
- Use the same body helper for initial PR creation and later stack metadata updates.
- Omit the message separator when a commit has no body, leaving only the stack metadata.
- Add a GitHub forge regression test for a subject plus multi-paragraph body commit.

## Verification
- `cargo build -p git-branchless -p git-branchless-submit`
- `TEST_GIT="$(which git)" TEST_GIT_EXEC_PATH="$(git --exec-path)" cargo test -p git-branchless-submit test_github_forge --test test_github_forge -- --nocapture`
- `cargo fmt --all -- --check`

### DIFF
--- a/git-branchless-lib/src/git/object.rs
+++ b/git-branchless-lib/src/git/object.rs
@@ -109,6 +109,12 @@ impl<'repo> Commit<'repo> {
         BString::from(self.inner.message_bytes())
     }
 
+    /// Get the commit message body.
+    #[instrument]
+    pub fn get_body(&self) -> Option<&str> {
+        self.inner.body()
+    }
+
     /// Get the commit message, without any whitespace trimmed.
     #[instrument]
     pub fn get_message_raw(&self) -> BString {
--- a/git-branchless-submit/src/github.rs
+++ b/git-branchless-submit/src/github.rs
@@ -18,6 +18,7 @@ use lib::core::eventlog::EventLogDb;
 use lib::core::repo_ext::RepoExt;
 use lib::core::repo_ext::RepoReferencesSnapshot;
 use lib::git::CategorizedReferenceName;
+use lib::git::Commit;
 use lib::git::GitErrorCode;
 use lib::git::GitRunInfo;
 use lib::git::RepoError;
@@ -56,6 +57,14 @@ fn commit_summary_slug(summary: &str) -> String {
     } else {
         summary_slug.to_owned()
     }
+}
+
+fn get_commit_message_body(commit: &Commit<'_>) -> String {
+    let mut body = commit.get_body().unwrap_or_default().to_owned();
+    if !body.is_empty() && !body.ends_with('\n') {
+        body.push('\n');
+    }
+    body
 }
 
 fn singleton<K: Debug + Eq + Hash, V: Clone>(
@@ -345,7 +354,7 @@ impl Forge for GithubForge<'_> {
 
             let commit = self.repo.find_commit_or_fail(commit_oid)?;
             let title = String::from_utf8_lossy(&commit.get_summary()?).into_owned();
-            let body = String::from_utf8_lossy(&commit.get_message_pretty()).into_owned();
+            let body = get_commit_message_body(&commit);
             try_exit_code!(self.client.create_pull_request(
                 effects,
                 client::CreatePullRequestArgs {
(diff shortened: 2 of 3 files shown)


git-branchless-submit/src/github.rs
```diff
@@ -58,6 +59,19 @@ fn commit_summary_slug(summary: &str) -> String {
     }
 }
 
+fn get_commit_message_body(commit: &Commit<'_>) -> eyre::Result<String> {
```
arxanas: suggestion: Instead, can we expose a wrapper for https://docs.rs/git2/latest/git2/struct.Commit.html#method.body? That should be more direct (and less likely to be buggy in case of invalid UTF-8, etc.).

arxanas (changes requested):

VrtxOmega:
Done, thanks. I added a thin `Commit::get_body()` wrapper around `git2::Commit::body()` and switched the GitHub PR-body path to use that instead of manually parsing the pretty commit message.

Local verification:

```bash
TEST_GIT=$(command -v git) TEST_GIT_EXEC_PATH=$(git --exec-path) cargo test -p git-branchless-submit --tests
cargo fmt --check
git diff --check
```

git-branchless-lib/src/git/object.rs
```diff
@@ -109,6 +109,12 @@ impl<'repo> Commit<'repo> {
         BString::from(self.inner.message_bytes())
     }
 
+    /// Get the commit message body.
+    #[instrument]
+    pub fn get_body(&self) -> Option<&str> {
```
arxanas, line 114: Ah, sorry, looking at the other wrapper methods, this should also return a `BString` (and call `body_bytes` internally). Then, downstream, we should call `String::from_utf8_lossy` to handle invalid UTF-8 and be sure that we still render _something_ in the PR body.

If `body_bytes` returns `None`, then you can propagate the error using the same pattern as for `get_summary` above.

git-branchless-submit/src/github.rs
```diff
@@ -58,6 +59,14 @@ fn commit_summary_slug(summary: &str) -> String {
     }
 }
 
+fn get_commit_message_body(commit: &Commit<'_>) -> String {
+    let mut body = commit.get_body().unwrap_or_default().to_owned();
+    if !body.is_empty() && !body.ends_with('\n') {
+        body.push('\n');
```
arxanas, line 65: I don't think this `\n` logic is necessary —

- The docs specify that trailing whitespace is removed, so the check for `body.ends_with('\n')` is redundant (the only case where this is not true is already handled by `body.is_empty()`).
- It might not be obvious, but the calling code uses a multi-line string literal, so it already adds a newline after `{commit_message_body}`.
- We can see in the current snapshot tests that the `body` ends with `\n\n`, indicating that there's an extra trailing newline.

I think we can remove this function altogether and write a line similar to this one at the call-sites:

```rust
            let title = String::from_utf8_lossy(&commit.get_summary()?).into_owned();
```

but for the body instead.

arxanas (changes requested):

VrtxOmega:
Addressed in d15f84601e6e84864ab629eaa784ffcfc7113c05. The wrapper now preserves BString bytes via body_bytes(), and both GitHub call sites use String::from_utf8_lossy. I removed the newline helper; the existing multi-paragraph snapshot now has one trailing newline.

One adjustment to the suggested error handling: the pinned libgit2 implementation returns NULL when a message has no body, including ordinary summary-only commits. I retained Option<BString> for that valid case instead of returning an error. A raw-commit test covers absent/whitespace-only bodies, multiple paragraphs and invalid UTF-8; an end-to-end GitHub test verifies invalid bytes render as replacement characters rather than dropping the entire body. Existing title-only submissions still pass.

Validation on Rust 1.86: the new library test passed, all 14 submit tests passed (including six GitHub forge tests), cargo build -p git-branchless passed, and cargo fmt --check / git diff --check passed.

